PR #188, opened 2015-10-07
(maint) Add license, update yard dependency in gemspec

haus:
Having a license in the gemspec is a good practice, and gem will warn
when building if a dependency has no version requirements, so this
commit adds the license field and adds a versioning requirement to the
yard development dependency.

### DIFF
--- a/vanagon.gemspec
+++ b/vanagon.gemspec
@@ -7,13 +7,14 @@ Gem::Specification.new do |gem|
 
   gem.summary = "Another mega-package build tool, now with more Make"
   gem.description = "Vanagon is a tool to build a single package out of a project, which can itself contain one or more components."
+  gem.license = "Apache-2.0"
 
   gem.authors  = ['Puppet Labs']
   gem.email    = 'info@puppetlabs.com'
   gem.homepage = 'http://github.com/puppetlabs/vanagon'
 
   gem.add_development_dependency('rspec', ["~> 3.0"])
-  gem.add_development_dependency('yard')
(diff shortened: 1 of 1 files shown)


mhaskel:
:+1:

vanagon.gemspec
```diff
@@ -7,13 +7,14 @@ Gem::Specification.new do |gem|
...
   gem.description = "Vanagon is a tool to build a single package out of a project, which can itself contain one or more components."
+  gem.license = "Apache-2.0"
 
   gem.authors  = ['Puppet Labs']
   gem.email    = 'info@puppetlabs.com'
   gem.homepage = 'http://github.com/puppetlabs/vanagon'
 
   gem.add_development_dependency('rspec', ["~> 3.0"])
-  gem.add_development_dependency('yard')
+  gem.add_development_dependency('yard', '~> 0.8')
```
stahnma: yard generation also requires redcarpet, but for some reason yard doesn't pull it in.

vanagon.gemspec
```diff
@@ -7,13 +7,14 @@ Gem::Specification.new do |gem|
 
   gem.summary = "Another mega-package build tool, now with more Make"
```
stahnma: Let's use the GH repo tagline rather than "Another...."

haus: I didn't even change it!

stahnma: now you know how it feels!

haus: updated

shrug:
:+1:

PR #190, opened 2015-10-09
(RE-5758) Move Vanagon bill-of-materials location on OSX

heathseals:
Starting with OSX 10.11, /usr is protected by default.  This commit
updates the location of the bill-of-materials to use the docdir in
/usr/local.

If we're in an upgrade, the preinstall script removes the old
bill-of-materials in /usr if it exists.

### DIFF
--- a/lib/vanagon/platform/osx.rb
+++ b/lib/vanagon/platform/osx.rb
@@ -18,8 +18,8 @@ class Vanagon
          "gunzip -c #{project.name}-#{project.version}.tar.gz | '#{@tar}' -C '$(tempdir)/osx/build/root/#{project.name}-#{project.version}' --strip-components 1 -xf -",
 
          # Move bill-of-materials into a docdir
-         "mkdir -p $(tempdir)/osx/build/root/#{project.name}-#{project.version}/usr/share/doc/#{project.name}",
-         "mv $(tempdir)/osx/build/root/#{project.name}-#{project.version}/bill-of-materials $(tempdir)/osx/build/root/#{project.name}-#{project.version}/usr/share/doc/#{project.name}/bill-of-materials",
+         "mkdir -p $(tempdir)/osx/build/root/#{project.name}-#{project.version}/usr/local/share/doc/#{project.name}",
+         "mv $(tempdir)/osx/build/root/#{project.name}-#{project.version}/bill-of-materials $(tempdir)/osx/build/root/#{project.name}-#{project.version}/usr/local/share/doc/#{project.name}/bill-of-materials",
 
          # Package the project
          "(cd $(tempdir)/osx/build/; #{@pkgbuild} --root root/#{project.name}-#{project.version} \
--- a/templates/osx/preinstall.erb
+++ b/templates/osx/preinstall.erb
@@ -3,11 +3,15 @@
 # If we appear to be in an upgrade unload services.
 
 foundpkg=`/usr/sbin/pkgutil --volume "$3" --pkgs=<%= @identifier-%>.<%= @name -%>`
+oldbom="/usr/share/doc/<%= @name %>/bill-of-materials"
 
 if [ -n "$foundpkg" ]; then
 <%- get_services.each do |service| -%>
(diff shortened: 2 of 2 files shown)


stahnma:
I think I'd prefer we put the doc dir in /opt/puppetlabs/puppet/share rather than in /usr/local. Not sure what others think...

heathseals:
@stahnma it seems that using a docdir in project.prefix for all platforms would make the most sense, but went with /usr/local to be more consistent with the other platforms.  any particular reason we're not using docdir in project.prefix?

shrug:
https://github.com/puppetlabs/puppet-specifications/blob/master/file_paths.md

We should be, I think if we're not it's a bug

stahnma:
@shrug was there something in the specification that we should be looking at there? I didn't see a docdir mentioned in it (but maybe my CMD+F has failed me). 

I'd say let's make one project wide.

haus:
There's no guarantee that there will be a docdir setting or a prefix setting in vanagon. vanagon doesn't currently have a concept of required settings.

haus:
I think changing bill-of-materials to take a target directory via a DSL method would make more sense here, rather than trying to enforce setting existence.

shrug:
The spec refers to /opt/puppetlabs/puppet/share, but nothing higher level

stahnma:
@haus does that mean it could have a default, like the one we want?

haus:
@stahnma the dsl method could have a default, but i don't think defaulting to a value that isn't guaranteed to exist is a good idea.

shrug:
Note: on the server side we use /opt/puppetlabs/server/share

stahnma:
Should we block this PR? We need to get El Cap support in ASAP.

haus:
This seems fine to me for now, we should ticket the other thing if we want the BOM to move or be user-configurable.

heathseals:
@haus i've created a ticket to add support for the BOM to take a target directory.

haus:
:+1: from me. @heathseals what happens during package upgrades? i.e. does the old BOM stay behind or get cleaned up?

stahnma:
Upgrade scenario I guess would be only on 10.10 and 10.9.

haus:
yup

heathseals:
@haus @stahnma i'll submit a PR to puppet-agent to clean up the BOM if the package is in an upgrade and the BOM exists in /usr.

heathseals:
the BOM cleanup actually needs to happen in the preinstall actions. i've updated the PR.

```
letunxav59szxb9:~ root# hdiutil attach puppet-agent-1.2.4-1.osx10.10.dmg
letunxav59szxb9:~ root# installer -target / -pkg /Volumes/puppet-agent-1.2.4-1.yosemite/puppet-agent-1.2.4-1-installer.pkg
installer: Package name is puppet-agent
installer: Installing at base path /
installer: The install was successful.
letunxav59szxb9:~ root# find / -name bill-of-materials
/usr/share/doc/puppet-agent/bill-of-materials
letunxav59szxb9:~ root# /opt/puppetlabs/bin/puppet --version
4.2.2
letunxav59szxb9:~ root# hdiutil detach /Volumes/puppet-agent-1.2.4-1.yosemite
letunxav59szxb9:~ root# hdiutil attach puppet-agent-1.2.6.58.g4968074-1.osx10.10.dmg
letunxav59szxb9:~ root# installer -target / -pkg /Volumes/puppet-agent-1.2.6.58.g4968074/puppet-agent-1.2.6.58.g4968074-1-installer.pkg
installer: Package name is puppet-agent
installer: Upgrading at base path /
installer: The upgrade was successful.
letunxav59szxb9:~ root# hdiutil detach /Volumes/puppet-agent-1.2.6.58.g4968074/
letunxav59szxb9:~ root# /opt/puppetlabs/bin/puppet --version
4.2.3
letunxav59szxb9:~ root# find / -name bill-of-materials
/usr/local/share/doc/puppet-agent/bill-of-materials
letunxav59szxb9:~ root#
```

haus:
:+1: still

PR #191, opened 2015-10-12
(maint) Finish build deps before configuring dependents

nicklewis:
Previously, a build_requires step would only cause the final, top-level
target of the dependent to have a dependency. This means that it's
possible for the entire dependent to be built before the dependency,
entirely defeating the purpose of build dependencies.

Now, the dependent-configure target depends on the dependency target,
which ensures that include directories, etc will be properly found
when the dependent is configured.

### DIFF
--- a/templates/Makefile.erb
+++ b/templates/Makefile.erb
@@ -49,7 +49,7 @@ cleanup-components: <%= @components.map {|comp| "#{comp.name}-cleanup" }.join(" 
 	touch <%= @name %>-project
 
 <%- @components.each do |comp| -%>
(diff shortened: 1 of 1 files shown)


mhaskel:
@nicklewis this looks reasonable to me, have you tested it / did things not explode? What does the new makefile look like?

haus:
Love this.

nicklewis:
I did a build of `puppet-agent` with this change and it didn't break anything.

haus:
that's probably the best test we have here

haus:
:+1:

mhaskel:
:+1:

PR #192, opened 2015-10-13
(maint) yum_repo - only install curl if missing

shermdog:
When using `yum_repo` only attempt to install curl if the package is not already present.  The following failure was seen on platforms that do not have any yum repos configured by default.

```
There are no enabled repos.
 Run "yum repolist all" to see the repos you have.
 You can enable repos with yum-config-manager --enable <repo>
```

### DIFF
--- a/lib/vanagon/platform/dsl.rb
+++ b/lib/vanagon/platform/dsl.rb
@@ -231,7 +231,7 @@ class Vanagon
       def yum_repo(definition)
         definition = URI.parse definition
 
-        self.provision_with "yum -y install curl"
+        self.provision_with "rpm -q curl || yum -y install curl"
         if definition.scheme =~ /^(http|ftp)/
           if File.extname(definition.path) == '.rpm'
             # repo definition is an rpm (like puppetlabs-release)


haus:
Looks sane to me. Have you tried it on systems other than cisco?

shermdog:
Built tools and agent for Cisco and Fedora 21.

haus:
@shermdog were you able to successfully use a yum_repo thingy in the platform config for cisco?

shermdog:
@haus I sure was!

lib/vanagon/platform/dsl.rb
```diff
@@ -231,7 +231,7 @@ def apt_repo(definition, gpg_key = nil)
       def yum_repo(definition)
         definition = URI.parse definition
 
-        self.provision_with "yum -y install curl"
+        self.provision_with "rpm -q curl || yum -y install curl"
```
mckern: Might want to redirect `STDOUT` here. Probably not dangerous, but line noise:

```
[root@vt4eluvdtc1ufnt ~]# rpm -q curl || echo "nope"
curl-7.12.1-17.el4
curl-7.12.1-17.el4
```

shermdog: The existing output is already the same level of noise

```
[root@ggizkxxnv9vxjdo ~]# yum -y install curl || echo "nope"
Loaded plugins: fastestmirror
Loading mirror speeds from cached hostfile
Setting up Install Process
Package curl-7.19.7-40.el6_6.4.x86_64 already installed and latest version
Nothing to do
```

mhaskel:
:+1:

lib/vanagon/platform/dsl.rb
mckern: Sure, but clean up as we go?

```ruby
self.provision_with "rpm -q curl >/dev/null || yum -y install curl"
```

mckern:
:+1: one nitpick, will merge in the morning if no one else cares.

lib/vanagon/platform/dsl.rb
stahnma: FWIW, I like the >/dev/null as well.

PR #194, opened 2015-10-14
(CPR-240) Ensure AIX doesn't include newlines when creating puppet subsystem

whopper:
Previously, when reading in the puppet and mcollective service
scripts to create their subsystems on AIX, a newline in the string
was causing the command and its arguments to appear on separate lines.
This caused `mksys` to not set signorm and sigforce, rendering
the service unstoppable.

We now chomp the newline to ensure the arguments are included with
the service command.


stahnma:
Wow, not sure how I missed that earlier. Thanks for the fixup. :+1: pending tests.

haus:
:+1:

PR #196, opened 2015-10-16
(CPR-242) Don't assume all smf services are top level

haus:
Previously the ips manifest made the naive assumption that all services
could live at /var/svc/manifest or /lib/svc/manifest. Solaris 11 won't
detect services when there, and unfortunately, when moving the services
into subdirs of those directories, the manifest did not previously drop
the subdirectory from the package. This commit updates the manifest to
drop all of the directories holding smf definitions and also to restart
the manifest-import service for each service.


whopper:
:+1: we verified this works yesterday!

PR #198, opened 2015-10-16
(maint) Update links in README.md to point to github

haus:
When viewed from github, the README.md links resolve correctly, but when
viewed in generated docs, they do not. This commit fully qualifies the
links so that they work in both cases.


shrug:
:+1:

PR #200, opened 2015-10-16
(maint) Fix user management arguments

shrug:
Currently the arguments passed to useradd vs usermod are
backwards. This swaps the arguments so the correct set
is passed to the usermod/useradd commands.


stahnma:
:+1:

PR #202, opened 2015-10-19
(maint) Remove nxos platform from vanagon

haus:
NXOS was long ago renamed to cisco-wrlinux, so this commit goes through
and removes the nxos specific methods, tests and logic. All repos that
were exercising nxos platforms have been updated to remove that usage.

### DIFF
--- a/spec/lib/vanagon/platform/dsl_spec.rb
+++ b/spec/lib/vanagon/platform/dsl_spec.rb
@@ -5,7 +5,7 @@ describe 'Vanagon::Platform::DSL' do
   let (:el_5_platform_block)   { "platform 'el-5-fixture'        do |plat| end" }
   let (:el_6_platform_block)   { "platform 'el-6-fixture'        do |plat| end" }
   let (:sles_platform_block)   { "platform 'sles-test-fixture'   do |plat| end" }
-  let (:nxos_5_platform_block) { "platform 'nxos-5-fixture'      do |plat| end" }
+  let (:cicso_wrlinux_platform_block) { "platform 'cisco-wrlinux-fixture'      do |plat| end" }
   let (:solaris_10_platform_block) { "platform 'solaris-10-fixture'      do |plat| end" }
   let (:solaris_11_platform_block) { "platform 'solaris-11-fixture'      do |plat| end" }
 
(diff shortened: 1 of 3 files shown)


ScottGarman:
:+1:

spec/lib/vanagon/platform/dsl_spec.rb
```diff
@@ -16,7 +16,7 @@
   let(:el_definition_rpm) { "http://builds.delivery.puppetlabs.net/puppet-agent/0.2.1/repo_configs/rpm/pl-puppet-agent-0.2.1-release.rpm" }
   let(:sles_definition) { "http://builds.delivery.puppetlabs.net/puppet-agent/0.2.2/repo_configs/rpm/pl-puppet-agent-0.2.2-sles-12-x86_64" }
   let(:sles_definition_rpm) { "http://builds.delivery.puppetlabs.net/puppet-agent/0.2.1/repo_configs/rpm/pl-puppet-agent-0.2.1-release.rpm" }
-  let(:nxos_definition) { "http://builds.delivery.puppetlabs.net/puppet-agent/0.2.1/repo_configs/rpm/pl-puppet-agent-0.2.1-nxos-5-x86_64.repo" }
+  let(:cisco_wrlinux_definition) { "http://builds.delivery.puppetlabs.net/puppet-agent/0.2.1/repo_configs/rpm/pl-puppet-agent-0.2.1-cisco-wrlinux-5-x86_64.repo" }
```
shermdog: We have cisco-wrlinux-5 and cisco-wrlinux-7.  Should there be some sort of differentiation here?  (Although I'm not certain we use this as the target is invalid).

haus: If the repo/rpm behavior differs, then we should test both. If we expect them to be the same, then I'm inclined to leave it as it.

haus: I'm happy to remove the -5 though if that seems clearer.

stahnma: Just a comment that it tests both is good. Behavior is the same from what I can tell.

haus: I added a comment to the test block below (Line 57)

PR #203, opened 2015-10-20
(RE-5565) Resolve PerceivedComplexity rubocop violations for more maintainable code

haus:
This PR refactors 4 problematic methods that were triggering PerceivedComplexity[1] cops. The methods in question were deeply nested with a lot of conditional logic. The refactors took a few paths to alleviate this: removing unneeded conditionals and refactoring behavior into smaller methods. Some of the refactoring included adding new subclasses for AIX and SLES. The final commit reenables the PerceivedComplexity cop in vanagon.

[1] - http://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Metrics/PerceivedComplexity

### DIFF
--- a/lib/vanagon/engine/pooler.rb
+++ b/lib/vanagon/engine/pooler.rb
@@ -37,7 +37,7 @@ class Vanagon
           '{"' + @platform.vmpooler_template + '":"1"}',
           { 'X-AUTH-TOKEN' => @token }
         )
-        if response and response["ok"]
+        if response["ok"]
           @target = response[@platform.vmpooler_template]['hostname'] + '.' + response['domain']
           Vanagon::Driver.logger.info "Reserving #{@target} (#{@platform.vmpooler_template}) [#{@token ? 'token used' : 'no token used'}]"
 
--- a/lib/vanagon/engine/pooler.rb
+++ b/lib/vanagon/engine/pooler.rb
@@ -49,7 +49,7 @@ class Vanagon
             }
           }
 
-          response_tag = Vanagon::Utilities.http_request(
+          Vanagon::Utilities.http_request(
             "#{@pooler}/vm/#{response[@platform.vmpooler_template]['hostname']}",
             'PUT',
             tags.to_json,
--- a/lib/vanagon/platform/dsl.rb
+++ b/lib/vanagon/platform/dsl.rb
@@ -1,5 +1,6 @@
 require 'vanagon/platform/deb'
 require 'vanagon/platform/rpm'
+require 'vanagon/platform/rpm/aix'
 require 'vanagon/platform/rpm/wrl'
 require 'vanagon/platform/swix'
 require 'vanagon/platform/osx'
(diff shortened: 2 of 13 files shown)


haus:
http://jenkins-staging.delivery.puppetlabs.net/job/release_vanagon_intn-sys_master/68/ green build against changes.

lib/vanagon/engine/pooler.rb
```diff
@@ -37,7 +37,7 @@ def select_target
           '{"' + @platform.vmpooler_template + '":"1"}',
           { 'X-AUTH-TOKEN' => @token }
         )
-        if response and response["ok"]
+        if response["ok"]
```
stahnma: Will this give nil::nillclass errors if response isn't there? Can that happen?

haus: http_request could return an empty hash, but will never return nil. http_request would raise an error in the nil case. indexing into a non-existent key on a hash returns nil.

lib/vanagon/platform/dsl.rb
```diff
@@ -198,83 +205,25 @@ def codename(name)
       #
       # @param definition [String] the repo setup file, must be a valid uri, fetched with curl
       # @param gpg_key [String] optional gpg key to be fetched via curl and installed
+      # @deprecated Please use the add_build_repository DSL method instead. apt_repo will be removed in a future vanagon release.
       def apt_repo(definition, gpg_key = nil)
-        # i.e., definition = http://builds.delivery.puppetlabs.net/puppet-agent/0.2.1/repo_configs/deb/pl-puppet-agent-0.2.1-wheezy.list
```
haus: @haus add some warnings here.

stahnma: As discussed we should warn for deprecation.

haus:
@stahnma warnings added.

stahnma:
:+1:

stahnma:
Did we try AIX with it yet?

haus:
nope

lib/vanagon/platform/rpm.rb
```diff
@@ -48,10 +48,29 @@ def rpm_defines
...
+
+      def add_repository(definition)
+        definition = URI.parse(definition)
+
+        commands = ["rpm -q curl > /dev/null || yum -y install curl"]
+        if definition.scheme =~ /^(http|ftp)/
+          if File.extname(definition.path) == '.rpm'
+            # repo definition is an rpm (like puppetlabs-release)
+            commands << "curl -o local.rpm '#{definition}'; rpm -Uvh local.rpm; rm -f local.rpm"
+          else
```
melissa: should we also be installing a gpg key in this case, like we are for deb? It's probably not something for this PR, but something to think about

melissa:
the logic seems solid to me.

I'm :+1: as long as it blends

lib/vanagon/platform/rpm.rb
haus: For rpms, they can list a gpg key in the repo file and yum will automatically fetch it and ask on the first use of the repo.

melissa: Ah, I didn't realize that. Cool.

haus:
will test on AIX this am

haus:
AIX seems to be going fine so far. will know more once this build finishes.

haus:
aix works.

melissa:
:+1: wooo

stahnma:
@haus rebase and I'll hit a button.

PR #204, opened 2015-10-20
DO NOT MERGE ARISTA-43 - EOS AIO fixes

ScottGarman:
EOS has a persistent storage area in /persist/sys that we need
to use for storing our config and cert files. This creates and
packages the /etc/puppetlabs symlink to that area.

On EOS we run the agent as root:root, not as the puppet user.

### DIFF
--- a/templates/rpm/project.spec.erb
+++ b/templates/rpm/project.spec.erb
@@ -90,6 +90,11 @@ install -d %{buildroot}
   cp -p <%= file %> %{buildroot}/<%= File.dirname(file) %>
 <%- end -%>
 
+<%- if @platform.is_eos? -%>
+  # Symlink /etc/puppetlabs to the persistent storage area
+  ln -s /persist/sys/etc/puppetlabs %{buildroot}/etc/puppetlabs
+<%- end -%>
+
 <%- if @platform.is_cisco_wrlinux? || @platform.is_huaweios? -%>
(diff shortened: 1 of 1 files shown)


templates/rpm/project.spec.erb
```diff
@@ -90,6 +90,11 @@ install -d %{buildroot}
   cp -p <%= file %> %{buildroot}/<%= File.dirname(file) %>
 <%- end -%>
 
+<%- if @platform.is_eos? -%>
+  # Symlink /etc/puppetlabs to the persistent storage area
+  ln -s /persist/sys/etc/puppetlabs %{buildroot}/etc/puppetlabs
```
haus: This linking should happen in puppet-agent via a component link call.

templates/rpm/project.spec.erb
```diff
@@ -152,6 +157,12 @@ done
   <%- end -%>
 <%- end -%>
 
+<%- if @platform.is_eos? -%>
+  # Explicity configure puppet to run as root:root
+  /opt/puppetlabs/bin/puppet config set user root
```
haus: Puppet should be running as root by default on eos in puppet 4.

haus:
@ScottGarman I don't think these changes belong in vanagon. They are all puppet-agent specific, so should probably happen in puppet-agent or one of its components.

PR #189, opened 2015-10-07
Vanagon should be able to appropriately add in packaging scripts

melissa:

PR #207, opened 2015-10-21
(CPR-234,CPR-235) Vanagon should be able to appropriately add in packaging scripts

underscorgan:


underscorgan:
See #189

lib/vanagon/project.rb
```diff
@@ -110,18 +110,65 @@ def get_provides
...
+    # @param pkg_state [String] the package state we want to run the given scripts for.
+    #   Can be one or more of 'removal' or 'upgrade'
+    # @return [Array] array of Bourne shell compatible scriptlets to execute during the postremove
+    #   phase of packaging during the state of the system defined by pkg_state (either removal or upgrade)
+    def get_postremove_actions(pkg_state)
+      scripts = []
+      @components.map(&:postremove_actions).flatten.compact.select { |s| s.pkg_state.include? pkg_state }.each do |action|
+        scripts << action.script
+      end
+      scripts
```
haus: Is there any case where the scripts are used individually? I notice there are `join("\n")` in the deb and rpm sections. Is there a reason not to just return the joined string here?

melissa: None that I can think of.

underscorgan: The only thing this might be weird is for posttrans when we need to get install and and upgrade steps and uniq them.

lib/vanagon/component/dsl.rb
```diff
@@ -289,18 +289,56 @@ def environment(env)
...
-      # Adds actions to run at the beginning of a package install
+      # Adds action to run during the preinstall phase of packaging
       #
-      # @param action [String] Bourne shell compatible scriptlets to execute
-      def add_preinstall_action(action)
-        @component.preinstall_actions << action
+      # @param pkg_state [Array] the state in which the scripts should execute. Can be
+      #   one or multiple of 'install' and 'upgrade'.
+      # @param block [Array] the Bourne shell compatible scriptlet(s) to execute
+      def add_preinstall_action(pkg_state, block)
```
haus: I think calling the parameter block is going to be confusing later if it isn't expected to actually be a block.

lib/vanagon/component/dsl.rb
```diff
@@ -289,18 +289,56 @@ def environment(env)
...
+        pkg_state = Array(pkg_state)
+        block = Array(block)
+
+        raise Vanagon::Error, "#{pkg_state} should be an array containing one or more of ['install', 'upgrade']" unless !pkg_state.empty? && (pkg_state - ["install", "upgrade"]).empty?
```
haus: This is hard to read as one long line, and with unless immediately followed by `!`. I'd recommend trying if instead of unless and moving the raise inside of an if conditional.

melissa:
:+1: testing this by building razor packages works. The packaging scripts are populated correctly for both rpm and deb packaging. 

Just so people have a visualization, here's an example of this being used:
```
  pkg.add_preremove_action ['upgrade', 'removal'],
    [
    "source /etc/razor/razor-torquebox.sh ||:",
    "/opt/razor-torquebox/jruby/bin/torquebox undeploy /opt/razor ||:"
    ]
```

haus:
@melissa seems like the commit messages need a once over for content and polish

underscorgan:
Closed in favor of #209

PR #208, opened 2015-10-22
(maint) Allow the rpm spec scriptlets to be run with bash, not sh

melissa:
There are some issues, like creating the homedir for a new user, that sh
fails out when trying to do. Bash is a more stable way to run these
scriptles.

### DIFF
--- a/lib/vanagon/platform/rpm.rb
+++ b/lib/vanagon/platform/rpm.rb
@@ -84,6 +84,7 @@ class Vanagon
         @patch ||= "/usr/bin/patch"
         @num_cores ||= "/bin/grep -c 'processor' /proc/cpuinfo"
         @rpmbuild ||= "/usr/bin/rpmbuild"
+        @bash = "/bin/bash"
         super(name)
       end
     end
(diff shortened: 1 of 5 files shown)


melissa:
@haus I think I'm doing something wrong with this... thoughts?

templates/rpm/project.spec.erb
```diff
@@ -114,7 +114,7 @@ for entry in `cat %{SOURCE1}`; do
   fi
 done
 
-%pre
+%pre -p <%= @bash %>
```
haus: `@platform.bash`

lib/vanagon/platform/rpm.rb
```diff
@@ -69,6 +69,7 @@ def initialize(name)
           @install = "/opt/freeware/bin/install"
         end
         @rpmbuild = "/usr/bin/rpmbuild"
+        @bash = "/bin/bash"
```
haus: Where does bash live on various platforms?

stahnma: Not in /bin on AIX.

melissa: I want to wait until Haus' other pull request refactoring this is merged, and then we'll have separate definitions for aix, rpm, etc.

melissa:
This may not actually be needed...

PR #209, opened 2015-10-22
(CPR-234,CPR-235) Vanagon should be able to appropriately add in packaging scripts

melissa:


lib/vanagon/component/dsl.rb
```diff
@@ -289,18 +289,64 @@ def environment(env)
...
-      # Adds actions to run at the beginning of a package install
+      # Adds action to run during the preinstall phase of packaging
       #
-      # @param action [String] Bourne shell compatible scriptlets to execute
-      def add_preinstall_action(action)
-        @component.preinstall_actions << action
+      # @param pkg_state [Array] the state in which the scripts should execute. Can be
+      #   one or multiple of 'install' and 'upgrade'.
+      # @param scripts [Array] the Bourne shell compatible scriptlet(s) to execute
+      def add_preinstall_action(pkg_state, scripts)
```
haus: This change breaks the method interface for any users of the method previously ( I think this was just @shrug 's change). This means that we would need to tightly coordinate the vanagon release with a code update to all 3 puppet agent branches to avoid failure. This seems undesirable.

haus: Unless the next vanagon release were 0.4.0, which might be fine.

melissa: what if we make pkg_state optional? That should allow us to avoid breaking the current way this is done

haus: pkg_state would need to be the second param, and be allowed to be nil, and if nil the action would need to be applied either unconditionally or just for upgrades.

haus: going to 0.4.0 here might be the best option, even though it will need coordination.

templates/rpm/project.spec.erb
```diff
@@ -175,6 +228,23 @@ done
 <%- end -%>
 
 %preun
+# Run pre-uninstall scripts on upgrade if defined
+<%- scripts = get_preremove_actions("upgrade") -%>
```
haus: What if get_remove_actions returned ':' if it were empty? then this wouldn't need to care if scripts were empty, it could always add scripts.

lib/vanagon/component/dsl.rb
melissa: So, should we try to change this or just leave it as is?

haus: not sure. does anyone else want to weigh in here? /cc @shrug @underscorgan @mckern

templates/rpm/project.spec.erb
melissa: What's more obnoxious, empty if statements, or comments that refer to nothing?

haus: Not sure. We already have comments that refer to nothing in some cases if services aren't delivered in a package.

melissa: That's the choice, comments that refer to nothing, or comments that refer to empty if statements. What's more obnoxious? Or, maybe, what's the least obnoxious?

lib/vanagon/component/dsl.rb
underscorgan: Hm, I'm wondering if we can keep the old method in place and make it deprecated and call this instead? if not I'm :+1: for 0.4.0

haus: @underscorgan i like that, but the names are the same right now.

underscorgan: @haus argh forgot ruby wasn't happy with multiple functions with the same name and different signatures :|

underscorgan: Since 0.4.0 is going to happen soon anyways let's just not worry about this.

melissa:
@haus @underscorgan I updated this so we always include the if statements in the templates, and return `:` if there are no scripts to include.

melissa:
```
$ rpm -qp --scripts output/el/7/PC1/x86_64/razor-server-9.9.9.5.g1cd1cfd-1.el7.x86_64.rpm
preinstall scriptlet (using /bin/sh):
# Save state so we know later if this is an upgrade or an install
mkdir -p /var/lib/rpm-state/razor-server
if [ $1 -eq 1 ] ; then
  touch /var/lib/rpm-state/razor-server/install
fi
if [ $1 -eq 2 ] ; then
  touch /var/lib/rpm-state/razor-server/upgrade
fi

# Add our user and group
if getent group 'razor' > /dev/null 2>&1; then
  /usr/sbin/groupmod 'razor'
else
  /usr/sbin/groupadd --system 'razor'
fi

if getent passwd 'razor' > /dev/null 2>&1; then
  /usr/sbin/usermod --gid 'razor' --shell '/bin/bash' --home '/var/lib/razor' 'razor'
else
  /usr/sbin/useradd --system --gid 'razor' --shell '/bin/bash' --home '/var/lib/razor' 'razor'
fi


# Run preinstall scripts on install if defined
if [ $1 -eq 1 ] ; then
  :
fi

# Run preinstall scripts on upgrade if defined
if [ $1 -eq 2 ] ; then
  :
fi
postinstall scriptlet (using /bin/sh):
  # switch based on systemd vs systemv vs smf vs aix
  #

if [ $1 -eq 1 ] ; then
        # Initial installation
        /usr/bin/systemctl preset razor-server.service >/dev/null 2>&1 || :
fi
preuninstall scriptlet (using /bin/sh):
# Run pre-uninstall scripts on upgrade if defined
if [ $1 -eq 1 ] ; then
  source /etc/razor/razor-torquebox.sh ||:
/opt/razor-torquebox/jruby/bin/torquebox undeploy /opt/razor ||:
fi

# Run pre-uninstall scripts on removal if defined
if [ $1 -eq 0 ] ; then
  source /etc/razor/razor-torquebox.sh ||:
/opt/razor-torquebox/jruby/bin/torquebox undeploy /opt/razor ||:
fi


if [ $1 -eq 0 ] ; then
        # Package removal, not upgrade
        /usr/bin/systemctl --no-reload disable razor-server.service > /dev/null 2>&1 || :
        /usr/bin/systemctl stop razor-server.service > /dev/null 2>&1 || :
fi
postuninstall scriptlet (using /bin/sh):
# Run post-uninstall scripts on upgrade if defined
if [ $1 -eq 1 ] ; then
  :
fi

# Run post-uninstall scripts on removal if defined
if [ $1 -eq 0 ] ; then
  :
fi

  # switch based on systemd vs systemv vs smf vs aix
  #

/usr/bin/systemctl daemon-reload >/dev/null 2>&1 || :
if [ $1 -ge 1 ] ; then
        # Package upgrade, not uninstall
        /usr/bin/systemctl try-restart razor-server.service >/dev/null 2>&1 || :
fi
posttrans scriptlet (using /bin/sh):
# Run post-transaction scripts on install if defined
if [ -e /var/lib/rpm-state/razor-server/install ] ; then
  /bin/chown -R razor:razor /var/lib/razor || :
/bin/chown -R razor:razor /var/log/razor-server || :
source /etc/razor/razor-torquebox.sh
/opt/razor-torquebox/jruby/bin/torquebox deploy /opt/razor --env=production
/bin/chown -R razor:razor /var/lib/razor-torquebox
/bin/chown -R razor:razor /opt/razor-torquebox
  rm /var/lib/rpm-state/razor-server/install
fi

# Run post-transaction scripts on upgrade if defined
if [ -e /var/lib/rpm-state/razor-server/upgrade ] ; then
  /bin/chown -R razor:razor /var/lib/razor || :
/bin/chown -R razor:razor /var/log/razor-server || :
/bin/chown -R razor:razor /var/lib/razor-torquebox
/bin/chown -R razor:razor /opt/razor-torquebox
  rm /var/lib/rpm-state/razor-server/upgrade
fi
```

for the following script definitions in the razor-server and razor-torquebox components.
```
  pkg.add_postinstall_action ['install', 'upgrade'],
    [
      "/bin/chown -R razor:razor /var/lib/razor-torquebox",
      "/bin/chown -R razor:razor /opt/razor-torquebox",
    ]
  pkg.add_postinstall_action ['install', 'upgrade'],
    [
      "/bin/chown -R razor:razor /var/lib/razor || :",
      "/bin/chown -R razor:razor /var/log/razor-server || :"
    ]

  if platform.servicetype == "sysv"
    pkg.add_postinstall_action ['install', 'upgrade'],
      [
        "/bin/chown -R razor:razor /var/run/razor-server || :"
      ]
  end

  pkg.add_postinstall_action ['install'],
    [
      "source /etc/razor/razor-torquebox.sh",
      "/opt/razor-torquebox/jruby/bin/torquebox deploy /opt/razor --env=production"
    ]


  pkg.add_preremove_action ['upgrade', 'removal'],
    [
      "source /etc/razor/razor-torquebox.sh ||:",
      "/opt/razor-torquebox/jruby/bin/torquebox undeploy /opt/razor ||:"
    ]
```

melissa:
It even works when I pass this in:
```
   pkg.add_postinstall_action ['install', 'upgrade'],
     [
       "/bin/chown -R razor:razor /var/lib/razor-torquebox",
       "/bin/chown -R razor:razor /opt/razor-torquebox",
+      "if [ $1 -eq 1 ] ; then
+        echo 'this is a thing'
+       fi"
     ]
```

which gives me 
```
posttrans scriptlet (using /bin/sh):
# Run post-transaction scripts on install if defined
if [ -e /var/lib/rpm-state/razor-server/install ] ; then
  /bin/chown -R razor:razor /var/lib/razor || :
/bin/chown -R razor:razor /var/log/razor-server || :
source /etc/razor/razor-torquebox.sh
/opt/razor-torquebox/jruby/bin/torquebox deploy /opt/razor --env=production
/bin/chown -R razor:razor /var/lib/razor-torquebox
/bin/chown -R razor:razor /opt/razor-torquebox
if [ $1 -eq 1 ] ; then
        echo 'this is a thing'
       fi
  rm /var/lib/rpm-state/razor-server/install
fi

# Run post-transaction scripts on upgrade if defined
if [ -e /var/lib/rpm-state/razor-server/upgrade ] ; then
  /bin/chown -R razor:razor /var/lib/razor || :
/bin/chown -R razor:razor /var/log/razor-server || :
/bin/chown -R razor:razor /var/lib/razor-torquebox
/bin/chown -R razor:razor /opt/razor-torquebox
if [ $1 -eq 1 ] ; then
        echo 'this is a thing'
       fi
  rm /var/lib/rpm-state/razor-server/upgrade
fi
```

lib/vanagon/project.rb
```diff
@@ -110,18 +110,81 @@ def get_provides
...
-    # @return [Array] array of Bourne shell compatible scriptlets to execute
-    def get_preinstall_actions
-      @components.map(&:preinstall_actions).flatten
+    # @param pkg_state [String] the package state we want to run the given scripts for.
+    #   Can be one or more of 'install' or 'upgrade'
+    # @return [Array] array of Bourne shell compatible scriptlets to execute during the preinstall
+    #   phase of packaging during the state of the system defined by pkg_state (either install or upgrade)
+    def get_preinstall_actions(pkg_state)
+      scripts = []
+      @components.map(&:preinstall_actions).flatten.compact.select { |s| s.pkg_state.include? pkg_state }.each do |action|
```
haus: What is the compact needed for? Were you getting nil entries?

melissa: `copy => paste`

haus: Also, i think this can just be:
`@components.map(&:preinstall_actions).flatten.compact.select { |s| s.pkg_state.include? pkg_state }.map(&:action).join("\n")`

haus: maybe...haven't tested.

lib/vanagon/project.rb
```diff
@@ -110,18 +110,69 @@ def get_provides
...
     #
-    # @return [Array] array of Bourne shell compatible scriptlets to execute
-    def get_preinstall_actions
-      @components.map(&:preinstall_actions).flatten
+    # @param pkg_state [String] the package state we want to run the given scripts for.
+    #   Can be one or more of 'install' or 'upgrade'
+    # @return [Array] array of Bourne shell compatible scriptlets to execute during the preinstall
+    #   phase of packaging during the state of the system defined by pkg_state (either install or upgrade)
+    def get_preinstall_actions(pkg_state)
+      scripts = @components.map(&:preinstall_actions).flatten.compact.select { |s| s.pkg_state.include? pkg_state }.map(&:scripts).join("\n")
```
underscorgan: This join should probably be removed.

lib/vanagon/project.rb
```diff
@@ -110,18 +110,69 @@ def get_provides
...
     #
-    # @return [Array] array of Bourne shell compatible scriptlets to execute
-    def get_postinstall_actions
-      @components.map(&:postinstall_actions).flatten
+    # @param pkg_state [String] the package state we want to run the given scripts for.
+    #   Can be one or more of 'removal' or 'upgrade'
+    # @return [Array] array of Bourne shell compatible scriptlets to execute during the preremove
+    #   phase of packaging during the state of the system defined by pkg_state (either removal or upgrade)
+    def get_preremove_actions(pkg_state)
+      scripts = @components.map(&:preremove_actions).flatten.compact.select { |s| s.pkg_state.include? pkg_state }.map(&:scripts).join("\n")
```
underscorgan: This join should probably be removed.

lib/vanagon/project.rb
```diff
@@ -110,18 +110,69 @@ def get_provides
...
+
+    # Collects the postremove packaging actions for the project and it's components
+    # for the specified packaging state
+    #
+    # @param pkg_state [String] the package state we want to run the given scripts for.
+    #   Can be one or more of 'removal' or 'upgrade'
+    # @return [Array] array of Bourne shell compatible scriptlets to execute during the postremove
+    #   phase of packaging during the state of the system defined by pkg_state (either removal or upgrade)
+    def get_postremove_actions(pkg_state)
+      scripts = @components.map(&:postremove_actions).flatten.compact.select { |s| s.pkg_state.include? pkg_state }.map(&:scripts).join("\n")
```
underscorgan: same.

lib/vanagon/component/dsl.rb
stahnma: I still think this should be named preinstall_action instead of add preinstall action. It mimics the way the other methods are named and the way rpm handles these items. (pkg.configure, pkg.build pkg.name, pkg.preinstall_action would be congruent with that).

melissa: There are other methods that begin with `add_${thing}`, so this fits in with how other methods are named... I honestly don't care that much, and if a majority think it's better without the `add_` prefix, I'll remove it

stahnma: The other add methods don't correspond to packaging directly...at least not a 1:1 mapping. I'd rather it not be there.

lib/vanagon/project.rb
```diff
@@ -110,18 +110,69 @@ def get_provides
...
-    # Collects the pre-install actions for the project and it's components
+    # Collects the preinstall packaging actions for the project and it's components
+    # for the specified packaging state
     #
-    # @return [Array] array of Bourne shell compatible scriptlets to execute
-    def get_preinstall_actions
-      @components.map(&:preinstall_actions).flatten
+    # @param pkg_state [String] the package state we want to run the given scripts for.
+    #   Can be one or more of 'install' or 'upgrade'
+    # @return [Array] array of Bourne shell compatible scriptlets to execute during the preinstall
```
haus: Return type is wrong now. It should be String.

templates/deb/postinst.erb
```diff
@@ -44,3 +43,13 @@ fi
     fi
   <%- end -%>
 <%- end -%>
+
+# Run postinstall scripts on install if defined
+if [ $1 = configure ] && [ -z $2] ; then
```
haus: This bash needs some tweaking. The test needs a space between `$2` and `]` to be valid.

haus: Looks like that needs updating in all of the deb scripts.

templates/rpm/project.spec.erb
```diff
@@ -125,12 +130,30 @@ done
...
+
+
+
 %post
 <%- if @platform.is_aix? || (@platform.is_el? && @platform.os_version.to_i == 4) -%>
   ## EL-4 and AIX RPM don't have %posttrans, so we'll put them here
-  <%- get_postinstall_actions.each do |task| -%>
-    <%= task %>
-  <%- end -%>
+  # Run postinstall scripts on install if defined
```
haus: Indentation here is inconsistent with the earlier scripts. Same issue 5 lines down.

templates/rpm/project.spec.erb
```diff
@@ -125,12 +130,30 @@ done
...
+if [ $1 -eq 1 ] ; then
+  <%= get_preinstall_actions("install") %>
+fi
+
+# Run preinstall scripts on upgrade if defined
+if [ $1 -eq 2 ] ; then
+  <%= get_preinstall_actions("upgrade") %>
+fi
+
+
```
haus: Seems like a lot of extra whitespace.

templates/deb/preinst.erb
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+# Run preinstall scripts on install if defined
+if [ $1 = install ] && [ -z $2] ; then
```
haus: Do we care about the case where preinst is called with install and a version?

templates/rpm/project.spec.erb
```diff
@@ -152,7 +175,18 @@ done
   <%- end -%>
 <%- end -%>
 
+
```
melissa: @haus I added more whitespace to annoy you. Also, because it helps my brain to have a little more space between the script macros

haus:
I'm :+1: here. I think adding some project level tests for this would be good someday, but obviously more complicated.

templates/rpm/project.spec.erb
melissa: naw, it's the perfect amount!

lib/vanagon/component/dsl.rb
stahnma: Disregard this, since we already have odd add_ names. Maybe another PR can clean that stuff up.

PR #210, opened 2015-10-22
(CPR-248) Add link to files list in link DSL method

haus:
Previously links were not added to the files list which meant that they
would not be collected if they fell outside of the directories being
watched by vanagon. This commit adds them to the files list.
The cp calls in the spec and rules files need the -R flag added to
ensure that links are not dereferenced and are copied as is, which would
otherwise break packaging.


stahnma:
:+1:

haus:
I need to look at AIX here a bit. -R might not be the right flag there.

haus:
Nope, this works on aix as well. all clear.

melissa:
Will this preserve the links, or just clobber them and turn them into normal files? How does tar handle this?

haus:
-R copies the link without dereferencing. tar archives the link.

PR #211, opened 2015-10-26
(maint) Add minimum ruby version and specification to gemspec

haus:
Because of the use of keyword arguments with defaults, vanagon won't
work on anything less than ruby 2.1.0, so this commit updates the
gemspec to reflect that. We also add the specification version to
enforce a relatively modern version of rubygems.

### DIFF
--- a/vanagon.gemspec
+++ b/vanagon.gemspec
@@ -12,6 +12,8 @@ Gem::Specification.new do |gem|
   gem.authors  = ['Puppet Labs']
   gem.email    = 'info@puppetlabs.com'
   gem.homepage = 'http://github.com/puppetlabs/vanagon'
+  gem.specification_version = 3
+  gem.required_ruby_version = '~> 2.1.0'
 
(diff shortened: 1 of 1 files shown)


vanagon.gemspec
```diff
@@ -12,6 +12,8 @@ Gem::Specification.new do |gem|
   gem.authors  = ['Puppet Labs']
   gem.email    = 'info@puppetlabs.com'
   gem.homepage = 'http://github.com/puppetlabs/vanagon'
+  gem.specification_version = 3
+  gem.required_ruby_version = '~> 2.1.0'
```
underscorgan: Do we want this this tightly bound or do we expect it to work with any 2.x >= 2.1.0?

haus: @underscorgan I don't expect it to break on a 2.x release. it works fine on 2.2, but I'm open to all of the possibilities.

underscorgan: @haus ok, sounds reasonable.

underscorgan:
:+1:

PR #212, opened 2015-10-27
(CPR-250) Ensure directories are created after initial file list

haus:
Previously the file-list-before-build could happen after the base
directories were created, which had the side effect of making it appear
as though those directories were not created by the current project.
This commit addresses that issue by ensuring that the
file-list-before-build is generated before any directories, and as early
as possibly in the Makefile.

### DIFF
--- a/templates/Makefile.erb
+++ b/templates/Makefile.erb
@@ -13,14 +13,14 @@ file-list-before-build:
 <%- if dirnames.empty? -%>
 	touch file-list-before-build
 <%- else -%>
-	(find -L <%= dirnames.join(' ') %> 2>/dev/null || find <%= dirnames.join(' ') %> -follow) | sort | uniq > file-list-before-build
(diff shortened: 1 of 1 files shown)


haus:
My local tests show no changes to rpms or debs, and the following changes to osx:

```
comm -3 osx_against_master ~/osx_against_my_branch 
	./etc/puppetlabs/code/environments/production/hieradata
	./etc/puppetlabs/code/environments/production/manifests
	./etc/puppetlabs/code/environments/production/modules
	./etc/puppetlabs/code/modules
	./opt/puppetlabs/facter
	./opt/puppetlabs/facter/facts.d
	./opt/puppetlabs/mcollective
	./opt/puppetlabs/mcollective/plugins
	./opt/puppetlabs/puppet/cache
	./opt/puppetlabs/puppet/modules
	./var
	./var/log
	./var/log/puppetlabs
	./var/log/puppetlabs/puppet
	./var/run
	./var/run/puppetlabs
```

Those are all empty directories that would be added to the osx payload.

stahnma:
:+1:

PR #213, opened 2015-11-02
(CPR-251) Move /var/run/* to /system/volatile on solaris 11

haus:
On solaris 11, /var/run is a symlink to /system/volatile, which results
in file conflicts when attempting to install a package which declares a
directory under /var/run. This commit addresses that by rewriting the
directory entries under /var/run to move them under /system/volatile
instead. /system/volatile must then be itself dropped to avoid conflicts
on package install.


underscorgan:
:+1:

PR #214, opened 2015-11-02
(CPR-247) Add BOM method to project DSL

haus:
Previously, the bill-of-materials was laid down by default in the docdir
on various platforms. This causes problems on platforms like newer osx,
and isn't very configurable, so this commit adds a method to the project
DSL which allows the location to be overridden. This is a backward
compatible addition, and contains some shims that can be removed when
bill-of-materials is no longer desired to be on by default.


haus:
Note: would love for #213 to get merged first so I can test this out on solaris 11 easier.

stahnma:
:+1:

PR #217, opened 2015-11-05
(RE-5679) Allow specification of which package type to build

melissa:


melissa:
If we have a platform config like 
```
platform "eos-4-i386" do |plat|
  plat.servicedir "/etc/rc.d/init.d"
  plat.defaultdir "/etc/sysconfig"
  plat.servicetype "sysv"

  plat.package_type "rpm"

  plat.yum_repo "http://pl-build-tools.delivery.puppetlabs.net/yum/fedora/14/i386/pl-build-tools-fedora-14.repo"
  plat.yum_repo "http://osmirror.delivery.puppetlabs.net/eos-4-i386/eos-4-i386.repo"
  plat.provision_with "yum install -y --nogpgcheck autoconf automake createrepo rsync gcc make rpm-build rpm-libs yum-utils zip"

  plat.install_build_dependencies_with "yum install -y --nogpgcheck"
  plat.vcloud_name "fedora-14-i386"
end
```
We should be building rpm style packages. This PR adds in the option to define `plat.package_type` in the platform config file. If that package type is valid on that platform, it will build a package in the specified format.

melissa:
Tested successfully!!
with `plat.package_type = 'rpm'`, it built an rpm package
with `plat.package_type = 'swix'`, it built a swix package
with `plat.package_type` not set, it built a swix package

melissa:
This probably isn't the best solution to this problem, but at least it works!

stahnma:
Should it be in the platform definition or in the project definition? It seems like you'd care about what was outputted per project more than the platform, but then again, I could see it both ways.

@haus  thoughts?

melissa:
@stahnma there is more discussion regarding your question in the ticket

melissa:
I was thinking about doing something along the lines of
```
--- a/lib/vanagon/platform/rpm/eos.rb
+++ b/lib/vanagon/platform/rpm/eos.rb
@@ -12,8 +12,10 @@ class Vanagon
         def generate_package(project)
           if project.platform.package_type == 'rpm'
             super
-          else
+          elsif project.platform.package_type == 'swix'
             Vanagon::Platform::RPM::SWIX.generate_package(project)
+          else
+            fail "Don't know how to build package type '#{project.platform.package_type}' on EOS"
           end
         end

@@ -24,8 +26,10 @@ class Vanagon
         def package_name(project)
           if project.platform.package_type == 'rpm'
             super
-          else
+          elsif project.platform.package_type == 'swix'
             Vanagon::Platform::RPM::SWIX.package_name(project)
+          else
+            fail "Don't know how to build package type '#{project.platform.package_type}' on EOS"
           end
         end
       end
```
but I'm doing something wrong. So, I'm looking into it now

haus:
I left some comments in the jira ticket. @melissa what you're trying above won't work because generate_package isn't a class method, it's an instance method so you need an instance of the class to invoke it.

melissa:
@haus so I was hesitant to create a new instance, but might that be worth it? Or do you think what I have is sufficient for the time being?

underscorgan:
@melissa I want to review this more in the morning when I've had more time to think about it, but for right now I don't think it makes sense for the package_type to be specified in the platform. I'd much prefer to see it in the project, though I'm not sure if that is the best implementation. 

I think this is a big change and needs some planning for the best way to actually implement it so we don't end up needing to refactor it near-term. 

I, however, also don't want this to block the windows work, would it be plausible to have multiple platform configs for the different types of windows packaging for a short term fix? This would be a super useful feature I just think current implementation is a little weird (and I'm having a hard time wrapping my head around this completely right now)

melissa:
Yeah... This won't block windows work until we get into the part where we build the MSI. For now we'll be able to just build nuget packages. I was originally thinking about doing it at the project level, and we could easily do that. With this implementation, it would be a very simple cutover. The main problem I was running in to was that I wanted to make the package selection either in the main platform instance (i.e., eos.rb or windows.rb) or in the platform dsl.rb where we're currently doing the selection.

In my head, I really want this to go from the dsl with 
```
    def platform(name, &block)
        @platform = case name
                    when /^aix-/
                      Vanagon::Platform::RPM::AIX.new(@name)
                    when /^(cisco-wrlinux|el|fedora)-/
                      Vanagon::Platform::RPM.new(@name)
                    when /^sles-/
                      Vanagon::Platform::RPM::SLES.new(@name)
                    when /^huaweios-/
                      Vanagon::Platform::RPM::WRL.new(@name)
                    when /^(cumulus|debian|ubuntu)-/
                      Vanagon::Platform::DEB.new(@name)
                    when /^eos-/
                      Vanagon::Platform::RPM::EOS.new(@name)
                    when /^win-/
                      Vanagon::Platform::Windows.new(@name)
```

Then, in the windows instance, we'd be able to programmatically decide which methods we want, based on the package type defined. I played around with it a little, but I was hesitant to create a new instance for swix vs. rpm if we have already created an eos instance (see above comment). Likewise, It would be strange (and probably wrong) to create a windows platform instance as well as an msi or nuget platform instance. The biggest problem was that I don't have access to the `package_type` definition when we call platform (the code above in this comment), so I couldn't tell a good way to structure this :(

Any and all design help is majorly appreciated

stahnma:
after discussion in the ticket, I think it's ok to keep it at platform. I am morbidly curious about what happens if I put in 'deb'.

stahnma:
:+1:

melissa:
@stahnma it will build a swix package. It doesn't care about what you want unless you want an rpm. In all other cases, it'll build swix ;)

stahnma:
@mckern has said he'd like to review again before the merges.

lib/vanagon/platform/rpm/eos.rb
```diff
@@ -0,0 +1,43 @@
...
+        # Method to derive the package name for the project
+        #
+        # @param project [Vanagon::Project] project to name
+        # @return [String] name of the SWIX package for this project
+        def package_name(project)
+          if project.platform.package_type == 'rpm'
+            super
+          else
+            "#{project.name}-#{project.version}-#{project.release}.#{os_name}#{os_version}.#{project.noarch ? 'noarch' : @architecture}.swix"
+          end
```
mckern: Are there any other "if/else" decisions likely to be made for this sort of paradigm in the future? If so, this pattern portends a lot of methods with cascading "if/else" decisions in them; is it worth abstracting this out into a module/mix-in pattern now so that the groundwork is there in the future or are we OK saying "we will cross that bridge when we get there"?

melissa: This should probably be a case statement, but my thought was to leave it with `if/else` until we get more conditions in. At the moment, all of our thinking around this has been for at most two different packaging types on a given system (or both package types for a given system). That second case I haven't given much thought to yet.

melissa:
What if directory structure:
windows.rb
windows/
    | _ msi.rb
    |_ nupkg.rb

And the windows.rb file would include
```
         def generate_package(project)
           case project.platform.package_type
           when 'msi'
               MSI.generate_package(project)
           when 'nuget' || 'nupkg'
               nupkg.generate_package(project)
           else
               fail "Don't know how to build package type '#{project.platform.package_type}' on EOS"
           end
         end
```

so then when we create the platform instance, we could just do
```
    def platform(name, &block)
        @platform = case name
                    when /^win-/
                      Vanagon::Platform::Windows.new(@name)
```
And not worry about what kind of package we're going to make until we make it?

Would that work? I haven't tested this at all, since I haven't had time to yet

underscorgan:
@melissa I like the idea in your last comment, though unsure if we want to actually fail if package_type isn't specified. Seems like for most things some default behavior makes sense. That abstraction seems to make more sense to me though.

stahnma:
Yeah, I don't think we should fail if package type isn't specified, that's a breaking change and probably not worth the headache. Plus, 90%+ of the time, you just want the default.

melissa:
So, I didn't mean to imply fail if not specified. I think we should default to one package type for each system. However, I do think we should fail if the user passes in a package type not specified for the platform (i.e., deb for eos).

stahnma:
Makes sense. Sounds good.

shrug:
I think we're good, needs a rebase

melissa:
This is still just out of my grasp when it comes to implementation :( I'll try to get help this afternoon

lib/vanagon/platform/rpm/eos.rb
```diff
@@ -0,0 +1,83 @@
...
+        # The specific bits used to generate an EOS package for a given project
+        #
+        # @param project [Vanagon::Project] project to build an EOS package of
+        # @return [Array] list of commands required to build the EOS package
+        # for the given project from an rpm or a swix
+        def generate_package(project)
+          # If nothing is passed in as platform type, default to building a swix package
+          if project.platform.package_type.nil? || project.platform.package_type.empty?
+            return generate_swix_package(project)
+          else
```
melissa: this strange format here is for when we eventually specify multiple package types. If we have an array of different package types to build, then we can iterate over that. If it's an empty array, or not even set, then we just build one package, the default. Otherwise, we can iterate over the array. I was going to do that, but then I realized there are issues with the package_name method that I didn't want to solve.

lib/vanagon/platform/rpm/eos.rb
```diff
@@ -0,0 +1,83 @@
...
+
+          commands
+        end
+
+        # Method to derive the package name for the project
+        #
+        # @param project [Vanagon::Project] project to name
+        # @return [String] name of the SWIX package for this project
+        def swix_package_name(project)
+          "#{project.name}-#{project.version}-#{project.release}.#{os_name}#{os_version}.#{project.noarch ? 'noarch' : @architecture}.swix"
```
mckern: I love this string. Much interpolation. So detail.

mckern: Is `@architecture` not going to say `noarch` if that's the case? The ternary operator here is a nitpick but it's something that can be smoothed over later.

mckern:
One nitpick (which may just be a lack of understanding on my part), no blockers from me :+1:

lib/vanagon/platform/rpm/eos.rb
melissa: I don't think so. I think it's going to return the architecture of the host we're building on, not the final architecture of the package. So, I think this needs to be this way. I'm not 100% sure though

PR #220, opened 2015-11-09
(RE-5676) SSH to windows box as Administrator

melissa:


melissa:
@haus sorry to bug you again, but I can't figure out why this is failing. I think it's calling the `target_user` function with no arguments, but it shouldn't be calling that method, right? It should just be checking to see if that variable has been set, and using root if not...

lib/vanagon/engine/base.rb
```diff
@@ -10,7 +10,7 @@ def initialize(platform, target = nil)
         @platform = platform
         @required_attributes = ["ssh_port"]
         @target = target if target
-        @target_user = "root"
+        @target_user = @platform.target_user ? @platform.target_user : "root"
```
haus: It seems like if this is being added to the platform level, then "root" should be added as a default there. then this conditional can just go away.

lib/vanagon/platform/dsl.rb
```diff
@@ -45,6 +46,8 @@ def platform(name, &block)
                       Vanagon::Platform::Solaris10.new(@name)
                     when /^solaris-11/
                       Vanagon::Platform::Solaris11.new(@name)
+                    when /^win-/
+                      Vanagon::Platform::Win.new(@name)
```
haus: Why not use Windows here?

melissa: versus `Win`? No reason. We can change it

lib/vanagon/engine/base.rb
```diff
@@ -10,7 +10,7 @@ def initialize(platform, target = nil)
         @platform = platform
         @required_attributes = ["ssh_port"]
         @target = target if target
-        @target_user = "root"
+        @target_user = @platform.target_user
```
melissa: @haus it seems like this is the line causing issues. It seems like this is calling the method and for some reason passing it in to the vmpooler token check, which is really confusing me. Everything else works just fine, just not tests

```
  1) Vanagon::Engine::Pooler #load_token prefers an env var to a file
     Failure/Error: expect(Vanagon::Engine::Pooler.new(platform).token).to eq('abcd')
       #<Double Vanagon::Platform> received unexpected message :target_user with (no args)
     # ./lib/vanagon/engine/base.rb:13:in `initialize'
     # ./lib/vanagon/engine/pooler.rb:12:in `initialize'
     # ./spec/lib/vanagon/engine/pooler_spec.rb:29:in `new'
     # ./spec/lib/vanagon/engine/pooler_spec.rb:29:in `block (3 levels) in <top (required)>'
```

lib/vanagon/platform/dsl.rb
geoffnichols: +1 on using `Windows` instead of `Win`

lib/vanagon/engine/base.rb
haus: The pooler check creates a new pooler engine instance, which inherits from base. The test needs to be updated to expect this.

lib/vanagon/platform/win.rb
```diff
@@ -0,0 +1,53 @@
...
+        File.join("win", @codename, target_repo)
+      end
+
+      # Constructor. Sets up some defaults for the windows platform and calls the parent constructor
+      #
+      # @param name [String] name of the platform
+      # @return [Vanagon::Platform::DEB] the win derived platform with the given name
+      def initialize(name)
+        super(name)
+        @target_user = "Administrator"
```
haus: You probably want to set this before the super call.

lib/vanagon/platform/windows.rb
```diff
@@ -0,0 +1,53 @@
...
+      #
+      # @param workdir [String] working directory to stage the evaluated templates in
+      # @param name [String] name of the project
+      # @param binding [Binding] binding to use in evaluating the packaging templates
+      def generate_packaging_artifacts(workdir, name, binding)
+        win_dir = File.join(workdir, "windows")
+        FileUtils.mkdir_p(win_dir)
+
+        # Lots of templates here
+        [""].each do |win_file|
```
underscorgan: It looks like this is missing something

lib/vanagon/platform/windows.rb
```diff
@@ -0,0 +1,53 @@
...
+          erb_file(File.join(VANAGON_ROOT, "templates/win/#{win_file}.erb"), File.join(win_dir, win_file), false, { :binding => binding })
+        end
+      end
+
+      # Method to derive the package name for the project
+      #
+      # @param project [Vanagon::Project] project to name
+      # @return [String] name of the windows package for this project
+      def package_name(project)
+        "#{project.name}-#{project.version}-#{project.release}-#{@architecture}.msi"
```
underscorgan: Are noarch projects a thing in windows or will there always be an arch specified?

lib/vanagon/platform/windows.rb
```diff
@@ -0,0 +1,53 @@
...
+      end
+
+      # Get the expected output dir for the windows packages. This allows us to
+      # use some standard tools to ship internally.
+      #
+      # @return [String] relative path to where windows packages should be staged
+      def output_dir(target_repo = "")
```
underscorgan: is this a valid default for the target repo?

lib/vanagon/platform/windows.rb
```diff
@@ -0,0 +1,53 @@
...
+      # @param project [Vanagon::Project] project to name
+      # @return [String] name of the windows package for this project
+      def package_name(project)
+        "#{project.name}-#{project.version}-#{project.release}-#{@architecture}.msi"
+      end
+
+      # Get the expected output dir for the windows packages. This allows us to
+      # use some standard tools to ship internally.
+      #
+      # @return [String] relative path to where windows packages should be staged
```
underscorgan: target_repo isn't commented.

spec/lib/vanagon/engine/pooler_spec.rb
```diff
@@ -1,7 +1,7 @@
 require 'vanagon/engine/pooler'
 
 describe 'Vanagon::Engine::Pooler' do
-  let (:platform) { double(Vanagon::Platform) }
+  let (:platform) { double(Vanagon::Platform, :target_user => 'root') }
```
underscorgan: It looks like target_user is defaulting to root in platform.rb, does it work without this?

spec/lib/vanagon/platform/windows_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+      end
+
+      describe '#target_user' do
+        it "sets the target_user to 'Administrator'" do
+          expect(cur_plat._platform.target_user).to eq('Administrator')
+        end
+      end
+    end
+  end
+end
```
underscorgan: It may be difficult to test the methods you added that require projects (but may be worth it), definitely would be good to have tests for `generate_packaging_artifacts` though.

lib/vanagon/platform/windows.rb
melissa: We have no templates yet

lib/vanagon/platform/windows.rb
melissa: Arch should always be specified

melissa:
I should have mentioned the windows.rb file is mostly fluff. We'll fill it out once we're closer to building packages, but at the moment we just need to have those methods available for testing

spec/lib/vanagon/platform/windows_spec.rb
melissa: To make sure the erb templates are parsed correctly? I'm not sure the best way to do that, especially since we don't have any templates to process yet. I'm happy to try to implement it though

lib/vanagon/platform/windows.rb
melissa: Good catch, I added it :)

lib/vanagon/platform/windows.rb
melissa: Yup! We want the option of setting a different repo, but don't necessarily do it all the time. So, for PC1, we want to be able to pass it in, but if we're defaulting to the original repo setup, we don't want to pass anything in for target_repo

spec/lib/vanagon/platform/windows_spec.rb
```diff
@@ -0,0 +1,43 @@
+require 'vanagon/platform'
+
+describe "Vanagon::Platform::Windows" do
+  platforms =[
+    {
+      :name                   => "win-2012-x64",
```
jcoconnor: Suggest change this to 2012R2

melissa: Done!

jcoconnor:
@melissa Have tested this on the puppet-agent build and it succeeds up to the same point as with the pl-build-tools-vanagon, i.e. the rsync failure due to the lack of a proper make (i.e. work to be done in later tickets).

So this looks good :+1:

lib/vanagon/engine/base.rb
melissa: grr, that was very difficult to track down what the heck `double` was

spec/lib/vanagon/platform/windows_spec.rb
underscorgan: Yeah, once we have templates this will probably make more sense, didn't realize those were expected to be missing :)

melissa: No worries! I should have been more explicit about their absence and that the windows.rb file is currently just a place holder at this point

PR #221, opened 2015-11-09
(maint) Fix minimum Ruby version to allow Ruby 2.x

whopper:
Previously, Vanagon's minimum Ruby version was set to '~> 2.1.0',
which meant that versions equal or greater than 2.1.0 but less than
2.2.0 were allowed. Any version of Ruby above 2.1.0 in the 2.x series
should be valid.


underscorgan:
:+1:

underscorgan:
@whopper I think it should be ok?

haus:
:+1:

PR #222, opened 2015-11-12
(RE-5979) Solaris-11 VERSION file should match other platforms

shrug:
On Solaris 11, the version transformation to an IPS-compatible
format (minus alpha characters) also makes it into the VERSION
file.

This updates the logic to operate on a duplicate object instead
of the actual version variable when making the transformation,
so that the transformed value only winds up where we explicitly
ask for it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ This changelog adheres to [Keep a CHANGELOG](http://keepachangelog.com/).
 
 ## [Unreleased]
 
+## [0.4.1] - 2015-11-12
+### Changed
+- Fixed an issue where, with test builds when the version string contains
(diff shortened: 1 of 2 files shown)


shrug:
Testing this now

shrug:
```
root@tnfzwa2rw20qre5:~# cat /opt/puppetlabs/puppet/VERSION
1.2.7.386.gd29a56d
```

haus:
@shrug wouldn't it be easier to do the dup in the ips_version method? then the callers wouldn't need to care about duping before passing in parameters.

melissa:
:+1:

shrug:
After the latest update, on solaris-11-i386:
```
root@hoax90utc4in717:~# cat /opt/puppetlabs/puppet/VERSION
1.2.7.395.gf0f24b1
```

PR #223, opened 2015-11-12
(RE-5680) Teach vanagon how to build nuget archives on windows

melissa:
These are changes needed to build nuget packages. Once https://github.com/puppetlabs/vanagon/pull/217 is sorted out, we can use the procedure established there to build either msi or nuget packages. That will likely require that the methods added here to windows.rb will have to be moved.

### DIFF
--- a/lib/vanagon/platform/windows.rb
+++ b/lib/vanagon/platform/windows.rb
@@ -52,7 +52,7 @@ class Vanagon
             commands << %(C:/ProgramData/chocolatey/bin/choco.exe source add -n #{definition.host}-#{definition.path.gsub('/', '-')} -s "#{definition}" --debug || echo "Oops, it seems that you don't have chocolatey installed on this system. Please ensure it's there by adding something like 'plat.add_repository 'https://chocolatey.org/install.ps1'' to your platform definition.")
           end
         else
-          fail "Invalid repo specification #{definition}"
+          raise Vanagon::Error "Invalid repo specification #{definition}"
         end
 
         commands
--- a/lib/vanagon/platform/windows.rb
+++ b/lib/vanagon/platform/windows.rb
@@ -74,6 +74,8 @@ class Vanagon
       # @return [Vanagon::Platform::DEB] the win derived platform with the given name
       def initialize(name)
         @target_user = "Administrator"
+        @find = "/usr/bin/find"
+        @sort = "/usr/bin/sort"
         super(name)
       end
     end
--- a/lib/vanagon/platform/windows.rb
+++ b/lib/vanagon/platform/windows.rb
@@ -6,23 +6,17 @@ class Vanagon
       # @param project [Vanagon::Project] project to build a windows package of
       # @return [Array] list of commands required to build a windows package for the given project from a tarball
       def generate_package(project)
-        target_dir = project.repo ? output_dir(project.repo) : output_dir
-        ["mkdir -p output/#{target_dir}",
-        "touch output/#{target_dir}/#{project.name}-#{project.version}-#{project.release}-#{@architecture}.msi"]
-      end
-
-      # Method to generate the files required to build a windows package for the project
-      #
-      # @param workdir [String] working directory to stage the evaluated templates in
-      # @param name [String] name of the project
-      # @param binding [Binding] binding to use in evaluating the packaging templates
-      def generate_packaging_artifacts(workdir, name, binding)
-        win_dir = File.join(workdir, "windows")
-        FileUtils.mkdir_p(win_dir)
-
-        # We'll uncomment this once we have templates to parse here
-        [""].each do |win_file|
-          #erb_file(File.join(VANAGON_ROOT, "templates/win/#{win_file}.erb"), File.join(win_dir, win_file), false, { :binding => binding })
+        # If nothing is passed in as platform type, default to building a nuget package
+        # We should default to building an MSI once that code has been implimented
+        if project.platform.package_type.nil? || project.platform.package_type.empty?
+          return generate_nuget_package(project)
+        else
+          case project.platform.package_type
+          when "nuget", "nupkg"
+            return generate_nuget_package(project)
+          else
+            raise Vanagon::Error "I don't know how to build package type '#{project.platform.package_type}' for Windows. Teach me?"
+          end
         end
       end
 
@@ -31,7 +25,67 @@ class Vanagon
       # @param project [Vanagon::Project] project to name
       # @return [String] name of the windows package for this project
       def package_name(project)
-        "#{project.name}-#{project.version}-#{project.release}-#{@architecture}.msi"
+        # If nothing is passed in as platform type, default to a nuget package
+        # We should default to an MSI once that code has been implimented
+        if project.platform.package_type.nil? || project.platform.package_type.empty?
+          return nuget_package_name(project)
+        else
+          case project.platform.package_type
+          when "nuget", "nupkg"
+            return nuget_package_name(project)
+          else
+            raise Vanagon::Error "I don't know how to name package type '#{project.platform.package_type}' for Windows. Teach me?"
+          end
+        end
+      end
+
+      # Method to generate the files required to build a windows package for the project
+      #
+      # @param workdir [String] working directory to stage the evaluated templates in
+      # @param name [String] name of the project
+      # @param binding [Binding] binding to use in evaluating the packaging templates
+      def generate_packaging_artifacts(workdir, name, binding)
+        # templates that do require a name change
+        erb_file(File.join(VANAGON_ROOT, "templates/windows/project.nuspec.erb"), File.join(workdir, "#{name}.nuspec"), false, { :binding => binding })
+
+        # templates that don't require a name change
+        ["chocolateyInstall.ps1", "chocolateyUninstall.ps1"].each do |win_file|
+          erb_file(File.join(VANAGON_ROOT, "templates/windows/#{win_file}.erb"), File.join(workdir, win_file), false, { :binding => binding })
+        end
+      end
+
+      # The specific bits used to generate a windows nuget package for a given project
+      # Nexus expects packages to be named #{name}-#{version}.nupkg. However, chocolatey
+      # will generate them to be #{name}.#{version}.nupkg. So, we have to rename the
+      # package after we build it. We also take advantage of the rename to sneak the
+      # architecture in the package name. Neither chocolatey nor nexus know how to deal
+      # with architecture, so we are just pretending it's part of the package name.
+      #
+      # @param project [Vanagon::Project] project to build a nuget package of
+      # @return [Array] list of commands required to build a nuget package for the given project from a tarball
+      def generate_nuget_package(project)
+        target_dir = project.repo ? output_dir(project.repo) : output_dir
+        ["mkdir -p output/#{target_dir}",
+        "mkdir -p $(tempdir)/#{project.name}/tools",
+        "cp #{project.name}.nuspec $(tempdir)/#{project.name}/",
+        "cp chocolateyInstall.ps1 chocolateyUninstall.ps1 $(tempdir)/#{project.name}/tools/",
+        "cp file-list $(tempdir)/#{project.name}/tools/file-list.txt",
+        "gunzip -c #{project.name}-#{project.version}.tar.gz | '#{@tar}' -C '$(tempdir)/#{project.name}/tools' --strip-components 1 -xf -",
+        "(cd $(tempdir)/#{project.name} ; #{self.drive_root}/ProgramData/chocolatey/bin/choco.exe pack #{project.name}.nuspec)",
+        "cp $(tempdir)/#{project.name}/#{project.name}-#{@architecture}.#{project.version}.#{project.release}.nupkg ./output/#{target_dir}/#{nuget_package_name(project)}"]
+      end
+
+      # Method to derive the package name for the project.
+      # Nexus expects packages to be named #{name}-#{version}.nupkg. However, chocolatey
+      # will generate them to be #{name}.#{version}.nupkg. So, we have to rename the
+      # package after we build it. We also take advantage of the rename to sneak the
+      # architecture in the package name. Neither chocolatey nor nexus know how to deal
+      # with architecture, so we are just pretending it's part of the package name.
+      #
+      # @param project [Vanagon::Project] project to name
+      # @return [String] name of the windows package for this project
+      def nuget_package_name(project)
+        "#{project.name}-#{@architecture}-#{project.version}.#{project.release}.nupkg"
       end
 
       # Add a repository (or install Chocolatey)
@@ -68,12 +122,32 @@ class Vanagon
         File.join("windows", target_repo, @architecture)
       end
 
+      # Return the drive root currently used on windows. At the moment, this is cygwin, but
+      # may at some unknown date change to bitvise.
+      #
+      # @return [String] the cygwin drive root
+      def drive_root
+        "/cygdrive/c"
+      end
+
+      # Get the windows path equivelant using cygpath
+      #
+      # @param path [String] the path to convert to windows style pathing
+      # @return [String] the windows style path for the given path
+      def convert_to_windows_path(path)
+        path.sub("/cygdrive/c", "C:")
+      end
+
       # Constructor. Sets up some defaults for the windows platform and calls the parent constructor
+      #
+      # Mingw varies on where it is installed based on architecure. We want to use which ever is on the system.
       #
       # @param name [String] name of the platform
       # @return [Vanagon::Platform::DEB] the win derived platform with the given name
       def initialize(name)
         @target_user = "Administrator"
+        @make = "(#{self.drive_root}/tools/mingw64/bin/mingw32-make || #{self.drive_root}/tools/mingw32/bin/mingw32-make)"
+        @tar = "/usr/bin/tar"
         @find = "/usr/bin/find"
         @sort = "/usr/bin/sort"
         super(name)
--- a/templates/windows/chocolateyInstall.ps1.erb
+++ b/templates/windows/chocolateyInstall.ps1.erb
@@ -1,0 +1,26 @@
+$toolsDir = "$(Split-Path -parent $MyInvocation.MyCommand.Definition)"
+
+<%- dirnames = get_directories.map {|d| d.path } -%>
+<%- dirnames.each do |dir| -%>
+  <%- windows_path = platform.convert_to_windows_path(dir) -%>
+if (Test-Path -path "<%= windows_path %>") {
+  Write-Debug "Warning: copying over '<%= windows_path %>'"
+} else {
+  New-Item -ItemType directory -Path "<%= windows_path %>"
+}
+$origin = Join-Path -path "$toolsDir" -childpath "<%= dir.gsub('/', '\\') %>"
+Copy-Item -path "$origin\*" -destination "<%= windows_path.gsub('/', '\\') %>\" -Recurse -Force
+<%- end -%>
+
+<%- filenames = get_files.map {|f| f.path} -%>
+<%- configfilenames = get_configfiles.map {|cf| cf.path} -%>
+<%- (filenames + configfilenames).each do |file| -%>
+  <%- windows_path = platform.convert_to_windows_path(file) -%>
+if (Test-Path -path "<%= windows_path %>") {
+  Write-Debug "Warning: copying over '<%= windows_path %>'"
+} elseif (!(Test-Path -path "<%= File.dirname windows_path %>")) {
+  New-Item -ItemType directory -Path "<%= File.dirname windows_path %>"
+}
+$origin = Join-Path -path "$toolsDir" -childpath "<%= file.gsub('/', '\\') %>"
+Copy-Item -path "$origin" -destination "<%= windows_path.gsub('/', '\\') %>" -Force
+<%- end -%>
--- a/templates/windows/chocolateyUninstall.ps1.erb
+++ b/templates/windows/chocolateyUninstall.ps1.erb
@@ -1,0 +1,52 @@
+$toolsDir = "$(Split-Path -parent $MyInvocation.MyCommand.Definition)"
+
+$contentFile=(join-path -path "$toolsDir" -childpath "file-list") + ".txt"
+if (Test-Path -path "$contentFile") {
+  Write-Debug "Removing files defiend in `'$contentFile`'"
+  $contents=get-content "$contentFile"
+  foreach ($fileInstalled in $contents) {
+    $fileInstalled=$fileInstalled.Replace("/cygdrive/c", "C:")
+    $fileInstalled=$fileInstalled.Replace("/", "\")
+    if (Test-Path -path "$fileInstalled" -PathType Container) {
+      Write-Debug "Not removing directory $fileInstalled"
+    } else {
+      remove-item -Path "$fileInstalled" -Force
+    }
+  }
+} else {
+  Write-Debug "'$contentFile' not found, not removing any files in non-standard chocolatey location"
+}
+
+<%- dirnames = get_directories.map {|d| d.path } -%>
+<%- dirnames.each do |dir| -%>
+  <%- windows_path = platform.convert_to_windows_path(dir) -%>
+if (Test-Path -path "<%= windows_path %>") {
+  # If this directory is empty, let's just delete it
+  if (!(Test-Path -Path "<%= windows_path %>\*")) {
+    Remove-Item -Path "<%= windows_path %>" -Force
+  } else {
+    # We need to try to remove all directories added in this package. We don't
+    # want to force remove a directory in case a different package has also
+    # installed files into that directory. We've already removed all the files
+    # for this package we are trying to remove. This half-way works, in that
+    # order matters, but is not well defined here.
+    Write-Debug "Removing any empty directories in '<%= windows_path %>'"
+    # Remove any empty subdirectories
+    $SearchRoot = "<%= windows_path %>"
+    Get-ChildItem -Path $SearchRoot -Recurse |
+      Where-Object {$_.PSIsContainer -eq $true -and (Get-ChildItem -Path $_.FullName) -eq $null} |
+        Remove-Item
+    # Remove the main directory if it's now empty after removing all empty subdirectories
+    Where-Object {(Get-ChildItem -Path "<%= windows_path %>" -Recurse) -eq $null} | Remove-Item
+  }
+}
+<%- end -%>
+
+<%- filenames = get_files.map {|f| f.path} -%>
+<%- configfilenames = get_configfiles.map {|cf| cf.path} -%>
+<%- (filenames + configfilenames).each do |file| -%>
+  <%- windows_path = platform.convert_to_windows_path(file) -%>
+if (Test-Path -path "<%= windows_path %>") {
+  Remove-Item -Path "<%= windows_path %>" -Force
+}
+<%- end -%>
--- a/templates/windows/project.nuspec.erb
+++ b/templates/windows/project.nuspec.erb
@@ -1,0 +1,31 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id><%= "#{@name}-#{@platform.architecture}" %></id>
+    <version><%= "#{@version}.#{@release}" %></version>
+    <title><%= @name %></title>
+    <authors><%= @vendor.match(/^(.*) <.*>$/)[1] %></authors>
+    <owners><%= @vendor.match(/^(.*) <.*>$/)[1] %></owners>
+    <projectUrl><%= @homepage %></projectUrl>
+    <packageSourceUrl><%= "https://github.com/puppetlabs/#{project}-vanagon" %></packageSourceUrl>
+    <projectSourceUrl><%= "https://github.com/puppetlabs/#{project}" %></projectSourceUrl>
+    <docsUrl><%= @homepage %></docsUrl>
+    <mailingListUrl><%= @homepage %></mailingListUrl>
+    <bugTrackerUrl><%= @homepage %></bugTrackerUrl>
+    <licenseUrl><%= @homepage %></licenseUrl>
+    <requireLicenseAcceptance>false</requireLicenseAcceptance>
+    <description>
+<%= @description %>
+    </description>
+    <summary><%= @description.lines.first.chomp %></summary>
+    <releaseNotes><%= @homepage %></releaseNotes>
+    <copyright><%= @vendor.match(/^(.*) <.*>$/)[1] %></copyright>
+    <tags>puppet configuration management infrastructure automation facter hiera mco mcollective marionette collective</tags>
+    <provides><%= get_replaces.map { |replace| "#{replace.replacement}" }.join(" ") %></provides>
+    <replaces><%= get_replaces.map { |replace| "#{replace.replacement}" }.join(" ") %></replaces>
+  </metadata>
+  <files>
+  <file src="tools\**" target="tools" />
+  </files>
+</package>
+<!-- character encoding: “UTF-8” -->
--- a/lib/vanagon/platform.rb
+++ b/lib/vanagon/platform.rb
@@ -6,7 +6,7 @@ class Vanagon
     attr_accessor :build_dependencies, :name, :vmpooler_template, :cflags, :ldflags, :settings
     attr_accessor :servicetype, :patch, :architecture, :codename, :os_name, :os_version
     attr_accessor :docker_image, :ssh_port, :rpmbuild, :install, :platform_triple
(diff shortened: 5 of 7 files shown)


templates/windows/project.nuspec.erb
```diff
@@ -0,0 +1,37 @@
...
+    <tags>puppet configuration management infrastructure automation facter hiera mco mcollective marionette collective</tags>
+    <provides><%= get_replaces.map { |replace| "#{replace.replacement}" }.join(" ") %></provides>
+    <replaces><%= get_replaces.map { |replace| "#{replace.replacement}" }.join(" ") %></replaces>
+  </metadata>
+  <files>
+<%- unless @bill_of_materials -%>
+    <file src="../bill-of-materials" target="usr/share/doc" />
+<%- end -%>
+<%- (get_files + get_configfiles + get_directories).each do |file| -%>
+    <file src="<%= file.path %>" target="<%= File.dirname file.path.sub(/^\/cygdrive\/c/,'') %>" />
```
melissa: This doesn't work yet. This depends on having the complete file list, but we're only passing in a subset of directories. More investigation required.

melissa:
Note: this is assuming we always want to build nuget archives on windows. Once we figure out the best way to define multiple package formats for a platform, we can restructure this work.

templates/windows/project.nuspec.erb
MikaelSmith: Let's not hard-code /cygdrive, Cygwin should recognize 'C:/'.

lib/vanagon/platform/windows.rb
```diff
@@ -8,7 +8,10 @@ class Windows < Vanagon::Platform
       def generate_package(project)
         target_dir = project.repo ? output_dir(project.repo) : output_dir
         ["mkdir -p output/#{target_dir}",
-        "touch output/#{target_dir}/#{project.name}-#{project.version}-#{project.release}-#{@architecture}.msi"]
+        "mkdir -p #{project.name}",
+        "cp #{project.name}.nuspec #{project.name}/",
+        "(cd #{project.name}; /cygdrive/c/ProgramData/chocolatey/choco pack)",
```
MikaelSmith: I'd rather use 'C:/' here as well.

joshcooper: :+1: to that.

If there are cases where we do need to use cygwin style paths, I recommend creating a `Shell` type object through which paths can be normalized. So today we're running with cygwin, and we can use `cygpath` tricks if necessary to map between windows (C:/) and cygwin bash (/cygdrive/c) and vice-versa. And in the future we can switch Shells (powershell, msys bash) and only have to change a few places.

melissa: hmm, @joshcooper can you elaborate on this? It sounds like a good idea, but I'm having trouble visualizing it

templates/windows/project.nuspec.erb
melissa: Yeah, I have to define the target directories with `/cygdrive/c` because something about how erb is populating the makefile is appending my working directory to the path when I use `C:/`, which is frustrating. Because of that, I have to get rid of the reference to `cygdrive` here :( More investigation on this is required

melissa:
Relates to https://github.com/puppetlabs/pl-build-tools-vanagon/pull/111

lib/vanagon/platform/windows.rb
MikaelSmith: I'm not sure if we need anything that complicated. We might get away with having a method that returns the drive root, in this case `/cygdrive/c`. In the future we could change that as needed.

MikaelSmith: To clarify @joshcooper's comment, Cygwin has a helper - `cygpath` that you can call with `cygpath C:/` and it will convert to `/cygdrive/c`. Josh was suggesting we have a method you pass all paths through (using `C:/` by default) that we could then have call `cygpath` to convert to a valid Cygwin path.

So something like
```
def normalize_path(p)
  `cygpath #{p}`
end

"(cd #{project.name}; #{platform.normalize_path('C:/ProgramData/chocolatey/choco`)} pack"
```

melissa: Ah, interesting. I've added in `@shell`, but it doesn't have the functionality yet that you're implying. I do think we should look into this, but I want to get things working, and then get that in there. I think we're just about there now, I just need to fix the `chocolateyInstall.ps1` and `chocolateyUninstall.ps1` scripts

templates/windows/chocolateyInstall.ps1.erb
```diff
@@ -0,0 +1,5 @@
+<%- dirnames = get_directories.map {|d| d.path } -%>
+
+<%- dirnames.each do |dir| -%>
+  Copy-Item <%= "#{dir.sub("/cygdrive/c", "C:/ProgramData/chocolatey/lib/#{project.name}")}/*" %> <%= dir.sub('/cygdirve/c', 'C:') %>
```
melissa: @ferventcoder this is where I am now. The files are getting installed into the chocolatey default directory, and I want to copy them from there to the specified location. I'm in the process of testing this at the moment, though, so I'll likely have more updates

lib/vanagon/platform/windows.rb
melissa:
```
$ cygpath /cygdrive/c/tools
/cygdrive/c/tools
```
```
$ cygpath C:/tools
/cygdrive/c/tools
```

It seems to not work the direction you're indicating here

melissa: nevermind, just saw the `--unix` flag

templates/windows/chocolateyInstall.ps1.erb
ferventcoder: Type at the end there.

ferventcoder: Why do we need to copy files around?

templates/windows/project.nuspec.erb
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id><%= @name %></id>
+    <version><%= "#{@version}.#{@release}" %></version>
```
ferventcoder: As long as version is `x.y.z` (or `x` or `x.y`) this will work. It only has `x.y.z.a` available and all need to be numerical.

templates/windows/project.nuspec.erb
```diff
@@ -0,0 +1,32 @@
...
+    <owners><%= @vendor.match(/^(.*) <.*>$/)[1] %></owners>
+    <projectUrl><%= @homepage %></projectUrl>
+    <packageSourceUrl><%= "https://github.com/puppetlabs/#{project}-vanagon" %></packageSourceUrl>
+    <projectSourceUrl><%= "https://github.com/puppetlabs/#{project}" %></projectSourceUrl>
+    <docsUrl><%= @homepage %></docsUrl>
+    <mailingListUrl><%= @homepage %></mailingListUrl>
+    <bugTrackerUrl><%= @homepage %></bugTrackerUrl>
+    <licenseUrl><%= @homepage %></licenseUrl>
+    <requireLicenseAcceptance>false</requireLicenseAcceptance>
+    <iconUrl>https://cdn.rawgit.com/ferventcoder/chocolatey-packages/02c21bebe5abb495a56747cbb9b4b5415c933fc0/icons/puppet.svg</iconUrl>
```
ferventcoder: We'll probably want to address this location to one more official?

ferventcoder: Unless only used internally.

templates/windows/chocolateyInstall.ps1.erb
melissa: What typo? I want to copy files around to get everything in the same directory. That way we can have all the libraries and binaries in predictable locations

templates/windows/chocolateyInstall.ps1.erb
```diff
@@ -0,0 +1,5 @@
+<%- dirnames = get_directories.map {|d| d.path } -%>
+<%- dirnames.each do |dir| -%>
+  <%- windows_path = platform.convert_to_windows_path(dir) -%>
+Copy-Item <%= "#{windows_path.sub("C:", "C:/ProgramData/chocolatey/lib/#{project.name}")}" %> <%= windows_path %> -recurse
```
melissa: HA! Something I'm doing here is definitely messing things up:
```
$ ls /cygdrive/c/ProgramData/chocolatey/lib/pl-boost/tools/cygdrive/c/tools/pl-build-tools/
bin  include  lib  share
```

melissa: Oooh, this is actually because of `"gunzip -c #{project.name}-#{project.version}.tar.gz | '#{@tar}' -C '#{project.name}/tools' --strip-components 1 -xf -"`. I haven't even parsed the chocolateyInstall scripts into the directory yet....

ferventcoder: Whatever it is we are trying to do here (I'm not sure I completely understand yet), is maybe similar to an already existing pattern:

~~~powershell
$toolsDir = "$(Split-Path -parent $MyInvocation.MyCommand.Definition)"
Install-ChocolateyZipPackage $packageName $url $toolsDir $url64
~~~

ferventcoder: `"C:/ProgramData/chocolatey/lib/#{project.name}"` is considered bad form as this may not be where a package is installed (if the chocolatey install directory is different).

Instead you should prefer the path to the folder where the chocolateyInstall.ps1 is - `$toolsDir = "$(Split-Path -parent $MyInvocation.MyCommand.Definition)"`

ferventcoder: If the components are zipped up, you can use `Get-ChocolateyUnzip` if it is local to the folder (embedded in the package).

~~~
$toolsDir = "$(Split-Path -parent $MyInvocation.MyCommand.Definition)"
$gzipFile = Join-Path $toolsDir 'somefile.tar.gz'
$destination = 'C:\tools\pl-build-tools'

# locally embedded
Get-ChocolateyUnzip $gzipFile $destination -packageName 'test'

# if it is remote, you can just use the following pattern
Install-ChocolateyZipPackage $packageName $url $toolsDir $url64
~~~

MikaelSmith: It'd be nice to avoid having them zipped, to avoid running unzip twice. Maybe that doesn't save us anything however, since we're already having to copy them.

lib/vanagon/platform/windows.rb
```diff
@@ -8,7 +8,12 @@ class Windows < Vanagon::Platform
       def generate_package(project)
         target_dir = project.repo ? output_dir(project.repo) : output_dir
         ["mkdir -p output/#{target_dir}",
-        "touch output/#{target_dir}/#{project.name}-#{project.version}-#{project.release}-#{@architecture}.msi"]
+        "mkdir -p #{project.name}/tools",
```
underscorgan: Should this be living under a tmpdir?

lib/vanagon/platform/windows.rb
```diff
@@ -44,12 +43,32 @@ def output_dir(target_repo = "")
...
       # Constructor. Sets up some defaults for the windows platform and calls the parent constructor
       #
       # @param name [String] name of the platform
       # @return [Vanagon::Platform::DEB] the win derived platform with the given name
       def initialize(name)
         @target_user = "Administrator"
+        @make = "#{self.drive_root}/tools/mingw64/bin/mingw32-make"
+        @tar = "/usr/bin/tar"
+        @find = "/usr/bin/find"
+        @sort = "/usr/bin/sort"
```
underscorgan: Are those paths right? They look very unix-y

lib/vanagon/platform/windows.rb
melissa: It already is! All of this gets run, by default, according to vanagon, under `/var/tmp/tmp.#{somerandomstring}`

lib/vanagon/platform/windows.rb
melissa: Yeah, it's cygwin, which is very deceptively unix-y :(

templates/windows/chocolateyInstall.ps1.erb
ferventcoder: We can also do that. What's the mimicking platforms use? Unzipped then? If so I can see having them already unzipped in the package. 

The first point still stands though - use `$toolsDir = "$(Split-Path -parent $MyInvocation.MyCommand.Definition)"` instead of `c:/programdata/chocolatey/lib/<pkgname>`

lib/vanagon/platform/windows.rb
ferventcoder: Likely they are used for cygwin, which allows them to look unix-y.

templates/windows/chocolateyInstall.ps1.erb
melissa: @MikaelSmith are you referring to not wanting to unzip multiple times during install? 

@ferventcoder what if we don't have a zip file? Should I stick with `Copy-Item` and iterate over our list of directories?

lib/vanagon/platform/windows.rb
```diff
@@ -8,7 +8,12 @@ class Windows < Vanagon::Platform
       def generate_package(project)
         target_dir = project.repo ? output_dir(project.repo) : output_dir
         ["mkdir -p output/#{target_dir}",
-        "touch output/#{target_dir}/#{project.name}-#{project.version}-#{project.release}-#{@architecture}.msi"]
+        "mkdir -p #{project.name}/tools",
+        "cp #{project.name}.nuspec #{project.name}/",
+        "cp chocolateyInstall.ps1 chocolateyUninstall.ps1 #{project.name}/tools",
+        "gunzip -c #{project.name}-#{project.version}.tar.gz | '#{@tar}' -C '#{project.name}/tools' -xf -",
+        "(cd #{project.name}; #{self.drive_root}/ProgramData/chocolatey/choco pack)",
```
ferventcoder: `#{self.drive_root}/ProgramData/chocolatey/choco` should use `#{self.drive_root}/ProgramData/chocolatey/bin/choco.exe` instead.

templates/windows/chocolateyInstall.ps1.erb
melissa: @ferventcoder what do you mean by "the mimicking platforms use"? This can be what makes the most sense for building nuget packages, as long as it's not too difficult to do programmatically :)

MikaelSmith: Yes, not wanting to unzip multiple times during install.

ferventcoder: Copy-Item is fine. You don't need to go further than one call though (but seeing an example of this in action would be helpful) because of -recurse.

Note that you will also need -force to overwrite.

~~~powershell
$toolsDir = "$(Split-Path -parent $MyInvocation.MyCommand.Definition)"
$sourceDir = Join-Path $toolsDir 'pl-build-tools'
$destination = 'C:\tools\pl-build-tools'
# the \* is very important
Copy-Item $sourceDir\* $destination -Recurse -Force
~~~

ferventcoder: > the mimicking platforms use

What I mean is the contents of the nuget packages should match somewhat closely with other platform packages to keep things simple.

ferventcoder: Are you also planning to put things on the path?

melissa: No, I will not put these on the path at install time

templates/windows/chocolateyInstall.ps1.erb
```diff
@@ -0,0 +1,8 @@
+$toolsDir = "$(Split-Path -parent $MyInvocation.MyCommand.Definition)"
+
+<%- dirnames = get_directories.map {|d| d.path } -%>
+<%- filenames = get_files.map {|f| f.path} -%>
+<%- configfilenames = get_configfiles.map {|cf| cf.path} -%>
+<%- (dirnames + filenames + configfilenames).each do |name| -%>
+Copy-Item (Join-Path $toolsDir '<%= name %>') <%= platform.convert_to_windows_path(name) %> -Force
```
melissa: @ferventcoder rather than copy over the directories recursively, I decided it would be a better approach to copy over all the files explicitly, since we have that information. This *should* include empty directories.

melissa: Shoot, I keep forgetting we don't actually have this information yet...

melissa: So, we have a list of files when we're building the package, stored in a text file called `file-list`. Would it be worthwhile to parse these files in the chocolateyInstall and chocolateyUninstall scripts when we copy them into the tools directory to we know exactly what is getting installed and removed? I'm more concerned about this for the uninstall script, so that we only uninstall boost files when boost is uninstalled, and not yaml-cpp files, even though they're both be installed under `C:\tools\pl-build\tools`

lib/vanagon/platform/windows.rb
```diff
@@ -8,7 +8,12 @@ class Windows < Vanagon::Platform
       def generate_package(project)
         target_dir = project.repo ? output_dir(project.repo) : output_dir
         ["mkdir -p output/#{target_dir}",
-        "touch output/#{target_dir}/#{project.name}-#{project.version}-#{project.release}-#{@architecture}.msi"]
+        "mkdir -p #{project.name}/tools",
+        "cp #{project.name}.nuspec #{project.name}/",
+        "cp chocolateyInstall.ps1 chocolateyUninstall.ps1 #{project.name}/tools",
+        "gunzip -c #{project.name}-#{project.version}.tar.gz | '#{@tar}' -C '#{project.name}/tools' -xf -",
+        "(cd #{project.name}; #{self.drive_root}/ProgramData/chocolatey/bin/choco.exe pack)",
```
ferventcoder: you don't have to drop into the folder to do this.

ferventcoder: you can just put the path to the nuspec file as part of the args for pack.

melissa: Will the nupkg file be dropped into the directory where pack is being run? Or in the directory where the nuspec lives?

ferventcoder: I believe and want to say where the pack is run from.

templates/windows/chocolateyUninstall.ps1.erb
```diff
@@ -0,0 +1,20 @@
...
+  Write-Debug "Removing files defiend in `'$contentFile`'"
+  $contentFile
+  $contents=get-content $contentFile
+  foreach ($fileInstalled in $contents) {
+    $fileInstalled.
+    if ($fileInstalled.PSIsContainer) {
+      Write-Debug "Not removing directory $fileInstalled"
+    } else {
+      try {
+        remove-item -Path "$fileInstalled" -ErrorAction SilentlyContinue -Recurse -Force
```
ferventcoder: The `-ErrorAction SilentlyContinue` will prevent the catch from happening.

ferventcoder: You'll want to remove -recurse most likely since it is a single file.

templates/windows/chocolateyUninstall.ps1.erb
```diff
@@ -0,0 +1,20 @@
+$toolsDir = "$(Split-Path -parent $MyInvocation.MyCommand.Definition)"
+
+$contentFile=(join-path $toolsDir 'file-list') + ".txt"
+if ((Test-Path -path $contentFile)) {
+  Write-Debug "Removing files defiend in `'$contentFile`'"
+  $contentFile
+  $contents=get-content $contentFile
+  foreach ($fileInstalled in $contents) {
+    $fileInstalled.
```
ferventcoder: What does this line do?

melissa: no idea, copied it from https://github.com/chocolatey/choco/blob/master/src/chocolatey.resources/helpers/functions/UnInstall-ChocolateyZipPackage.ps1#L50 though the `.` is a typo

lib/vanagon/platform/windows.rb
stahnma: Are those defaults correct for AIX?

shrug: Have we vetted those on aix or solaris yet?

haus: This is the windows platform config, so it won't affect aix or solaris, which will get the default of just "sort" and "find".

stahnma: Ah, my bad.

stahnma:
I'm :+1: . Not sure if there are outstanding items.

lib/vanagon/platform/windows.rb
shrug: ah, oops, reading filenames is hard

melissa: Thanks, @haus :) Yeah, these, for all other platforms, will default to `find` and `sort`, which has been working on AIX and Solaris so far...

lib/vanagon/platform/windows.rb
melissa: Actually, you're right, it should be in a different tempdir

lib/vanagon/platform/windows.rb
melissa: Would it be bad practice to just leave this as is?

jcoconnor:
:+1: Looks good

templates/windows/chocolateyInstall.ps1.erb
```diff
@@ -0,0 +1,22 @@
...
+<%- filenames = get_files.map {|f| f.path} -%>
+<%- configfilenames = get_configfiles.map {|cf| cf.path} -%>
+<%- (filenames + configfilenames).each do |file| -%>
+  <%- windows_path = platform.convert_to_windows_path(file) -%>
+if (Test-Path -path "<%= windows_path %>") {
+  Write-Debug "Warning: copying over '<%= windows_path %>'"
+}
+$origin = Join-Path -path "$toolsDir" -childpath "<%= file.gsub('/', '\\') %>"
+Copy-Item -path "$origin" -destination "<%= windows_path.gsub('/', '\\') %>" -Force
+<%- end -%>
```
melissa: shoot, I realized this isn't doing what we want....... I haven't figured out why yet.

it should be copying 
```
$ ls /cygdrive/c/ProgramData/chocolatey/lib/pl-boost/tools/cygdrive/c/tools/pl-build-tools/
bin  include  lib  share
```

out to 
```
$ ls /cygdrive/c/tools/pl-build-tools/
$Recycle.Bin  b2.exe  bin  bjam.exe  crash_dumps  cygwin64  include  lib  share
```
but, as you can see, we're getting extra things I what shouldn't be there, namely `cygwin64`

templates/windows/chocolateyUninstall.ps1.erb
```diff
@@ -0,0 +1,20 @@
...
+    $fileInstalled
+    if ($fileInstalled.PSIsContainer) {
+      Write-Debug "Not removing directory $fileInstalled"
+    } else {
+      remove-item -Path "$fileInstalled" -ErrorAction SilentlyContinue -Force
+    }
+  }
+} else {
+  Write-Debug "'$contentFile' not found, not removing any files in non-standard chocolatey location"
+}
```
melissa: uninstall also unsuccessful

```
 C:\ProgramData\chocolatey\lib\pl-boost\tools\file-list.txt
 C:\tools\pl-build-tools
pl-boost uninstall not successful.
Error while running 'C:\ProgramData\chocolatey\lib\pl-boost\tools\chocolateyUninstall.ps1'.
 See log for details.
pl-boost not uninstalled. An error occurred during uninstall:
 pl-boost uninstall not successful.

Chocolatey uninstalled 0/1 packages. 1 packages failed.
 See the log for details (C:\ProgramData\chocolatey\logs\chocolatey.log).
Failures
 - pl-boost
```

melissa: There isn't anything useful in the error longs about what exactly failed in the powershell script :(

lib/vanagon/platform/windows.rb
```diff
@@ -8,7 +8,13 @@ class Windows < Vanagon::Platform
       def generate_package(project)
         target_dir = project.repo ? output_dir(project.repo) : output_dir
         ["mkdir -p output/#{target_dir}",
-        "touch output/#{target_dir}/#{project.name}-#{project.version}-#{project.release}-#{@architecture}.msi"]
+        "mkdir -p $(tempdir)/#{project.name}/tools",
+        "cp #{project.name}.nuspec $(tempdir)/#{project.name}/",
+        "cp chocolateyInstall.ps1 chocolateyUninstall.ps1 $(tempdir)/#{project.name}/tools/",
+        "cp file-list $(tempdir)/#{project.name}/tools/file-list.txt",
+        "gunzip -c #{project.name}-#{project.version}.tar.gz | '#{@tar}' -C '$(tempdir)/#{project.name}/tools' --strip-components 1 -xf -",
```
ferventcoder: Perhaps cygwin64 is in this tar?

ferventcoder: I'm guessing you may need to do some additional work to clean up after you've unpacked.

lib/vanagon/platform/windows.rb
```diff
@@ -8,7 +8,13 @@ class Windows < Vanagon::Platform
...
+        "(cd $(tempdir)/#{project.name} ; #{self.drive_root}/ProgramData/chocolatey/bin/choco.exe pack #{project.name}.nuspec)",
+        "cp $(tempdir)/#{project.name}/*.nupkg ./output/#{target_dir}"]
```
ferventcoder: As mentioned earlier, you could eliminate these two steps with a well formed pack command that puts the package in target directory on build.

templates/windows/chocolateyInstall.ps1.erb
ferventcoder: I commented above. I think you may have extra stuff coming in with the gunzip of the tar that need to be cleaned up.

What's in the package?

templates/windows/chocolateyUninstall.ps1.erb
```diff
@@ -0,0 +1,20 @@
+$toolsDir = "$(Split-Path -parent $MyInvocation.MyCommand.Definition)"
+
+$contentFile=(join-path -path "$toolsDir" -childpath "file-list") + ".txt"
+if (Test-Path -path "$contentFile") {
+  Write-Debug "Removing files defiend in `'$contentFile`'"
+  $contentFile
```
ferventcoder: What is this? I know you copied it from somewhere, I just don't understand that is does.

templates/windows/chocolateyUninstall.ps1.erb
```diff
@@ -0,0 +1,20 @@
...
+  $contents=get-content "$contentFile"
```
ferventcoder: What is the encoding of `file-list.txt`? You may need to tell PowerShell what the encoding is if it is not UTF-8 or ASCII.

templates/windows/chocolateyUninstall.ps1.erb
```diff
@@ -0,0 +1,20 @@
...
+  foreach ($fileInstalled in $contents) {
+    $fileInstalled=$fileInstalled.Replace("/cygdrive/c", "C:")
+    $fileInstalled=$fileInstalled.Replace("/", "\")
+    $fileInstalled
```
ferventcoder: Either do Write-Host "$fileInstalled" or remove this line.

templates/windows/chocolateyUninstall.ps1.erb
```diff
@@ -0,0 +1,20 @@
...
+    if ($fileInstalled.PSIsContainer) {
+      Write-Debug "Not removing directory $fileInstalled"
+    } else {
+      remove-item -Path "$fileInstalled" -ErrorAction SilentlyContinue -Force
```
ferventcoder: Take away the silentlycontinue error action for now to debug.

templates/windows/chocolateyUninstall.ps1.erb
ferventcoder: What's in file-list.txt?

templates/windows/chocolateyInstall.ps1.erb
melissa: The fact that `/cygdrive/c/ProgramData/chocolatey/lib/pl-boost/tools/cygdrive/c/tools/pl-build-tools/` makes me think that line of thought is a red herring. We should be copying files directly from the above path to the specified one

templates/windows/chocolateyUninstall.ps1.erb
melissa: file-list.txt contains a list of files that *should* be contained in the package. That's the list we use to populate the tarball.

lib/vanagon/platform/windows.rb
melissa: Unpacked the tarball? This should be a clean list. I would be very surprised if it contains mingw64. I'll definitely dig into that line of questioning though

lib/vanagon/platform/windows.rb
melissa: I know... I'm inclined to leave it like this though. It isolates the package build process from everything, and makes things a little easier to maintain (in my opinion). Would it be bad practice to leave it like this? I prefer explicit rather then clean, though clean does have its merits.

templates/windows/chocolateyUninstall.ps1.erb
melissa: Nor do I... I'll take it out if you don't think it's necessary

templates/windows/chocolateyUninstall.ps1.erb
melissa: How do I do that? I have no idea what the encoding of the file is...

lib/vanagon/platform/windows.rb
ferventcoder: I think it's fine to leave it like this.

templates/windows/chocolateyUninstall.ps1.erb
ferventcoder: might be good to know. But it does look like it at least pulls the first item out of there.

templates/windows/chocolateyUninstall.ps1.erb
ferventcoder: I meant I'd like to see the list from a known one that is failing. We can't really debug without a complete picture.

templates/windows/chocolateyInstall.ps1.erb
ferventcoder: Ah true. Could just be other files that were already there.

lib/vanagon/platform/windows.rb
melissa: Found the error, it was in the chocolateyInstall.ps1 script. Fixed now.

templates/windows/chocolateyUninstall.ps1.erb
melissa: Fixed now :) Uninstall succeeds, but leaves a lot of the directory structure on the file system. I did my best to thoroughly clean it up, but wasn't entirely successful

melissa:
All that's left is to fix up the commits, I think

melissa:
Installs and uninstalls without failure, installs into the expected location, without strange extra files!!
http://builds.puppetlabs.lan/pl-build-tools/848f0fae666f231ba65f45ddd9420dbbd02331b4/artifacts/windows/x64/ is the nupkg file I built if you want to check it out!

MikaelSmith:
All the `bin` and `share` for Boost.Build aren't needed, but I suspect that's not new to the build process. Any reason you install to the long location path?

Neither of those are reasons not to merge it, just observations.

lib/vanagon/platform/windows.rb
```diff
@@ -17,21 +45,42 @@ def generate_package(project)
...
+      end
+
       # Method to derive the package name for the project
       #
       # @param project [Vanagon::Project] project to name
       # @return [String] name of the windows package for this project
-      def package_name(project)
-        "#{project.name}-#{project.version}-#{project.release}-#{@architecture}.msi"
+      def nuget_package_name(project)
+        "#{project.name}-#{@architecture}-#{project.version}.#{project.release}.nupkg"
```
ferventcoder: You might add some more comments that nexus requires the - between name and version, so there is no confusion on why it gets renamed as so.

ferventcoder:
Besides git commit cleanup  and the comments I asked for above, I think this is :+1: .

lib/vanagon/platform/windows.rb
melissa: Check out the comment above `generate_nuget_package`. That has the explanation. I'll add it here too though, just to be thorough

melissa:
@MikaelSmith your boost question doesn't really have anything to do with this pr. In the PR where we define how boost it built (https://github.com/puppetlabs/pl-build-tools-vanagon/pull/111), those directories are included because that's what gets installed. It's how we're doing things on *nix too

jcoconnor:
:+1:

PR #225, opened 2015-11-16
(RE-5976) - Install Chocolatey as part of Windows/Vanagon bootstrap

jcoconnor:
Ticket/vanagon for the win/re 5976 install chocolatey
Initial ticket for discussion.

### DIFF
--- a/lib/vanagon/platform/windows.rb
+++ b/lib/vanagon/platform/windows.rb
@@ -34,6 +34,30 @@ class Vanagon
         "#{project.name}-#{project.version}-#{project.release}-#{@architecture}.msi"
       end
 
+      # Add a repository (or install Chocolatey)
+      # Note - this only prepares the list of commands to be executed once the Platform
+      # has been setup
+      #
+      # @param definition [String] Definition for adding repo, can be a Repo URL (including file:)
+      #    If file suffix is 'ps1' it is downloaded and executed to install chocolatey
+      # @return [Array]  Commands to executed after platform startup
+      def add_repository(definition)
+        definition = URI.parse(definition)
+        commands = []
+
+        if definition.scheme =~ /^(http|ftp|file)/
+          if File.extname(definition.path) == '.ps1'
+            commands << %(powershell.exe -NoProfile -ExecutionPolicy Bypass -Command 'iex ((new-object net.webclient).DownloadString(\"#{definition}\"))')
+          else
+            commands << %(C:/ProgramData/chocolatey/bin/choco.exe source add -n #{definition.host}-#{definition.path.gsub('/', '-')} -s "#{definition}" --debug || echo "Oops, it seems that you don't have chocolatey installed on this system. Please ensure it's there by adding something like 'plat.add_repository 'https://chocolatey.org/install.ps1'' to your platform definition.")
+          end
+        else
+          fail "Invalid repo specification #{definition}"
+        end
+
+        commands
+      end
+
       # Get the expected output dir for the windows packages. This allows us to
       # use some standard tools to ship internally.
       #


lib/vanagon/platform/windows.rb
```diff
@@ -34,6 +34,143 @@ def package_name(project)
...
+  $nupkg = Join-Path $chocoPkgDir 'chocolatey.nupkg'
+  if (![System.IO.Directory]::Exists($chocoPkgDir)) { [System.IO.Directory]::CreateDirectory($chocoPkgDir); }
+  Copy-Item "$file" "$nupkg" -Force -ErrorAction SilentlyContinue
+}
+catch
+{
+  Write-Host "$($_.Exception.Message)"
+  exit 1
+}
+        BATCH
```
ferventcoder: Doesn't this need to be without spaces to finish the HEREDOC?

lib/vanagon/platform/windows.rb
```diff
@@ -34,6 +34,143 @@ def package_name(project)
...
+      end
+      # Enclose the supplied command with the Powershell wrapper.
+      #
+      # @param command [String] optional repo target for built packages defined
+      #   at the project level
+      # @return [String] Wrapped Powershell Command
+      def powershell_command(command)
+        result = "powershell.exe -NoProfile -ExecutionPolicy Unrestricted -InputFormat None -Command #{command}"
```
ferventcoder: You may want to do something similar to what we do when we look for the powershell.exe in known locations prior to settling on one, see https://github.com/puppetlabs/puppetlabs-powershell/blob/e896f8b4546b4c34d1144fb6fa31da2df78274c7/lib/puppet/provider/exec/powershell.rb#L6-L13

lib/vanagon/platform/windows/install-chocolatey.ps1
```diff
@@ -0,0 +1,65 @@
...
+
+  Write-Output 'Ensuring chocolatey.nupkg is in the lib folder'
+  $chocoPkgDir = Join-Path $chocoPath 'lib\chocolatey'
+  $nupkg = Join-Path $chocoPkgDir 'chocolatey.nupkg'
+  if (![System.IO.Directory]::Exists($chocoPkgDir)) { [System.IO.Directory]::CreateDirectory($chocoPkgDir); }
+  Copy-Item "$file" "$nupkg" -Force -ErrorAction SilentlyContinue
+}
+catch
+{
+  Write-Host "$($_.Exception.Message)"
```
ferventcoder: Write-Error

lib/vanagon/platform/windows.rb
mckern: No, that's why there's a dash (`<<-`) at the opening of the HEREDOC.

ferventcoder: Thanks. :)

lib/vanagon/platform/windows.rb
```diff
@@ -34,6 +34,143 @@ def package_name(project)
...
+        commands
+      end
+
+      # Batch up powershell commands to extract out and install chocolatey
+      #
+      # @param command [String] optional repo target for built packages defined
+      #   at the project level
+      # @return [String] Wrapped Powershell Command
+      def install_chocolately_ps1(url)
+        batch_contents = <<-BATCH
```
mckern: Vanagon has the `[String#undent](https://github.com/puppetlabs/vanagon/blob/master/lib/vanagon/extensions/string.rb)` extension method, so this HEREDOC doesn't need to be strictly aligned to column 1. Usage would look like this:

```ruby
  batch_contents = <<-BATCH.undent
    some big long string, which will have all leading whitespace to the length of the first line stripped out
  BATCH
```

lib/vanagon/platform/windows.rb
```diff
@@ -34,6 +34,143 @@ def package_name(project)
...
+        batch_path = "./#{batch_name}"
+
+        log_path = "\\install-puppet-#{timestamp}.log"
+
+        Tempfile.open(batch_name) do |tmp_file|
+          batch_contents = install_chocolately_ps1(url)
+
+          File.open(tmp_file.path, 'w') { |file| file.puts(batch_contents) }
+
+          require 'pry'
```
mckern: I don't think we want to leave `pry` bindings in?

lib/vanagon/platform/windows.rb
```diff
@@ -34,6 +34,143 @@ def package_name(project)
         "#{project.name}-#{project.version}-#{project.release}-#{@architecture}.msi"
       end
 
+      def add_repository(definition)
+        definition = URI.parse(definition)
+
+        ps1_command = install_chocolately_ps1 (definition)
+
+
```
mckern: Maybe too much whitespace?
</nitpick>

lib/vanagon/platform/windows.rb
```diff
@@ -34,6 +34,143 @@ def package_name(project)
         "#{project.name}-#{project.version}-#{project.release}-#{@architecture}.msi"
       end
 
+      def add_repository(definition)
+        definition = URI.parse(definition)
+
+        ps1_command = install_chocolately_ps1 (definition)
```
mckern: space between function and argument

lib/vanagon/platform/windows.rb
```diff
@@ -34,6 +34,143 @@ def package_name(project)
         "#{project.name}-#{project.version}-#{project.release}-#{@architecture}.msi"
       end
 
+      def add_repository(definition)
+        definition = URI.parse(definition)
```
mckern: If the repository to be added will always be a URL, "definition" is an undescriptive variable name. Consider something more descriptive like "url", "repo_url", "a_url_to_be_named_later", "there_are_many_urls_like_this_but_this_url_is_mine", etc.

mckern: (And if it's not always going to be a URL, just passing it to `URI` might explode hilariously)

lib/vanagon/platform/windows.rb
```diff
@@ -34,6 +34,143 @@ def package_name(project)
...
+
+        ps1_command = install_chocolately_ps1 (definition)
+
+
+        commands = ["cat <<EOF; #{ps1_command}; EOF > install-chocolatey.ps1"]
+
+        #install_chocolately_on_host(definition)
+        puts "Now doing Add_Repo #{definition}"
```
mckern: Is this boilerplate notification or debugging text?

lib/vanagon/platform/windows.rb
```diff
@@ -34,6 +34,143 @@ def package_name(project)
...
+
+        if definition.scheme =~ /^(http|ftp)/
+          # use choco to install the package.
+          # choco source add -n=bob -s "https://somewhere/out/there/api/v2/"
+          puts "Adding Repo #{definition}"
+          commands << "echo choco add -s #{definition}"
```
mckern: The command being appended doesn't quite line up with the comment above it; is that intentional?

lib/vanagon/platform/windows.rb
```diff
@@ -34,6 +34,143 @@ def package_name(project)
...
+
+          require 'pry'
+          binding.pry
+
+          #Vanagon::Utilities.rsync_to("#{batch_path}", "#{@target_user}@#{@target}", "xxxxx", @ssh_port)
+          #Vanagon::Engine::Base.ship_workdir("./#{batch_path}")
+        end
+
+
+        return batch_path, log_path
```
mckern: `install_chocolately_on_host` returns an array of two paths but it has a name that implies it's more of an imperative than a builder/constructor method. It also implicitly returns an array, but if you're explicitly calling `return` you should make sure that the object you return is also explicit:

```ruby
>> def some_thing(x, y)
|    return x, y
|  end  
=> :some_thing
>> some_thing true, false
=> [
    [0] true,
    [1] false
]
```

The name should probably better reflect what it's doing, and the return value should probably be explicit.

lib/vanagon/platform/windows.rb
```diff
@@ -34,6 +34,143 @@ def package_name(project)
...
+
+      # Given a host, path to MSI and MSI options, will create a batch file
+      #   on the host, returning the path to the randomized batch file and
+      #   the randomized log file
+      # @param [Host] host An object implementing {Beaker::Hosts}'s interface.
+      # @param [String] msi_path The path of the MSI - can be a local Windows style file path like
+      #                   c:\temp\puppet.msi OR a url like https://download.com/puppet.msi or file://c:\temp\puppet.msi
+      # @param  [Hash{String=>String}] msi_opts MSI installer options
+      #                   See https://docs.puppetlabs.com/guides/install_puppet/install_windows.html#msi-properties
+      # @api private
```
mckern: Is all of this comment/yardoc copypasta?

lib/vanagon/platform/windows.rb
ferventcoder: This was mentioned as very much a WIP.

mckern: Still don't think we want to leave `pry` bindings in?

lib/vanagon/platform/windows.rb
jcoconnor: It's Copypasta of the most excellent @haus kind.

WIP = WIP

jcoconnor:
@mckern @ferventcoder @Iristyle @melissa 

Trying an alternative approach here of simply curling down install.ps1 from chocolaty.org to install it.
Still think we need to generalise it a bit more, but am tempted to ticket that separately as this will move the project on.

Still need to update pathing etc.,  so this is still very much WIP-DO-NOT-MERGE

lib/vanagon/platform/windows.rb
haus: are definitions expected to be unique for windows? if so why make a random string for the name instead of just using part (ideally the unique part) of the definition?

lib/vanagon/platform/windows.rb
```diff
@@ -34,6 +37,17 @@ def package_name(project)
         "#{project.name}-#{project.version}-#{project.release}-#{@architecture}.msi"
       end
 
+      def add_repository(definition)
```
haus: Having some yard doc here would help to know what definition is expected to be (always a url, or something else).

melissa:
what about this:
`plat.add_repository 'https://chocolatey.org/install.ps1'` or  `plat.add_repository 'http://buildsources.delivery.puppetlabs.net/windows/install-chocolatey.ps1'` or `plat.add_repository 'https://www.myget.org/F/puppetlabs'`
along with the following in windows.rb
```
def add_repository(definition)
  commands = []
  if definition.scheme =~ /^(http|ftp)/
    if File.extname(definition.path) == '.ps1'
      commands << "powershell.exe -NoProfile -ExecutionPolicy Bypass -Command 'iex ((new-object net.webclient).DownloadString(\'#{definition}\'))'"
    else
      commands << %Q{C:/ProgramData/chocolatey/bin/choco.exe source add -n #{definition.host}-#{definition.path.gsub('/', '-')} -s "#{definition}" --debug || echo "Oops, it seems that you don't have chocolatey installed on this system. Please ensure it's there by adding something like 'plat.add_repository 'https://chocolatey.org/install.ps1''"}
    end
  end
end
```

lib/vanagon/platform/windows.rb
jcoconnor: Good suggestion.  The random string was a quick thought to get it to work for this stage.

stahnma:
I like the last comment from @melissa. This would make chocolatey installation a platform setting per project and map fairly cleanly to the existing abstractions.

lib/vanagon/platform/windows.rb
```diff
@@ -1,3 +1,6 @@
+
+require 'securerandom'
```
melissa: Are we still using this?

lib/vanagon/platform/windows.rb
```diff
@@ -34,6 +37,30 @@ def package_name(project)
...
+      # @return [Array]  Commands to executed after platform startup
+      def add_repository(definition)
+        definition = URI.parse(definition)
+        commands = []
+
+        if definition.scheme =~ /^(http|ftp|file)/
+          if File.extname(definition.path) == '.ps1'
+            commands << %Q{powershell.exe -NoProfile -ExecutionPolicy Bypass -Command 'iex ((new-object net.webclient).DownloadString(\"#{definition}\"))'}
+          else
+            commands << %Q{C:/ProgramData/chocolatey/bin/choco.exe source add -n #{definition.host}-#{definition.path.gsub('/', '-')} -s "#{definition}" --debug || echo "Oops, it seems that you don't have chocolatey installed on this system. Please ensure it's there by adding something like 'plat.add_repository 'https://chocolatey.org/install.ps1'' to your platform definition."}
```
melissa: I'm not sure how useful `-n #{definition.host}-#{definition.path.gsub('/', '-')}` is going to be.... do you think it's a reasonable assumption that will be relatively unique for every new repo a user adds?

melissa: Also, is there any way to verify the url passed in is actually a valid nuget repo?

jcoconnor: @melissa 
1. The gsub format above gives an untidy name, but it does work out as unique as each URL itself *should* be unique.
2. Its difficult to very the URL is an actual repo until we do the choco add source - but since we have error detection in the actual add source command, that will fail if the repo is invalid, so we kind of do have verification.

jcoconnor:
The travis failure above is debatable - 3 "offences" were picked up as follows:
Offenses:

    lib/vanagon/platform/windows.rb:51:25: C: Use % instead of %Q.
                commands << %Q{powershell.exe -NoProfile -ExecutionPolicy Bypass -Command 'iex ((new-object net.webclient).DownloadString(\"#{definition}\"))'}
                        ^^^
    lib/vanagon/platform/windows.rb:53:25: C: Use % instead of %Q.
            commands << %Q{C:/ProgramData/chocolatey/bin/choco.exe source add -n #{definition.host}-#{definition.path.gsub('/', '-')} -s "#{definition}" --debug || echo "Oops, it seems that you don't have chocolatey installed on this system. Please ensure it's there by adding something like 'plat.add_repository 'https://chocolatey.org/install.ps1'' to your platform definition."}
                        ^^^
    lib/vanagon/platform/windows.rb:53:25: C: %Q-literals should be delimited by ( and )
            commands << %Q{C:/ProgramData/chocolatey/bin/choco.exe source add -n #{definition.host}-#{definition.path.gsub('/', '-')} -s "#{definition}" --debug || echo "Oops, it seems that you don't have chocolatey installed on this system. Please ensure it's there by adding something like 'plat.add_repository 'https://chocolatey.org/install.ps1'' to your platform definition."}
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

    36 files inspected, 3 offenses detected

mckern:
@jcoconnor they're debateable but we're trying to broadly follow the rubocop conventions as our "house style" for Ruby. Changing from `%Q{` to `%Q(` is pretty straightforward and easy enough to amend to your commit.

lib/vanagon/platform/windows.rb
```diff
@@ -34,6 +34,30 @@ def package_name(project)
...
       end
 
+      # Add a repository (or install Chocolatey)
+      # Note - this only prepares the list of commands to be executed once the Platform
+      # has been setup
+      #
+      # @param definition [String] Definition for adding repo, can be a Repo URL (including file:)
+      #    If file suffix is 'ps1' it is downloaded and executed to install chocolatey
+      # @return [Array]  Commands to executed after platform startup
+      def add_repository(definition)
```
mckern: I know it's just a variable name, so this is a nitpick, but `definition` is still sort of nebulous. I won't block over it, but wanted to mention it.

melissa: `definition` is what all other `add_repository` methods use. It made sense to follow that established pattern.

mckern: Noted.

jcoconnor:
@mckern %Q offences corrected.

mckern:
:+1: pending Travis

ferventcoder:
LGTM.

ferventcoder:
Minor nit* - the git commit message itself could use some love and typo fixups.

jcoconnor:
@ferventcoder That better?

melissa:
:+1: 

tested with
```
$ cat configs/platforms/win-2012-x64.rb
platform "win-2012-x64" do |plat|
  plat.vmpooler_template "win-2012r2-x86_64"

  # Temporary while `provision_with` and `add_repository` are still in progress
  plat.add_build_repository "https://chocolatey.org/install.ps1"
  plat.add_build_repository "https://www.myget.org/F/puppetlabs"

  plat.install_build_dependencies_with "C:/ProgramData/chocolatey/bin/choco.exe install -y"
end
```

ferventcoder:
Close enough. Spelled "Chocolatey"

jcoconnor:
:scream:

PR #226, opened 2015-11-19
Add Hardware engine to vanagon

stahnma:
Read the commit messages now. They are very detailed and kept updated as this is reviewed. But, here's a picture.

### DIFF
--- a/lib/vanagon/driver.rb
+++ b/lib/vanagon/driver.rb
@@ -23,16 +23,18 @@ class Vanagon
 
       @platform = Vanagon::Platform.load_platform(platform, File.join(@@configdir, "platforms"))
       @project = Vanagon::Project.load_project(project, File.join(@@configdir, "projects"), @platform, components)
+      @@logger = Logger.new('vanagon_hosts.log')
+      @@logger.progname = 'vanagon'
 
       # If a target has been given, we don't want to make any assumptions about how to tear it down.
       engine = 'base' if target
+      # Hardware has explicit teardown to unlock the node
+      engine = 'hardware' if @platform.build_host
       require "vanagon/engine/#{engine}"
       @engine = Object.const_get("Vanagon::Engine::#{engine.capitalize}").new(@platform, target)
 
-      @@logger = Logger.new('vanagon_hosts.log')
-      @@logger.progname = 'vanagon'
     rescue LoadError => e
-      raise Vanagon::Error.wrap(e, "Could not load the desired engine '#{@engine_name}'.")
+      raise Vanagon::Error.wrap(e, "Could not load the desired engine '#{engine}'.\n#{e}")
     end
 
     def cleanup_workdir
@@ -88,7 +90,11 @@ class Vanagon
       puts e
       puts e.backtrace.join("\n")
       raise e
-    end
+    ensure
+      if @engine == "hardware"
+        @engine.teardown
+      end
+  end
 
     def prepare(workdir = nil)
       @workdir = workdir ? FileUtils.mkdir_p(workdir).first : Dir.mktmpdir
--- a/lib/vanagon/driver.rb
+++ b/lib/vanagon/driver.rb
@@ -29,12 +29,12 @@ class Vanagon
       # If a target has been given, we don't want to make any assumptions about how to tear it down.
       engine = 'base' if target
       # Hardware has explicit teardown to unlock the node
-      engine = 'hardware' if @platform.build_host
+      engine = 'hardware' if @platform.build_hosts
       require "vanagon/engine/#{engine}"
       @engine = Object.const_get("Vanagon::Engine::#{engine.capitalize}").new(@platform, target)
 
     rescue LoadError => e
-      raise Vanagon::Error.wrap(e, "Could not load the desired engine '#{engine}'.\n#{e}")
+      raise Vanagon::Error.wrap(e, "Could not load the desired engine '#{engine}'")
     end
 
     def cleanup_workdir
--- a/lib/vanagon/engine/hardware.rb
+++ b/lib/vanagon/engine/hardware.rb
@@ -4,7 +4,7 @@ require 'lock_manager'
 
 LOCK_MANAGER_HOST = ENV['LOCK_MANAGER_HOST'] || 'redis'
 LOCK_MANAGER_PORT = ENV['LOCK_MANAGER_PORT'] || 6379
-VANAGON_LOCK_USER = "#{ENV['USER']}"
+VANAGON_LOCK_USER = ENV['USER']
 
 
 class Vanagon
@@ -16,30 +16,48 @@ class Vanagon
       # This method is used to obtain a vm to build upon
       # For the base class we just return the target that was passed in
       def select_target
-        Vanagon::Driver.logger.info "Polling for a lock on #{@build_host}."
-        @lockman.polling_lock(@build_host, VANAGON_LOCK_USER, "Vanagon automated lock")
-        Vanagon::Driver.logger.info "Lock acquired on #{@build_host}."
-        warn "Lock acquired on #{@build_host} for #{VANAGON_LOCK_USER}."
-        @target = @build_host
-        @build_host
+        @target = node_lock(@build_hosts)
+      end
+
+      # Poll for a lock
+      def polling_lock(host)
+        Vanagon::Driver.logger.info "Polling for a lock on #{host}."
+        @lockman.polling_lock(host, VANAGON_LOCK_USER, "Vanagon automated lock")
+        Vanagon::Driver.logger.info "Lock acquired on #{host}."
+        puts "Lock acquired on #{host} for #{VANAGON_LOCK_USER}."
+        host
+      end
+
+      # Iterarte over the options and find a node open to lock.
+      def node_lock(hosts)
+        hosts.each do |h|
+          Vanagon::Driver.logger.info "Attempting  to lock #{h}."
+          if @lockman.lock(h, VANAGON_LOCK_USER, "Vanagon automated lock")
+            Vanagon::Driver.logger.info "Lock acquired on #{h}."
+            puts "Lock acquired on #{h} for #{VANAGON_LOCK_USER}."
+            return h
+          end
+        end
+        # If they are all locked, fall back to a polling lock on last item
+        polling_lock(hosts.pop)
       end
 
       # Steps needed to tear down or clean up the system after the build is
       # complete. In this case, we'll attempt to unlock the hardware
       def teardown
-        Vanagon::Driver.logger.info "Removing lock on #{@build_host}."
-        warn "Removing lock on #{@build_host}."
-        @lockman.unlock(@build_host, VANAGON_LOCK_USER)
+        Vanagon::Driver.logger.info "Removing lock on #{@target}."
+        puts "Removing lock on #{@target}."
+        @lockman.unlock(@target, VANAGON_LOCK_USER)
       end
 
       def initialize(platform, target)
         Vanagon::Driver.logger.debug "Hardware engine invoked."
         @platform = platform
-        @build_host = platform.build_host
+        @build_hosts = platform.build_hosts
         # Redis is the only backend supported in lock_manager currently
         @lockman = LockManager.new(type: 'redis', server: LOCK_MANAGER_HOST)
         super
-        @required_attributes << "build_host"
+        @required_attributes << "build_hosts"
       end
     end
   end
--- a/lib/vanagon/engine/hardware.rb
+++ b/lib/vanagon/engine/hardware.rb
@@ -1,0 +1,46 @@
+require 'vanagon/engine/base'
+require 'json'
+require 'lock_manager'
+
+LOCK_MANAGER_HOST = ENV['LOCK_MANAGER_HOST'] || 'redis'
+LOCK_MANAGER_PORT = ENV['LOCK_MANAGER_PORT'] || 6379
+VANAGON_LOCK_USER = "#{ENV['USER']}"
+
+
+class Vanagon
+  class Engine
+    # Class to use when building on a hardware device (e.g. AIX, Switch, etc)
+    #
+    class Hardware < Base
+
+      # This method is used to obtain a vm to build upon
+      # For the base class we just return the target that was passed in
+      def select_target
+        Vanagon::Driver.logger.info "Polling for a lock on #{@build_host}."
+        @lockman.polling_lock(@build_host, VANAGON_LOCK_USER, "Vanagon automated lock")
+        Vanagon::Driver.logger.info "Lock acquired on #{@build_host}."
+        warn "Lock acquired on #{@build_host} for #{VANAGON_LOCK_USER}."
+        @target = @build_host
+        @build_host
+      end
+
+      # Steps needed to tear down or clean up the system after the build is
+      # complete. In this case, we'll attempt to unlock the hardware
+      def teardown
+        Vanagon::Driver.logger.info "Removing lock on #{@build_host}."
+        warn "Removing lock on #{@build_host}."
+        @lockman.unlock(@build_host, VANAGON_LOCK_USER)
+      end
+
+      def initialize(platform, target)
+        Vanagon::Driver.logger.debug "Hardware engine invoked."
+        @platform = platform
+        @build_host = platform.build_host
+        # Redis is the only backend supported in lock_manager currently
+        @lockman = LockManager.new(type: 'redis', server: LOCK_MANAGER_HOST)
+        super
+        @required_attributes << "build_host"
+      end
+    end
+  end
+end
--- a/lib/vanagon/platform/dsl.rb
+++ b/lib/vanagon/platform/dsl.rb
@@ -161,6 +161,13 @@ class Vanagon
         @platform.servicetype = type
       end
 
+      # Set the name of host to perform a build on (when not using pooler or CLI flags)
+      #
+      # @param type [String] the name of the host (must be resolvable)
+      def build_host(name)
+        @platform.build_host = name
+      end
+
       # Set the name of this platform as the vm pooler expects it
       #
       # @param name [String] name of the target template to use from the vmpooler
--- a/lib/vanagon/platform/dsl.rb
+++ b/lib/vanagon/platform/dsl.rb
@@ -161,12 +161,18 @@ class Vanagon
         @platform.servicetype = type
       end
 
-      # Set the name of host to perform a build on (when not using pooler or CLI flags)
+      # Set the list of possible host to perform a build on (when not using
+      # pooler or CLI flags)
       #
-      # @param type [String] the name of the host (must be resolvable)
-      def build_host(name)
-        @platform.build_host = name
+      # @param type [Array] the names of the hosts (must be resolvable)
+      # @rase ArgumentError if builds_hosts has no arguments
+      def build_hosts(*args)
+        raise ArgumentError, "build_hosts requires at least one host to be a build target." if args.empty?
+        @platform.build_hosts = Array(args).flatten
       end
+
+      # Because single vs plural is annoying to remember
+      alias_method :build_host, :build_hosts
 
       # Set the name of this platform as the vm pooler expects it
       #
--- a/Gemfile
+++ b/Gemfile
@@ -1,9 +1,20 @@
 source ENV['GEM_SOURCE'] || "https://rubygems.org"
 
+def lock_manager_location_for(place)
+  if place =~ /^(git[:@][^#]*)#(.*)/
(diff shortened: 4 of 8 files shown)


geoffnichols:
If we wanted to obtain a lock on one of multiple build hosts, could we also do that here? 

For example, if given something like:

`plat.build_host_pool ["pe-aix-71-builder.delivery.puppetlabs.net", "pe-aix-71-builder2.delivery.puppetlabs.net"]`,

iterate over those until able to acquire a lock, then set `build_host` to the target we are able to lock.

stahnma:
@geoffnichols no. Right now the platform definition ties to a host 1:1. That's what we have. If we had a pool would you just check and see which one you could get a lock on and move forward on that?

geoffnichols:
@stahnma yes, exactly

geoffnichols:
@stahnma this is a tangent, but I that capability (check a pool, see which one you can get a lock on) might also be useful in beaker?

haus:
@stahnma what if your build_host was an array and you tried all of the elements, attempting to get a lock? that might help address @geoffnichols concern/idea.

lib/vanagon/driver.rb
```diff
@@ -23,16 +23,22 @@ def initialize(platform, project, options = { :configdir => nil, :target => nil,
...
+      @@logger = Logger.new('vanagon_hosts.log')
+      @@logger.progname = 'vanagon'
 
       # If a target has been given, we don't want to make any assumptions about how to tear it down.
       engine = 'base' if target
+      # Hardware has explicit teardown to unlock the node
+      engine = 'hardware' if @platform.build_host
       require "vanagon/engine/#{engine}"
-      @engine = Object.const_get("Vanagon::Engine::#{engine.capitalize}").new(@platform, target)
+      if engine == 'hardware'
```
haus: You shouldn't need this conditional. If you look at what the pooler engine does, it takes 2 parameters to the constructor and just ignores the second.

lib/vanagon/engine/hardware.rb
```diff
@@ -0,0 +1,102 @@
+require 'vanagon/engine/base'
+require 'json'
+require 'redis'
+
+REDIS_HOST = 'redis.delivery.puppetlabs.net'
```
haus: This makes this unusable for anyone else. This should be moved to some sort of configuration (.vanagonrc or something), or at the very least environment variables.

lib/vanagon/engine/hardware.rb
```diff
@@ -0,0 +1,102 @@
...
+        @required_attributes << "build_host"
+        @build_host = platform.build_host
+        @target = @build_host
+        aquire_lock
+      end
+
+      # Lock the build_host as specificied in the platform definition. This
+      # implement a rudimentary backoff sequence in order to wait on the host
+      # to unlock. Note: this polls forever.
+      def aquire_lock
```
haus: aquire => acquire

lib/vanagon/engine/hardware.rb
```diff
@@ -0,0 +1,102 @@
...
+          sleep sleepcount
+          temp = sleepcount
+          sleepcount += psleepcount
+          psleepcount = temp
+        end
+        lock_contents = {}
+        lock_contents[:user] = LOCK_USER
+        lock_contents[:time] = Time.now.to_s
+        lock_contents[:reason] = "Vanagon build on #{@build_host}"
+        redis = connect_redis
```
haus: the redis connection should probably be an instance variable that is initialized once and reused. right now each time you check for the lock you open a new connection to redis.

lib/vanagon/engine/hardware.rb
```diff
@@ -0,0 +1,102 @@
...
+        temp = 0
+        while locked?
+          print "#{@build_host} is locked..."
+          puts "waiting #{sleepcount} seconds."
+          sleep sleepcount
+          temp = sleepcount
+          sleepcount += psleepcount
+          psleepcount = temp
+        end
+        lock_contents = {}
```
haus: I think the lock setting should probably be broken out into a method of its own.

lib/vanagon/engine/hardware.rb
```diff
@@ -0,0 +1,102 @@
...
+
+      # Lock the build_host as specificied in the platform definition. This
+      # implement a rudimentary backoff sequence in order to wait on the host
+      # to unlock. Note: this polls forever.
+      def aquire_lock
+        # Wait if locked
+        if locked?
+          warn "Unable to aquire lock."
+        end
+        sleepcount = 1
```
haus: Instead of this, method of sleeping, why not set sleep_duration to 1 and then double it each loop? That would clean up a lot of the variables that are in use currently. Could look something like

```
sleep_duration = 1
while locked?
  ...
  sleep sleep_duration
  sleep_duration *= 2
end
```

lib/vanagon/engine/hardware.rb
```diff
@@ -0,0 +1,102 @@
...
+
+      # Steps needed to tear down or clean up the system after the build is
+      # complete. In this case, we'll attempt to unlock the hardware
+      def teardown
+        release_lock
+      end
+
+      def initialize(platform)
+        # see if we can aquire a lock
+        Vanagon::Driver.logger.debug "Hardware engine invoked."
```
haus: I don't think this message actually belongs in the vanagon_hosts logfile. However, a message in select_target would be appropriate. Then the message could be something like:

`Vanagon::Driver.logger.info "Locked and checked out #{@build_host}"`

Then the teardown could say something about unlocking and returning `@build_host`.

lib/vanagon/engine/hardware.rb
```diff
@@ -0,0 +1,102 @@
...
+        @platform = platform
+        super
+        @required_attributes << "build_host"
+        @build_host = platform.build_host
```
haus: This logic belongs in select_target, which should be responsible for checking and getting the lock. select_target is invoked as part of the engine startup.

lib/vanagon/engine/hardware.rb
```diff
@@ -0,0 +1,102 @@
...
+        @target = @build_host
+        aquire_lock
+      end
+
+      # Lock the build_host as specificied in the platform definition. This
+      # implement a rudimentary backoff sequence in order to wait on the host
+      # to unlock. Note: this polls forever.
+      def aquire_lock
+        # Wait if locked
+        if locked?
```
haus: This looks like it might do better as a basic loop and not a while loop, as locked? is called twice.

That would look something like...

```
loop do
  if locked?
    ...
  else
    break
  end
end
```

lib/vanagon/engine/hardware.rb
```diff
@@ -0,0 +1,102 @@
...
+          temp = sleepcount
+          sleepcount += psleepcount
+          psleepcount = temp
+        end
+        lock_contents = {}
+        lock_contents[:user] = LOCK_USER
+        lock_contents[:time] = Time.now.to_s
+        lock_contents[:reason] = "Vanagon build on #{@build_host}"
+        redis = connect_redis
+        redis.set "node_lock_#{@build_host.split('.')[0]}", lock_contents.to_json
```
haus: redis.set can raise exceptions in certain cases. This should probably handle at least CommandError exceptions.

lib/vanagon/engine/hardware.rb
```diff
@@ -0,0 +1,102 @@
...
+        Vanagon::Driver.logger.info "Hardware locked: #{@build_host}"
+      end
+
+      # Release the lock on build_host
+      def release_lock
+        redis = connect_redis
+        r = redis.get("node_lock_#{@build_host.split('.')[0]}")
+        if r
+          result  = JSON.parse(r)
+          if result['user'] != LOCK_USER
```
haus: In what circumstances would this happen?

lib/vanagon/engine/hardware.rb
stahnma: Like a method to actually just do the lock? Pass the json contents to it, or build it out there via params?

lib/vanagon/engine/hardware.rb
```diff
@@ -0,0 +1,102 @@
...
+          end
+        end
+      end
+
+      # Boolean to figure out if a host is locked.
+      #
+      # @return [Bool] whether or not the host is lockd.
+      def locked?
+        redis = connect_redis
+        return true if redis.get("node_lock_#{@build_host.split('.')[0]}")
```
haus: you should be able to just return the redis.get directly. If the key doesn't exist, redis.get will return nil, which is falsey. Otherwise it will return the string, which is truthy.

lib/vanagon/engine/hardware.rb
stahnma: Which part? I can move the call to acquire_lock in there. Should the \@taget get set there as well?

lib/vanagon/engine/hardware.rb
```diff
@@ -0,0 +1,102 @@
...
+        redis = connect_redis
+        redis.set "node_lock_#{@build_host.split('.')[0]}", lock_contents.to_json
+        puts "Hardware locked: #{@build_host}"
+        Vanagon::Driver.logger.info "Hardware locked: #{@build_host}"
+      end
+
+      # Release the lock on build_host
+      def release_lock
+        redis = connect_redis
+        r = redis.get("node_lock_#{@build_host.split('.')[0]}")
```
haus: It seems like using the FQDN here will be better than splitting, unless kerminator and the other tools are also splitting on `.` before locking (which they aren't). That way a user that locks pe-aix-53-builder.delivery.puppetlabs.net won't be surprised when their build gets clobbered.

lib/vanagon/engine/hardware.rb
haus: I was thinking params. Like set_lock(user, time, reason) or whatever.

lib/vanagon/engine/hardware.rb
haus: yea. if you look at how the pooler does it, it should make sense.

lib/vanagon/engine/hardware.rb
stahnma: Yeah, this was an issue I ran into when testing. I think all locks should split at the dot. Otherwise a lock on pe-aix-71-builder won't match pe-aix-71-builder.delivery.puppetlabs.net. That's not so good. I was going to jump in the kerminator code and fix that.

lib/vanagon/engine/hardware.rb
```diff
@@ -0,0 +1,100 @@
...
+      # @return [Bool] whether or not the host is lockd.
+      def locked?
+        return true if @redis.get("node_lock_#{@build_host.split('.')[0]}")
+        false
+      end
+
+
+      # Create a connection to redis.
+      def connect_redis
+        Redis.new(:host => REDIS_HOST, :port => REDIS_PORT)
```
haus: Probably worth checking for exceptions here.

lib/vanagon/engine/hardware.rb
haus: So this would be `set_lock(LOCK_USER, Time.now.to_s, "Vanagon build on #{@build_host}"` or something.

lib/vanagon/engine/hardware.rb
haus: That makes sense to me. As long as all of the lock entry points do the same thing it'll be fine (and as long as we don't have name collisions).

stahnma: PR for kerminator: https://github.com/puppetlabs/kerminator/pull/227

stahnma:
@geoffnichols I like your idea, but it would change the way a bit of this works. I could probably implement it, but I'd like to get a first pass done first.

stahnma:
Items still outstanding: Loading the lock_manager gem properly. I am not sure if I have done that correctly in the Gemfile and gemspec. In my early testing with puppet-agent, I also had to add it to that Gemfile, which is less than ideal.

shrug:
@stahnma needs a rebase? I think we're ready to get this in now?

shrug:
Also failing a test...

stahnma:
@geoffnichols @haus implemented the multiple hosts for the hardware engine idea in the second commit I just put up.

lib/vanagon/driver.rb
```diff
@@ -23,16 +23,18 @@ def initialize(platform, project, options = { :configdir => nil, :target => nil,
...
+      # Hardware has explicit teardown to unlock the node
+      engine = 'hardware' if @platform.build_hosts
       require "vanagon/engine/#{engine}"
       @engine = Object.const_get("Vanagon::Engine::#{engine.capitalize}").new(@platform, target)
 
-      @@logger = Logger.new('vanagon_hosts.log')
-      @@logger.progname = 'vanagon'
     rescue LoadError => e
-      raise Vanagon::Error.wrap(e, "Could not load the desired engine '#{@engine_name}'.")
+      raise Vanagon::Error.wrap(e, "Could not load the desired engine '#{engine}'.\n#{e}")
```
haus: You don't need to add `#{e}` to the message here, that is what the wrap does.

lib/vanagon/engine/hardware.rb
```diff
@@ -0,0 +1,72 @@
+require 'vanagon/engine/base'
+require 'json'
+require 'lock_manager'
+
+LOCK_MANAGER_HOST = ENV['LOCK_MANAGER_HOST'] || 'redis'
+LOCK_MANAGER_PORT = ENV['LOCK_MANAGER_PORT'] || 6379
+VANAGON_LOCK_USER = "#{ENV['USER']}"
```
haus: You don't need quotes around this or interpolation. ENV entries are always strings or nil.

lib/vanagon/engine/hardware.rb
```diff
@@ -0,0 +1,72 @@
...
+          @target = multi_node_lock(@build_hosts)
+        end
+      end
+
+      # Go through the lock process for a single node
+      def single_node_lock(host)
+        Vanagon::Driver.logger.info "Polling for a lock on #{host}."
+        @lockman.polling_lock(host, VANAGON_LOCK_USER, "Vanagon automated lock")
+        Vanagon::Driver.logger.info "Lock acquired on #{host}."
+        warn "Lock acquired on #{host} for #{VANAGON_LOCK_USER}."
```
haus: Why use warn?

lib/vanagon/platform/dsl.rb
```diff
@@ -161,6 +161,22 @@ def servicetype(type)
...
       end
 
+      # Set the list of possible host to perform a build on (when not using
+      # pooler or CLI flags)
+      #
+      # @param type [Array] the names of the hosts (must be resolvable)
+      def build_hosts(*args)
+        raise ArgumentError if args.length < 1
+        if args[0].is_a?(Array)
+          @platform.build_hosts = args[0] if args[0].is_a?(Array)
```
haus: there are 2 `if args[0].is_a?(Array)` here. you probably only want one.

lib/vanagon/engine/hardware.rb
```diff
@@ -0,0 +1,72 @@
...
+      def select_target
+        if @build_hosts.length == 1
+          @target = single_node_lock(@build_hosts[0])
+        else
+          @target = multi_node_lock(@build_hosts)
+        end
+      end
+
+      # Go through the lock process for a single node
+      def single_node_lock(host)
```
haus: This seems like an unnecessary special case. Even for 1 node you can still do `hosts.each`. In the multi_node_lock you can put in a timeout on a polling lock and then there would be just one locking method.

lib/vanagon/engine/hardware.rb
```diff
@@ -0,0 +1,72 @@
...
+class Vanagon
+  class Engine
+    # Class to use when building on a hardware device (e.g. AIX, Switch, etc)
+    #
+    class Hardware < Base
+
+      # This method is used to obtain a vm to build upon
+      # For the base class we just return the target that was passed in
+      def select_target
+        if @build_hosts.length == 1
```
haus: As I mentioned below, I'd recommend just using 1 lock method. Then this code gets a lot simpler.

lib/vanagon/engine/hardware.rb
stahnma: warn is on stdout. Logger goes into a file.

lib/vanagon/platform/dsl.rb
stahnma: Yup, refactor FTL.

lib/vanagon/engine/hardware.rb
haus: Okay, i guess i'm asking why stderr here when it doesn't seem errorful.

lib/vanagon/platform/dsl.rb
```diff
@@ -161,6 +161,22 @@ def servicetype(type)
         @platform.servicetype = type
       end
 
+      # Set the list of possible host to perform a build on (when not using
+      # pooler or CLI flags)
+      #
+      # @param type [Array] the names of the hosts (must be resolvable)
+      def build_hosts(*args)
+        raise ArgumentError if args.length < 1
```
haus: args.empty? may be clearer here.

lib/vanagon/platform/dsl.rb
```diff
@@ -161,6 +161,22 @@ def servicetype(type)
...
+        if args[0].is_a?(Array)
```
haus: This can be simplified to:
`@platform.build_hosts = Array(args).flatten`

That way is args is an array, it will get re-arrayed and then de-arrayed, and if it is just a string it will be an array of 1 element.

lib/vanagon/engine/hardware.rb
stahnma: You'd rather just puts?

lib/vanagon/platform/dsl.rb
stahnma: Sweet. I was trying to think of a better way to do that.

lib/vanagon/platform/dsl.rb
stahnma: Can do.

lib/vanagon/engine/hardware.rb
haus: I think puts makes more sense to me, unless/until there is a clear breakdown of where messages go during a build

lib/vanagon/platform/dsl.rb
```diff
@@ -161,6 +161,19 @@ def servicetype(type)
         @platform.servicetype = type
       end
 
+      # Set the list of possible host to perform a build on (when not using
+      # pooler or CLI flags)
+      #
+      # @param type [Array] the names of the hosts (must be resolvable)
+      def build_hosts(*args)
+        raise ArgumentError if args.empty?
```
haus: Also having some kind of text in the ArgumentError would be good, so the error is meaningful.

The yard doc should also mention the `@raise`.

underscorgan:
:+1:

lib/vanagon/engine/hardware.rb
```diff
@@ -0,0 +1,64 @@
...
+        hosts.each do |h|
+          Vanagon::Driver.logger.info "Attempting  to lock #{h}."
+          if @lockman.lock(h, VANAGON_LOCK_USER, "Vanagon automated lock")
+            Vanagon::Driver.logger.info "Lock acquired on #{h}."
+            puts "Lock acquired on #{h} for #{VANAGON_LOCK_USER}."
+            return h
+          end
+        end
+        # If they are all locked, fall back to a polling lock on last item
+        polling_lock(hosts.pop)
```
McdonaldSeanp: @stahnma is there a specific reason that the function only passes once? it seems like it might be better to just loop around and keep trying them all again and again.

stahnma: The polling lock blocks. The attempt to lock others does not block, and is VERY fast with no back off. 

The main reason is that the original design only handled 1 host, not N. Now that we have N, it seems like most of the time we'll just grab one in the list. If they are all full, just wait. I agree it's not totally amazing, but it also probably won't happen that often.

McdonaldSeanp: Makes sense.

McdonaldSeanp:
:+1:

shermdog:
On the multihost polling fallback, what is the likelihood we'll have several builds all waiting on the last node to become available at once?  It seems slightly counter intuitive to go from being able to execute on N hosts to polling a single host.

stahnma:
@shermdog yeah. The likelihood isn't that high, as we normally don't build the same target at the same time. I do agree the logic is a little odd, but I'm kind of at the "perfect is the enemy of done" portion of the show.

shermdog:
Then I'm :+1: and will gladly push the button if everyone else is done beating you up :)

PR #230, opened 2015-12-07
(CPR-239) On Solaris 10, vanagon projects change group ownership fro…

McdonaldSeanp:
…m 'bin' to 'sys'

Before this commit vanvagon runs would cause the permissions of the /lib /lib/svc and /lib/svc/method directories to change from "bin" to "sys". This was due to incorrectly generating the solaris 10 prototype file with the pkgproto command. This commit will fix the issue by only including puppet directories, thus implicitly excluding dirs such as root/lib.

NOTE**** This fully fixes the problem but has opened a seperate problem: https://tickets.puppetlabs.com/browse/CPR-259


mckern:
(Quick, when no one is looking, truncate your commit subject to less than 70 chars)

mckern:
@McdonaldSeanp Have you opened a CPR ticket for the incorrect prototype file? If you open a ticket, then you can remove this note or reference that in this commit message instead.

lib/vanagon/platform/solaris_10.rb
```diff
@@ -42,14 +42,16 @@ def generate_package(project)
           # - The bin directory and all bin files are owned by root:bin instead of root:sys
           # - All files under lib are owned by root:bin instead of root:sys
           # - All .so files are owned by root:bin instead of root:sys
-          %((cd $(tempdir)/#{name_and_version}; pkgproto . | sort | awk ' \
```
mckern: Is the change from running this command pipeline inside a subshell to running it inline intentional?

McdonaldSeanp:
I did not open a ticket as I did not know how imperative it was that we "fix" this. Everything works perfect as it is, we are just not "technically" doing things right so I'm not sure if the extra effort would be necessary

lib/vanagon/platform/solaris_10.rb
McdonaldSeanp: I didn't write that, so youll have to ask @haus

mckern:
@McdonaldSeanp it's always worth opening the ticket, even if you don't have the time to resolve it right away; capturing the work to be done is as important as doing it. We probably won't be able to pull it in this sprint but it's a good exercise.

lib/vanagon/platform/solaris_10.rb
mckern: @McdonaldSeanp your commit is the one that removes the extra set of parentheses; what is the effect (if any) of removing that second set of parens?

McdonaldSeanp: Oops, That is true. I didn't mean to remove the extra parens, though it didn't have any effect on the output, the build still worked and installing on solaris-10 still worked fine. However for posterity I will reintroduce the parens

lib/vanagon/platform/solaris_10.rb
```diff
@@ -42,14 +42,16 @@ def generate_package(project)
           # - The bin directory and all bin files are owned by root:bin instead of root:sys
           # - All files under lib are owned by root:bin instead of root:sys
           # - All .so files are owned by root:bin instead of root:sys
+          # - Explicity only include puppet directories 
+          #   (this should exclude things like root/bin root/var and such)
           %((cd $(tempdir)/#{name_and_version}; pkgproto . | sort | awk ' \
-            $$1 ~ /^d$$/ {print "d",$$2,$$3,"0755 root sys";} \
+            $$1 ~ /^d$$/ && $$3 ~ /^.*puppet.*/ {print "d",$$2,$$3,"0755 root sys";} \
```
haus: As this is a generic application, non-specific to puppet, it would be better to not special case puppet here.

Vanagon projects already register directories that should be included in the packages, so maybe that would be useful here.

McdonaldSeanp: Thanks @haus

haus: Below on line https://github.com/puppetlabs/pe_acceptance_tests/blob/2015.3.x/acceptance/tests/file-sync/file-sync_commit.rb you can see `project.get_directories` is called. That will return a list of directories, which have a path attribute, which is probably what you care about most.

McdonaldSeanp: Yeah, I just found that function actually, thanks @haus, I was thinking this might be bad and I hoped someone would point me in the right direction when I opened the PR.

PR #231, opened 2015-12-09
(maint) Allow environment to be specified when unpacking tarballs

melissa:
On windows, we need to be able to set certain environment variables
prior to unpacking our source. The is mainly for dealing with sticky
things like symlinks in cygwin versus windows environments. However,
this only works when we actually have something to unpack. Otherwise,
our extract command returns an empty string. In order for the makefile
not to fail, we return a simple ':' when there is nothing to extract.


MikaelSmith:
:+1:

PR #232, opened 2015-12-10
(CPR-239) On Solaris 10, vanagon projects change group ownership from…

McdonaldSeanp:
… 'bin' to 'sys'

Before this commit vanvagon runs would cause the permissions of the /lib /lib/svc and /lib/svc/method directories to change from "bin" to "sys". This was due to incorrectly generating the solaris 10 prototype file with the pkgproto command. This commit will fix the issue by only including puppet directories, thus implicitly excluding dirs such as root/lib.

NOTE**** This issue has been resolved and Vanagon is full operational but it has opened the following: https://tickets.puppetlabs.com/browse/CPR-259

### DIFF
--- a/lib/vanagon/platform/solaris_10.rb
+++ b/lib/vanagon/platform/solaris_10.rb
@@ -10,6 +10,19 @@ class Vanagon
         name_and_version = "#{project.name}-#{project.version}"
         pkg_name = package_name(project)
 
+        # Generate list of dirs in the package and create an exlplicit search
+        # string for AWK to use in order to explicitly define which directories
+        # to put in to the prototype file. Note that the Regexp object was avoided
+        # because the output from Regexp would create failures from AWK as the conversion
+        # from a Regexp obj to a string is formatted in a sub-optimal way that would have
+        # required more string manipulation anyway. the string should be formatted like so:
+        #         && ($$3 ~ /directory\/regex.*/ || $$3 ~ /another\/directory\/regex.*/)
+        # for as many iterations as there are directries in the package
+        pkgdirs = project.get_root_directories.map { |dir| dir.sub(/^\//, "").gsub(/([\/\.])+/, '\\\\\1') + '.*' }
+        explicit_search_string = pkgdirs.map do |dir_regex|
+          " $$3 ~ /" + dir_regex + "/ "
+        end.join("||")
+
         # Here we maintain backward compatibility with older vanagon versions
         # that did this by default.  This shim should get removed at some point
         # in favor of just letting the makefile deliver the bill-of-materials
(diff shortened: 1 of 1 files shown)


stahnma:
@haus  can we get a review from you on this one?

haus:
@McdonaldSeanp can you gist the makefile from a build based off of this change? I'm interested in what this script looks like after it's generated.

McdonaldSeanp:
@haus do you want the whole makefile or just the bit that this changes?

McdonaldSeanp:
Here is a gist including just that one part of the script, let me know if you need anything else: https://gist.github.com/McdonaldSeanp/4c26a1c2bb99f66fe76a

McdonaldSeanp:
Also a note on why this doesn't just make a big union: if you just try to make one big union out of all the dirs and just do $$3 ~ #{some huge union} AWK will segfault. I don't know why, and theres no reason why if you union too many things together it would segfault byut that's what I was seeing, so instead I had to do it this way

McdonaldSeanp:
https://gist.github.com/McdonaldSeanp/86115fa13af448a41e64

Thats the whole thing, the line is line 18

stahnma:
@haus another look?

haus:
Looks functional. I gave @McdonaldSeanp some ruby style points in meatspace, so I'll wait for that to appear and then I think this is gtg.

McdonaldSeanp:
@haus I changed what we talked about a little b/c I think it makes a little more sense this way (specifically it looks better to have && (#{explicit_search_string}) in the awk script rather than put the && ( ) in the explicit search string. functionally they are identical though, and I tested it and everything works

underscorgan:
Do you have a before/after for the makefile? It's difficult for me to parse out the regex

lib/vanagon/platform/solaris_10.rb
```diff
@@ -10,6 +10,19 @@ def generate_package(project)
...
 
+        # Generate list of dirs in the package and create an exlplicit search
+        # string for AWK to use in order to explicitly define which directories
+        # to put in to the prototype file. Note that the Regexp object was avoided
+        # because the output from Regexp would create failures from AWK as the conversion
+        # from a Regexp obj to a string is formatted in a sub-optimal way that would have
+        # required more string manipulation anyway. the string should be formatted like so:
+        #         && ($$3 ~ /directory\/regex.*/ || $$3 ~ /another\/directory\/regex.*/)
+        # for as many iterations as there are directries in the package
+        pkgdirs = project.get_root_directories.map { |dir| dir.sub(/^\//, "").gsub(/[\/]/, "\\/").gsub(/\./, "\\.") + '.*' }
```
haus: The two gsubs can be combined, as all you are adding is two slashes in front of them.

`gsub(/([\/\.])+/, '\\\\\1')

haus: Also, I think the `sub(/^\//, '')` pattern appears often enough in vanagon it would be worth tossing into https://github.com/puppetlabs/vanagon/blob/master/lib/vanagon/extensions/string.rb

McdonaldSeanp:
@underscorgan Here is the before Makefile https://gist.github.com/McdonaldSeanp/049f44a4f2b68a38ed0b

underscorgan:
@McdonaldSeanp do you have a link to a successful puppet-agent build or something for sol10?

McdonaldSeanp:
Yes, here is a successful build on jenkins of the updated vanagon: http://jenkins-release.delivery.puppetlabs.net/job/vanagon_generic_job/233/

underscorgan:
:+1: 

Pulled up a sol10 vm and checked the output of ```pkgchk -d <pkg> -l``` between this build and an older puppet-agent build, and looks like this removes the management of the offending high-level directories and nothing else suspect, so YAY.

McdonaldSeanp:
Yep, exactly as it should @underscorgan I'm glad to see it verified

mckern:
@McdonaldSeanp I'm not merging this until we update your Gravitar and make the green sprite thing go away.

robot devil

McdonaldSeanp:
There, done

mckern:
bravo

PR #233, opened 2015-12-10
Add helper for out of tree builds

adrienthebo:
Tools like cmake use out of tree builds to prevent source files and
build files from intermingling; when they're used all commands are run
from a separate build directory. This commit adds a #build_dir method to
the component (and DSL) that when set will create the build dir relative
to the source directory for each component, and when running any build
command will cd into the build directory before running the command.

### DIFF
--- a/templates/Makefile.erb
+++ b/templates/Makefile.erb
@@ -79,8 +79,11 @@ cleanup-components: <%= @components.map {|comp| "#{comp.name}-cleanup" }.join(" 
 	touch <%= comp.name %>-patch
 
 <%= comp.name %>-configure: <%= comp.name %>-patch <%= list_component_dependencies(comp).join(" ") %>
+	<%- if comp.get_build_dir -%>
+	[ -d <%= comp.get_build_dir %> ] || mkdir -p <%= comp.get_build_dir %>
+	<%- end -%>
 	<%- unless comp.configure.empty? -%>
-	cd <%= comp.dirname %> && \
+	cd <%= comp.get_build_dir %> && \
 	<%= comp.get_environment %> && \
 	<%= comp.configure.join(" && \\\n\t") %>
 	<%- end -%>
@@ -88,7 +91,7 @@ cleanup-components: <%= @components.map {|comp| "#{comp.name}-cleanup" }.join(" 
 
 <%= comp.name %>-build: <%= comp.name %>-configure
 	<%- unless comp.build.empty? -%>
-	cd <%= comp.dirname %> && \
+	cd <%= comp.get_build_dir %> && \
 	<%= comp.get_environment %> && \
 	<%= comp.build.join(" && \\\n\t") %>
 	<%- end -%>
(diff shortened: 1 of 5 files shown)


branan:
Looks reasonable to me. I've been wanting to start using proper build dirs for a while (it makes jumping into a Vanagon tmp dir and doing manual debugging so much easier).

templates/Makefile.erb
```diff
@@ -79,8 +79,11 @@ cleanup-components: <%= @components.map {|comp| "#{comp.name}-cleanup" }.join("
 	touch <%= comp.name %>-patch
 
 <%= comp.name %>-configure: <%= comp.name %>-patch <%= list_component_dependencies(comp).join(" ") %>
+	<%- if comp.get_build_dir -%>
+	[ -d <%= comp.get_build_dir %> ] || mkdir <%= comp.get_build_dir %>
```
mckern: Testing for existence isn't strictly necessary here since mkdir will still exit successfully if the directory already exists. (Nitpick, not blocking).

mckern:
1 nitpick, 1 outstanding question, and :+1: overall.

templates/Makefile.erb
adrienthebo: It looks like mkdir will actually fail if the directory exists: 

```
└> mkdir foo; mkdir foo; echo $?
mkdir: cannot create directory ‘foo’: File exists
1
```

+1 for `mkdir -p`; adding it now.

mckern: Ugh, I bet I was using some variant `mkdir` (bsd, busybox, whatever) and it just exited successfully. Disregard nitpick; safety first!

mckern: Oh, I was using GNU `mkdir`. Adorable!

```
mckern@flexo vanagon $ which mkdir
/opt/local/opt/coreutils/libexec/gnubin/mkdir
mckern@flexo vanagon $ mkdir bin
mckern@flexo vanagon $ echo $?
0
```

adrienthebo: Safety third!

mckern: Ah, here we go:

```
mckern@flexo vanagon $ type mkdir
mkdir is aliased to `mkdir -p'
mckern@flexo vanagon $ \mkdir bin
mkdir: cannot create directory ‘bin’: File exists
```

So, `mkdir` fails if a directory exists but `mkdir -p` does not. Today we learned something.

adrienthebo: Ahhh, I was about to comment - I ran this on Arch Linux and EL6 and it had this behavior. Weird about `mkdir -p`. But yeah - adding the check makes life easier for all involved.

branan: I believe this difference in behavior of `mkdir` and `mkdir -p` is actually part of POSIX. I can try to find the citation if you like.

Aliasing `mkdir` to `mkdir -p` is right up there with aliasing `rm` to `rm -i` - it encourages dumb behavior and fails to teach anyone how the system actually works.

adrienthebo:
The build (finally) passed, who would be the best person to pester to get this merged and released?

adrienthebo:
Thanks!

PR #234, opened 2015-12-15
(maint) Handle git versioning scheme in nuspec template

melissa:
Nuget is challenging when it comes to versions. We can only have four
numeric elements seperated by periods. We have the option of specifying
a build version, but this must be the last element in the version tag
and must be a dash followed immediately by a letter and then either
numbers or letters but no punctuation. Since we cannot guarentee a git
short sha will begin with a letter, this isn't an options.

As a result, we must have a version of at most four elements containing
only numbers. If we only have three elements in the version string, then
we get to use the release variable. Otherwise, we just just the first
four elements from the version.

Gross.

### DIFF
--- a/lib/vanagon/platform/windows.rb
+++ b/lib/vanagon/platform/windows.rb
@@ -72,7 +72,7 @@ class Vanagon
         "cp file-list $(tempdir)/#{project.name}/tools/file-list.txt",
         "gunzip -c #{project.name}-#{project.version}.tar.gz | '#{@tar}' -C '$(tempdir)/#{project.name}/tools' --strip-components 1 -xf -",
         "(cd $(tempdir)/#{project.name} ; #{self.drive_root}/ProgramData/chocolatey/bin/choco.exe pack #{project.name}.nuspec)",
-        "cp $(tempdir)/#{project.name}/#{project.name}-#{@architecture}.#{project.version}.#{project.release}.nupkg ./output/#{target_dir}/#{nuget_package_name(project)}"]
+        "cp $(tempdir)/#{project.name}/#{project.name}-#{@architecture}.#{nuget_package_version(project.version, project.release)}.nupkg ./output/#{target_dir}/#{nuget_package_name(project)}"]
       end
 
       # Method to derive the package name for the project.
@@ -85,7 +85,50 @@ class Vanagon
       # @param project [Vanagon::Project] project to name
       # @return [String] name of the windows package for this project
       def nuget_package_name(project)
-        "#{project.name}-#{@architecture}-#{project.version}.#{project.release}.nupkg"
+        "#{project.name}-#{@architecture}-#{nuget_package_version(project.version, project.release)}.nupkg"
+      end
+
+      # Nuget versioning is awesome!
+      #
+      # Nuget and chocolatey have some expectations about version numbers.
+      #
+      # First, if this is a final package (built from a tag), the version must
+      # only contain digits with each element of the version separated by
+      # periods.
+      #
+      # If we are creating the version for a prerelease package (built from a
+      # commit that does not have a corresponding tag), we have the
+      # option to append a string to the version. The string must start with a
+      # letter, be separated from the rest of the version with a dash, and
+      # contain no punctuation.
+      #
+      # We assume we are working from a semver tag as the base of our version.
+      # If this is a final release, we only have to worry about that tag. We
+      # can also include the release number in the package version. If this is
+      # a prerelease package, then we assume we have a semver compliant tag,
+      # followed by the number of commits beyond the tag and the short sha of
+      # the latest change. Because we are working with git, if the version
+      # contains a short sha, it will begin with 'g'. We check for this to
+      # determine what version type to deliver.
+      #
+      # Examples of final versions:
+      #   1.2.3
+      #   1.5.3.1
+      #
+      # Examples of prerelease versions:
+      #   1.2.3.1234-g124dm9302
+      #   3.2.5.23-gd329nd
+      #
+      # @param project [Vanagon::Project] project to version
+      # @return [String] the version of the nuget package for this project
+      def nuget_package_version(version, release)
+        version_elements = version.split('.')
+        if version_elements[-1].start_with?('g')
+          # Version for the prerelease package
+          "#{version_elements[0..-2].join('.').gsub(/[a-zA-Z]/, '')}-#{version_elements[-1]}"
+        else
+          "#{version}.#{release}".gsub(/[a-zA-Z]/, '')
+        end
       end
 
       # Add a repository (or install Chocolatey)
--- a/templates/windows/project.nuspec.erb
+++ b/templates/windows/project.nuspec.erb
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id><%= "#{@name}-#{@platform.architecture}" %></id>
-    <version><%= "#{@version}.#{@release}" %></version>
+    <version><%= @platform.nuget_package_version(@version, @release) %></version>
     <title><%= @name %></title>
     <authors><%= @vendor.match(/^(.*) <.*>$/)[1] %></authors>
     <owners><%= @vendor.match(/^(.*) <.*>$/)[1] %></owners>


melissa:
ping @MikaelSmith @ferventcoder

ferventcoder:
> We have the option of specifying
a build version, but this must be the last element in the version tag
and must be a dash followed immediately by a letter and then either
numbers or letters but no punctuation. Since we cannot guarentee a git
short sha will begin with a letter, this isn't an options.

@melissa Actually, I'm not sure this is the case.  Here's the deal. When you have a `-` somewhere in the version, it considers that version a prerelease. It treats all text to the right of the dash as the prerelease name, which some folks use as a build version. As to whether that should be a number or a letter, I'm not completely sure, but I think it is fine for it to begin with a number.

> As a result, we must have a version of at most four elements containing
only numbers. If we only have three elements in the version string, then
we get to use the release variable. Otherwise, we just just the first
four elements from the version.

You may need to explain this one a bit more to me, you can absolutely have that prerelease string after any number of version elements.  Here's proof that you can use four elements and then a prerelease - https://chocolatey.org/packages/chocolatey/0.9.8.31-beta2

ferventcoder:
Here's a followup for you - https://docs.nuget.org/create/versioning#prerelease-versions 

Note that this is called prereleases. Prereleases are only installed by either specifying the `--version` switch with that exact version or by using `--pre` - by default NuGet and thus Chocolatey will only grab the latest stable version.

lib/vanagon/platform/windows.rb
```diff
@@ -85,7 +85,27 @@ def generate_nuget_package(project)
       # @param project [Vanagon::Project] project to name
       # @return [String] name of the windows package for this project
       def nuget_package_name(project)
-        "#{project.name}-#{@architecture}-#{project.version}.#{project.release}.nupkg"
+        "#{project.name}-#{@architecture}-#{nuget_package_version(project)}.nupkg"
+      end
+
+      # Nuget versioning is awesome!
```
ferventcoder: :trollface:

templates/windows/project.nuspec.erb
```diff
@@ -2,7 +2,40 @@
...
+    Nuget doesn't handle versions well. We have the option of specifying a
+    development release, but that requires a dash followed first by a letter,
+    and then a random string with no punctuation (only letters and numbers).
+    Otherwise, we require at most four elements seperated by period. Each element
+    must only contain numbers. This doesn't play well with our versioning scheme
+    based on git.
+
+    We assume we either have a git tag with three elements or some number of
+    commits beyond a tag, which would take the form of '1.3.2.123.g45e6cd8'. We
+    can't guarentee the sha at the end will begin with a letter. We also can't
```
ferventcoder: minor *nit spelling

lib/vanagon/platform/windows.rb
```diff
@@ -85,7 +85,27 @@ def generate_nuget_package(project)
...
+      #
+      # We can only have at most four elements in the version, all must be numeric
+      # and seperated by periods. We can't handle the version derived from the git
+      # version, which usually contains the short sha of the project. If we have
+      # four or more elements in our version string, we assume we have commits beyond
+      # the latest tag. In that case, we assume we have a version like "1.3.2.123.2g1928uf".
```
ferventcoder: Must be numeric.

ferventcoder:
I agree, the limitations of versioning with NuGet are somewhat short sighted. That's why we are moving towards NVRA - Name version release (package version) architecture as part of the package name. It's kind of a big undertaking though.

melissa:
@ferventcoder in my testing last night, it was complaining when I edited the prerelease version to start with a number. I'll do more testing though. I figured it'd just be easier not to deal with it though... but I'll poke at it some more

lib/vanagon/platform/windows.rb
melissa: See the line below :) I'm just assuming the first four elements are going to be numeric. The first three from the tag and the fourth the number of commits on top of the tag. The fifth element is the short sha, which is not necessarily numeric.

melissa:
@ferventcoder yeah, I can't create a package with `1.3.2.126-2gdkasjdfoiw` as the version but I can with `1.3.2.126-gdkasjdfoiw` as the version

MikaelSmith:
Could hack it to always put in a prefix (like `1.3.2.126-v2gdkasjdfoiw`).

melissa:
What's the most predictable option we have? Other platforms *should* be versioning with both the version and release numbers (i.e., `1.3.2.123.djf84832.1` or `1.3.2.1` depending on if we're building from a tag or there are additional commits to the tag). I feel like  `1.3.2.123` is a little more predictable (or easier to parse out) then `1.3.2.123-vdjf848321`. What do you guys think?

MikaelSmith:
I don't have strong feelings about it.

lib/vanagon/platform/windows.rb
haus: The solaris ips versioning just strips out any non-numerics from the version, so that would be a way of doing this that is more consistent with other platforms.

ferventcoder: @melissa fifth element?

ferventcoder:
> What's the most predictable option we have? Other platforms should be versioning with both the version and release numbers (i.e., 1.3.2.123.djf84832.1 or 1.3.2.1 depending on if we're building from a tag or there are additional commits to the tag). I feel like 1.3.2.123 is a little more predictable (or easier to parse out) then 1.3.2.123-vdjf848321. What do you guys think?

The most predictable depends on whether you are working with prerelease packages or not. If you want the packages to always be considered the stable version, then the only predictable option is version and release number e.g. `1.0.0.t` where `t` equates to `0` if same as tag number and any other number signifying commits since that tag.

ferventcoder:
Perhaps the explanation part that is missing here is what it means to be a prerelease. I feel that is where we are falling down. I don't think the other package managers equate that in the same way.


Currently none of our choco scripts install prerelease packages - it is easily remediable, but I don't think it's something we want to get into.

ferventcoder:
And you are all probably wondering how we deal with this kind of crazy in the .NET world. Somehow, we deal with it. Of course we have an assembly version (four numbered elements), a file version (also four numbered elements) and an informational version (whatever you want here, this is where I usually stuff the git sha)! Talk about crazy right? :D

lib/vanagon/platform/windows.rb
melissa: @ferventcoder `2g1928uf` from my example above

ferventcoder: I wasn't aware you could put a fifth element into the nuget versioning scheme.

ferventcoder: And it doesn't appear to like it (just tried it using choco pack).

melissa:
This most recent update creates packages like `puppet-agent-x86-1.3.2.128-vg46ede48.nupkg`, which must be installed either explicitly stating the version or using the `-pre` flag during install time, since it's considered a prerelease package

lib/vanagon/platform/windows.rb
melissa: It's actually not the worst. It's probably a good thing that we're being confined to semver

templates/windows/project.nuspec.erb
```diff
@@ -2,7 +2,41 @@
...
+    four elements in the version, then we can include the release number.
+
+    Versioning is difficult...
+    -->
+    <%- version_elements = @version.split('.') -%>
+    <%- if version_elements.count >= 4 -%>
+      <%- version = "#{version_elements[0..3].join('.').gsub(/[a-zA-Z]/, '')}-v#{version_elements[4..-1].join('')}" -%>
+    <%- else -%>
+      <%- version = "#{@version}.#{@release}".gsub(/[a-zA-Z]/, '') -%>
+    <%- end -%>
```
underscorgan: Not sure where this template is executed, but do you have access to the methods above? This looks like a lot of copy/pasted code

stahnma:
Man, git describe is close to that format. 
 0.4.1-16-g84babf7

The g there after the hypen (which stands for git) is supposed to help with this. I think you could take a git describe and swap the first - for a . and then you'd be compliant with the versioning, unless I am misreading/misunderstanding this.

stahnma:
Basically, my hope is we can stay as close git describe as we can. I really like using my version control system to do the versioning :octocat:

melissa:
@stahnma are we guaranteed to always have the `g` at the beginning of the short sha?

templates/windows/project.nuspec.erb
melissa: Oh my goodness, you're so right

stahnma:
@melissa with git describe, yes, unless the ref is tag like 1.1.0.

underscorgan:
This looks much better! Want to clean up the commits?

melissa:
@underscorgan still testing, once that's done, I'll clean up the commits

underscorgan:
@melissa ah, ok, cool!

lib/vanagon/platform/windows.rb
```diff
@@ -85,7 +85,37 @@ def generate_nuget_package(project)
...
+      # We can only have at most four elements in the version, all must be numeric
+      # and separated by periods. We can have a prerelease element at the end of the
+      # version string, but it is distinguished by a dash and from what I can tell,
+      # must begin with a letter. We can't handle the version derived from the git
+      # version, which usually contains the short sha of the project since this short
+      # sha often contains a random selection of numbers and letters.
+      #
+      # If we have four or more elements in our version string, we assume we have commits beyond
+      # the latest tag. In that case, we assume we have a version like "1.3.2.123.2g1928uf".
+      # We just take the first four elements and hope they are numeric. Otherwise, we assume
```
stahnma: Hope, or verify?

lib/vanagon/platform/windows.rb
```diff
@@ -85,7 +85,37 @@ def generate_nuget_package(project)
...
+      # we are working from a tag, which would only contain numeric elements. We can then
+      # append the release number to the version string.
+      #
+      # @param project [Vanagon::Project] project to version
+      # @return [String] the version of the nuget package for this project
```
stahnma: Maybe give an example of the expected string here.

lib/vanagon/platform/windows.rb
```diff
@@ -85,7 +85,37 @@ def generate_nuget_package(project)
...
+      def nuget_package_version(version, release)
+        version_elements = version.split('.')
+        # We're dealing with a pre-release> We have to assume we don't have too many elements
```
stahnma: I'm unclear on this comment.

lib/vanagon/platform/windows.rb
```diff
@@ -85,7 +85,37 @@ def generate_nuget_package(project)
...
+      #
+      # If we have four or more elements in our version string, we assume we have commits beyond
+      # the latest tag. In that case, we assume we have a version like "1.3.2.123.2g1928uf".
+      # We just take the first four elements and hope they are numeric. Otherwise, we assume
+      # we are working from a tag, which would only contain numeric elements. We can then
+      # append the release number to the version string.
+      #
+      # @param project [Vanagon::Project] project to version
+      # @return [String] the version of the nuget package for this project
+      def nuget_package_version(version, release)
```
stahnma: Is there a nuget file rather than windows.rb ? (Like we have rpm and deb?) (Random I know) .

melissa:
Ignore the comments for now, focus on the code, does it make sense? I'll update the comments to reflect the new take on dealing with transforming our versioning scheme into a format nuget can handle. The current comments refer to my previous attempt as solving this. I'm still testing this new format, so I want to make sure the new attempt actually works before restructuring the comments so that I only have to do that once.

melissa:
Okay, I updated the comments and tested and everything.

ping @ferventcoder @MikaelSmith @stahnma @underscorgan

stahnma:
:+1:

lib/vanagon/platform/windows.rb
melissa: nope. I couldn't get that to work. Pulling in some people who are better at ruby then I am only lead to the decision that I was trying to do things that were too complicated.

ferventcoder:
Big :+1:

PR #236, opened 2015-12-17
(maint) Pass ':' if there is nothing to extract

melissa:
Because of excitement on windows, we need to set certain environment
variables when we unpack tarballs that contain symlinks. That means
we've modified the Makefile to include the environment settings as well
as '&& \'. When this is followed by an empty line, make fails. So, in
order to prevent this failure, we need to pass in ':' so that nothing
happens when there is no extract command to pass in.


mckern:
excited swanson

McdonaldSeanp:
:+1:

PR #238, opened 2015-12-19
(maint) Skip deleting missing files

MikaelSmith:
We should still succeed at uninstall if files are missing.


templates/windows/chocolateyUninstall.ps1.erb
```diff
@@ -9,6 +9,8 @@ if (Test-Path -path "$contentFile") {
     $fileInstalled=$fileInstalled.Replace("/", "\")
     if (Test-Path -path "$fileInstalled" -PathType Container) {
       Write-Debug "Not removing directory $fileInstalled"
+    } else if (Test-Path -path "$fileInstalled") {
```
MikaelSmith: Blah, `else if` isn't right.

jcoconnor:
:+1:

PR #239, opened 2015-12-21
(CPR-263) Teach vanagon how to add post-installation gem patches

melissa:
We need to be able to add patches to some of the gems we're providing in
our windows build. This commits adds in a new gem-patch task to the
Makefile if we have gems. It also exposes the extension type for the
source that we're currently working with. In the special case we're
dealing with no source or a git repo, we have to manually specify the
extension type. There's probably a better way to do this....

### DIFF
--- a/lib/vanagon/component/dsl.rb
+++ b/lib/vanagon/component/dsl.rb
@@ -80,10 +80,13 @@ class Vanagon
       # Add a patch to the list of patches to apply to the component's source after unpacking
       #
       # @param patch [String] Path to the patch that should be applied
+      # @param destination [String] Path to the location where the patch should be applied
       # @param strip [String, Integer] directory levels to skip in applying patch
       # @param fuzz [String, Integer] levels of context miss to ignore in applying patch
-      def apply_patch(patch, strip: 1, fuzz: 0)
-        @component.patches << OpenStruct.new('path' => patch, 'strip' => strip.to_s, 'fuzz' => fuzz.to_s)
+      # @param after [String] the location in the makefile where the patch command should be run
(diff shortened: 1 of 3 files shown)


haus:
I think simply adding a patch_gem dsl method to the component would be more straightforward than autodetecting.

haus:
@melissa workdir is already set on line 5

melissa:
@haus cool, thanks! Removed

lib/vanagon/component/dsl.rb
```diff
@@ -80,10 +80,12 @@ def environment(&block)
       # Add a patch to the list of patches to apply to the component's source after unpacking
       #
       # @param patch [String] Path to the patch that should be applied
+      # @param destination [String] Path to the location where the patch should be applied
       # @param strip [String, Integer] directory levels to skip in applying patch
       # @param fuzz [String, Integer] levels of context miss to ignore in applying patch
-      def apply_patch(patch, strip: 1, fuzz: 0)
-        @component.patches << OpenStruct.new('path' => patch, 'strip' => strip.to_s, 'fuzz' => fuzz.to_s)
+      # @param after [String] the location in the makefile where the patch command should be run
+      def apply_patch(patch, destination = @component.dirname, strip: 1, fuzz: 0, after: 'unpack')
```
underscorgan: Is there a reason you aren't using a keyword argument for destination?

melissa: No, not really. Do you think it'd be cleaner if I did?

underscorgan: I do :)

ScottGarman:
:+1:

underscorgan:
can this have updated tests, since we do  have those for vanagon?

PR #241, opened 2015-12-23
(CPR-263) Allow patch to apply patches after installation

melissa:


melissa:
cc/ @underscorgan @mckern @adrienthebo This is a change that was merged into the vanagon_for_the_win branch. I know @adrienthebo has done a lot of work to refactor the makefiles here, so I wanted to see what we should do to make this work and that amazing work merge together well. Anyway, here's a cherry pick of the commit to the feature branch. I haven't looked into what it would take to integrate this with what Adrien has already done. But that should probably happen at some point.

adrienthebo:
I think it would be easiest to merge this in and then I can rebase on top of it; your change is more limited in scope and should be easier to pull in.

shrug:
:+1:

PR #243, opened 2015-12-28
(CPR-266) Allow vanagon to work with zip files

stahnma:
Prior to this patch, vanagon could not work with a zip file as a source
to fetch/extract and build a package from. This simply adds zip to the
ARCHIVE_EXTENSIONS constant and returns the correct command to use for
unpacking the archive.


shermdog:
:+1:

PR #244, opened 2015-12-30
(maint) Update windows defaults

melissa:


melissa:
In pl-build-tools-vanagon and puppet-agent, we're already overriding the make definition (https://github.com/puppetlabs/pl-build-tools-vanagon/blob/master/configs/platforms/win-2012-x64.rb#L15), so this should have no real impact on our current build process.

shrug:
:+1:

PR #247, opened 2015-12-31
(maint) Allow Vanagon to override `cp`

melissa:
This PR also sets `@install` for windows and fixes a typo


stahnma:
Why is this needed if the default is still /usr/bin/cp ?  (I'd prefer why to be in the commit messages)

melissa:
It's cygwin cp versus windows cp. We want to ensure it's using cygwin cp. Just like with find and sort. I need to fix the tests though.

melissa:
@stahnma okay, I fixed tests and updated the commit messages. We're defaulting to `cp` for all platforms except windows, where we need to use the cygwin `/usr/bin/cp`

stahnma:
Im still confused, isn't cp normally /usr/bin/cp?

ScottGarman:
:+1:

PR #248, opened 2016-01-04
(maint) Fix case typo

melissa:
This PR updates a typo made when defining what extension we are dealing
with. Rather then `||`, ruby case requires `,` to define multiple
possibilities for satisfying when we enter this block.


shermdog:
:+1:

PR #249, opened 2016-01-05
(PA-121) adding support for .vim syntax files

McdonaldSeanp:
In order to facilitate config files for vim, this commit
adds the .vim extension to the list of NON_ARCHIVE_EXTENSIONS
in the component/source/http.rb file


ScottGarman:
:+1:

PR #250, opened 2016-01-07
WIP (CPR-273) Allow repeatable builds

MikaelSmith:
Vanagon builds should be repeatable over the same temporary directory
(via make). Currently attempting to repeat a link fails the build
because the symlink already exists. Use `-f` to force recreation of the
symlink.

### DIFF
--- a/lib/vanagon/component/dsl.rb
+++ b/lib/vanagon/component/dsl.rb
@@ -241,7 +241,10 @@ class Vanagon
       # @param target [String] path to the desired symlink
       def link(source, target)
         @component.install << "#{@component.platform.install} -d '#{File.dirname(target)}'"
-        @component.install << "ln -s '#{source}' '#{target}'"
+        # Use a bash conditional to only create the link if it doesn't already point to the correct source.
+        # This allows rerunning the install step to be idempotent, rather than failing because the link
+        # already exists.
+        @component.install << "([[ `readlink '#{target}'` == '#{source}' ]] || ln -s '#{source}' '#{target}')"
         @component.add_file Vanagon::Common::Pathname.file(target)
       end
 
--- a/lib/vanagon/component.rb
+++ b/lib/vanagon/component.rb
@@ -9,7 +9,7 @@ class Vanagon
     attr_accessor :name, :version, :source, :url, :configure, :build, :check, :install
     attr_accessor :environment, :extract_with, :dirname, :build_requires, :build_dir
     attr_accessor :settings, :platform, :patches, :requires, :service, :options
-    attr_accessor :directories, :replaces, :provides, :cleanup_source, :environment
+    attr_accessor :directories, :replaces, :provides, :cleanup_source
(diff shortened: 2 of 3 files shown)


MikaelSmith:
@melissa does this make sense? Link steps can be repeated if an earlier dependency is retriggered.

stahnma:
I'm curious if we should have two methods, one for link and one for link! or something. The -f will override if the link was a directory or file prior to existing. We could also check to see if it's a link already and then -sf it.

haus:
-f also has the downside of not exploding if it would normally fail

MikaelSmith:
How about adding a check that the link already exists and points to the expected target? If it does, do nothing, otherwise do `ln -s`.

haus:
that seems reasonable

lib/vanagon/component/dsl.rb
```diff
@@ -241,7 +241,10 @@ def install_configfile(source, target)
       # @param target [String] path to the desired symlink
       def link(source, target)
         @component.install << "#{@component.platform.install} -d '#{File.dirname(target)}'"
-        @component.install << "ln -s '#{source}' '#{target}'"
+        # Use a bash conditional to only create the link if it doesn't already point to the correct source.
+        # This allows rerunning the install step to be idempotent, rather than failing because the link
+        # already exists.
+        @component.install << "([[ `readlink '#{target}'` == '#{source}' ]] || ln -s '#{source}' '#{target}')"
```
MikaelSmith: I'm not sure whether the bashism is a safe assumption, but I couldn't think of another way to do it within the context of 'a && b && c' we use to concatenate install steps.

stahnma: readlink isn't readily available on AIX. Not sure about other platforms. Would something like `ls -l #{target} | awk '{print $NF}'` work ?

MikaelSmith: Darn, thought I was safe checking Cygwin. `ls | awk` should work.

mckern: Suggestion: 

```bash
mckern@flexo Downloads $ ln -s ./IMG_0910.JPG somelink.jpg
mckern@flexo Downloads $ [[ somelink.jpg -ef somelink.jpg ]] && echo "true"
true
mckern@flexo Downloads $ [[ somelink.jpg -ef IMG_0910.JPG ]] && echo "true"
true
mckern@flexo Downloads $ [[ somelink.jpg -ef IMG_0911.JPG ]] && echo "true" || echo "false"
false
```

No subshells, no pipes, still a Bashism but portable to any shell if `[` is installed or it supports the extended test operator (`[[]]`).

MikaelSmith: Thanks, I can try that.

MikaelSmith:
Retargeting against master.

melissa:
Closed in favor of https://github.com/puppetlabs/vanagon/pull/268

PR #251, opened 2016-01-08
(CPR-269) Assume unknown extension types are just files

stahnma:
Prior to this patch, we had to explicitly add every extension we wanted
to use a source with vanagon. This resulted in lots of small patches
with a simple addition of .json or .txt or something. It also meant
that to use those extensions, a release of vanagon was needed. This
slows down work.

Instead, we'll default to any extension that is not explicitly called
out as an archive, is simply a flat file and able to be used that way.
The net result here is that only new archive extensions will require
patches.

### DIFF
--- a/lib/vanagon/component/source/http.rb
+++ b/lib/vanagon/component/source/http.rb
@@ -12,9 +12,6 @@ class Vanagon
         # Extensions for files we intend to unpack during the build
         ARCHIVE_EXTENSIONS = '.tar.gz', '.tgz', '.zip'
 
-        # Extensions for files we aren't going to unpack during the build
-        NON_ARCHIVE_EXTENSIONS = '.gem', '.ru', '.txt', '.conf', '.ini', '.gpg', '.rb', '.sh', '.csh', '.xml', '.vim', '.json', '.service'
-
         # Constructor for the Http source type
         #
         # @param url [String] url of the http source to fetch
@@ -106,11 +103,9 @@ class Vanagon
             when ".zip"
               return "unzip '#{@file}'"
             end
-          elsif NON_ARCHIVE_EXTENSIONS.include?(@extension)
-            # Don't need to unpack gems, ru, txt, conf, ini, gpg
-            return nil
           else
-            fail "Extraction unimplemented for '#{@extension}' in source '#{@file}'. Please teach me."
+            # Extension does not appear to be an archive
+            return nil
           end
         end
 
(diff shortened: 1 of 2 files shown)


stahnma:
I built two vanagon projects using this patch set and tests are updated.

shermdog:
:+1:

lib/vanagon/component/source/http.rb
```diff
@@ -121,24 +116,20 @@ def extract(tar)
...
         # Returns the extension for @file
         #
         # @return [String] the extension of @file
-        # @raise [RuntimeError] an exception is raised if the extension isn't in the current list
         def get_extension
-          extension_match = @file.match(/.*(#{Regexp.union(ARCHIVE_EXTENSIONS + NON_ARCHIVE_EXTENSIONS)})/)
+          extension_match = @file.match(/.*(#{Regexp.union(ARCHIVE_EXTENSIONS)})/)
           unless extension_match
-            fail "Unrecognized extension for '#{@file}'. Don't know how to extract this format. Please teach me."
+            return @file.split('.').last
```
haus: Are there any plaintext files with more than one `.` in the extension?

Also, I think this will return `txt` instead of `.txt`, which used to be returned.

spec/lib/vanagon/component/source/http_spec.rb
```diff
@@ -30,7 +30,7 @@
 
   describe "#get_extension" do
     it "returns the extension for valid extensions" do
-      (Vanagon::Component::Source::Http::ARCHIVE_EXTENSIONS + Vanagon::Component::Source::Http::NON_ARCHIVE_EXTENSIONS).each do |ext|
+      Vanagon::Component::Source::Http::ARCHIVE_EXTENSIONS.each do |ext|
```
haus: In light of the above comment about `txt` vs `.txt` it would be worth adding a test for non-archive extensions.

lib/vanagon/component/source/http.rb
stahnma: Does it matter? Basically all the checks in the code are checking if the extension is in ARCHIVE_EXTENSIONS, I think. Maybe I am missing something. 

I assume you could have something like .repo.txt files or something.

spec/lib/vanagon/component/source/http_spec.rb
stahnma: Ok, I'll look at that.

lib/vanagon/component/source/http.rb
haus: I think being consistent with the return for archive extensions is valuable for least surprise.

I'm not especially concerned with the double . extension, mostly curious.

lib/vanagon/component/source/http.rb
```diff
@@ -121,24 +116,25 @@ def extract(tar)
...
         #
         # @return [String] the extension of @file
-        # @raise [RuntimeError] an exception is raised if the extension isn't in the current list
         def get_extension
-          extension_match = @file.match(/.*(#{Regexp.union(ARCHIVE_EXTENSIONS + NON_ARCHIVE_EXTENSIONS)})/)
+          extension_match = @file.match(/.*(#{Regexp.union(ARCHIVE_EXTENSIONS)})/)
           unless extension_match
-            fail "Unrecognized extension for '#{@file}'. Don't know how to extract this format. Please teach me."
+            if @file.split('.').last.include?('.')
+              return '.' +  @file.split('.').last
```
mckern: Nitpick: we can do this with split or positions:

```ruby
>> name = '/path/to/file.txt'
=> "/path/to/file.txt"
>> file = Pathname.new(name)
=> #<Pathname:/path/to/file.txt>
>> file.extname
=> ".txt"
>> file = File.extname(name)
=> ".txt"
```

underscorgan:
:+1:

McdonaldSeanp:
:+1: :+1: :+1: :+1:

PR #252, opened 2016-01-08
(feature) Merge windows feature branch into master

melissa:


melissa:
@haus @shrug @MikaelSmith @underscorgan @jcoconnor @stahnma 

@MikaelSmith made the point we should probably get the work in the feature branch merged down into our production branch sooner rather then later. This will help ensure these changes are causing as little churn as possible, and will help get people testing the work (even though we will not officially declare windows support until it's fully done). What are your thoughts on this?

MikaelSmith:
Not just testing the Windows work, but get some validation that other functionality isn't broken.

lib/vanagon/platform/dsl.rb
```diff
@@ -43,6 +44,8 @@ def platform(name, &block)
                       Vanagon::Platform::Solaris10.new(@name)
                     when /^solaris-11/
                       Vanagon::Platform::Solaris11.new(@name)
+                    when /^win-/
```
haus: I think using the full string windows in the platform name will be clearer/cleaner.

lib/vanagon/platform/dsl.rb
```diff
@@ -97,6 +100,27 @@ def package_type(pkg_type)
         @platform.package_type = pkg_type
       end
 
+      # Set the path to find for the platform
+      #
+      # @param find_cmd [String] Full path to the find command for the platform
```
haus: Does vanagon care if the find is gnu find or bsd find?

lib/vanagon/platform/dsl.rb
```diff
@@ -97,6 +100,27 @@ def package_type(pkg_type)
...
+      def find(find_cmd)
+        @platform.find = find_cmd
+      end
+
+      # Set the path to sort for the platform
+      #
+      # @param sort_cmd [String] Full path to the sort command for the platform
```
haus: Same as for find. Does it matter if gnu or bsd?

lib/vanagon/platform/windows.rb
```diff
@@ -0,0 +1,203 @@
...
+      # @param project [Vanagon::Project] project to build a windows package of
+      # @return [Array] list of commands required to build a windows package for the given project from a tarball
+      def generate_package(project)
+        # If nothing is passed in as platform type, default to building a nuget package
+        # We should default to building an MSI once that code has been implimented
+        if project.platform.package_type.nil? || project.platform.package_type.empty?
+          return generate_nuget_package(project)
+        else
+          case project.platform.package_type
+          when "nuget", "nupkg"
```
haus: Are nuget and nupkg different in any way?

lib/vanagon/platform/windows.rb
```diff
@@ -0,0 +1,203 @@
...
+      #
+      # @param project [Vanagon::Project] project to build a windows package of
+      # @return [Array] list of commands required to build a windows package for the given project from a tarball
+      def generate_package(project)
+        # If nothing is passed in as platform type, default to building a nuget package
+        # We should default to building an MSI once that code has been implimented
+        if project.platform.package_type.nil? || project.platform.package_type.empty?
+          return generate_nuget_package(project)
+        else
+          case project.platform.package_type
```
haus: As this is a case of one it probably makes more sense to be rolled into the outer if as an elsif.

lib/vanagon/platform/windows.rb
```diff
@@ -0,0 +1,203 @@
...
+      #
+      # @param project [Vanagon::Project] project to name
+      # @return [String] name of the windows package for this project
+      def package_name(project)
+        # If nothing is passed in as platform type, default to a nuget package
+        # We should default to an MSI once that code has been implimented
+        if project.platform.package_type.nil? || project.platform.package_type.empty?
+          return nuget_package_name(project)
+        else
+          case project.platform.package_type
```
haus: Same as above with respect to case moving to an elsif in outer conditional.

lib/vanagon/platform/windows.rb
```diff
@@ -0,0 +1,203 @@
...
+      #
+      # Examples of prerelease versions:
+      #   1.2.3.1234-g124dm9302
+      #   3.2.5.23-gd329nd
+      #
+      # @param project [Vanagon::Project] project to version
+      # @return [String] the version of the nuget package for this project
+      def nuget_package_version(version, release)
+        version_elements = version.split('.')
+        if version_elements[-1].start_with?('g')
```
haus: `.last` would be more clear than `[-1]`.

lib/vanagon/platform/dsl.rb
melissa: It cares if you're using cygwin find versus MS-DOS find

lib/vanagon/platform/dsl.rb
melissa: It cares if you're using cygwin sort versus MS-DOS sort

lib/vanagon/platform/windows.rb
melissa: Nope. I figured I'd allow for some wiggle room with how people wanted to specify this. Nuget is the name of the package format, nupkg is the file extension designating the nuget package.

haus: It's probably better to just pick one and go with it. I would go with nuget.

templates/Makefile.erb
```diff
@@ -37,13 +37,13 @@ file-list: file-list-before-build <%= @name %>-project
 
 <%- if @bill_of_materials -%>
 <%= @bill_of_materials %>:
-	mkdir -p '<%= @bill_of_materials.path %>'
-	mv bill-of-materials '<%= @bill_of_materials.path %>'
+	mkdir -p "<%= @bill_of_materials.path %>"
```
haus: Why the change from single to double quotes? That seems less safe.

lib/vanagon/platform/dsl.rb
melissa: We're using the short name in the platform definitions: `win-2012-x64.rb`. Although we could change this.

lib/vanagon/platform/windows.rb
melissa: Probably true

lib/vanagon/platform/dsl.rb
haus: right, i get that. windows would be more consistent with debian, ubuntu, solaris, sles, ...

templates/Makefile.erb
melissa: I think windows handles double quotes better then single quotes while dealing with spaces. This also could have been me desperately trying to get something to work and is unnecessary, but I'm not sure

melissa: Should I change it back to single quotes? I vaguely remember something about double quotes being preferable for dealing with spaces... cc/ @MikaelSmith was it you who mentioned that?

MikaelSmith: I don't think it matters for spaces. Double quotes are for interpolating variables.

lib/vanagon/platform/dsl.rb
melissa: Addressed in https://github.com/puppetlabs/vanagon/pull/258

lib/vanagon/platform/windows.rb
melissa: Addressed in https://github.com/puppetlabs/vanagon/pull/258

lib/vanagon/platform/windows.rb
melissa: Addressed in https://github.com/puppetlabs/vanagon/pull/258

lib/vanagon/platform/windows.rb
melissa: Addressed in https://github.com/puppetlabs/vanagon/pull/258

templates/Makefile.erb
melissa: Alright, addressed in https://github.com/puppetlabs/vanagon/pull/258

melissa:
I've tested building el-7-x86_64 and solaris-10-i386 with this plus the changes in https://github.com/puppetlabs/vanagon/pull/258 and got successful builds

melissa:
@haus a lot of your questions have been addressed and are now in this PR

haus:
nothing crazy here

MikaelSmith:
:+1:

lib/vanagon/component/source/http.rb
```diff
@@ -105,7 +105,7 @@ def extract(tar)
             end
           else
             # Extension does not appear to be an archive
-            return nil
+            return ':'
```
mckern: I'm seeing a lot of the `:` operator being spread around; are we worried that this will paint us into an explicitly bourne-shell shaped corner?

lib/vanagon/engine/base.rb
```diff
@@ -57,7 +57,7 @@ def ship_workdir(workdir)
 
       def retrieve_built_artifact
         FileUtils.mkdir_p("output")
-        Vanagon::Utilities.rsync_from("#{@remote_workdir}/output/*", "#{@target_user}@#{@target}", "output", @platform.ssh_port)
+        Vanagon::Utilities.rsync_from("#{@remote_workdir}/output/*", "#{@target_user}@#{@target}", "output/", @platform.ssh_port)
```
mckern: Trailing slash doesn't/shouldn't matter on the rsync receiver side, only on the sender side.

lib/vanagon/component/source/http.rb
melissa: Should I change this to `/bin/true`?

lib/vanagon/platform.rb
```diff
@@ -101,6 +101,10 @@ def initialize(name)
       @ssh_port = 22
       @provisioning = []
       @install ||= "install"
+      @target_user ||= "root"
+      @find ||= "find"
+      @sort ||= "sort"
+      @copy ||= "cp"
```
mckern: Out of scope for a merge or even a casual PR but as the number of built-in tools we rely on grows, we're probably going to have subclass platform utilities.

lib/vanagon/component/source/http.rb
mckern: Nah, `:` is fine. I just wonder if we're using that functionality too much like Debian init scripts so and papering over weak abstractions.

lib/vanagon/engine/base.rb
melissa: I can't remember... I make this change a while ago. I assume I did run into an issue moving files to/from the windows box though https://github.com/puppetlabs/vanagon/commit/b36aefa54a3f9c84681ce72b3bdd61a144aa0041 is the commit

lib/vanagon/platform.rb
```diff
@@ -226,7 +230,7 @@ def is_unix?
     #
     # @return [true, false] true if it is a windows variety, false otherwise
     def is_windows?
-      return !!@name.match(/^win-.*$/)
+      return !!@name.match(/^windows-.*$/)
```
mckern: Rubocop is weird about this. It suggests using `!name.match(/^windows-.*$/).nil?` instead. I have mixed feelings, as the double-negation is common enough in our codebase. Does anyone have thoughts about this?

(I know that this isn't the PR to fix this stuff in, but we might as well get a list together).

lib/vanagon/platform/windows.rb
```diff
@@ -0,0 +1,180 @@
+class Vanagon
+  class Platform
+    class Windows < Vanagon::Platform
+      # The specific bits used to generate a windows package for a given project
+      #
+      # @param project [Vanagon::Project] project to build a windows package of
+      # @return [Array] list of commands required to build a windows package for the given project from a tarball
+      def generate_package(project)
+        # If nothing is passed in as platform type, default to building a nuget package
+        # We should default to building an MSI once that code has been implimented
```
underscorgan: [nitpick] implemented spelled incorrectly.

lib/vanagon/platform/dsl.rb
```diff
@@ -97,6 +100,27 @@ def package_type(pkg_type)
         @platform.package_type = pkg_type
       end
 
+      # Set the path to find for the platform
+      #
+      # @param find_cmd [String] Full path to the find command for the platform
+      def find(find_cmd)
```
mckern: Shouldn't this be

```ruby
def find=(find_cmd)
```
?

lib/vanagon/platform/dsl.rb
```diff
@@ -97,6 +100,27 @@ def package_type(pkg_type)
...
+        @platform.find = find_cmd
+      end
+
+      # Set the path to sort for the platform
+      #
+      # @param sort_cmd [String] Full path to the sort command for the platform
+      def sort(sort_cmd)
```
mckern: See comment, re: `find=`

lib/vanagon/platform/dsl.rb
```diff
@@ -97,6 +100,27 @@ def package_type(pkg_type)
...
+        @platform.sort = sort_cmd
+      end
+
+      # Set the path to copy for the platform
+      #
+      # @param copy_cmd [String] Full path to the copy command for the platform
+      def copy(copy_cmd)
```
mckern: See comment, re: `find=`

lib/vanagon/platform/dsl.rb
melissa: No.... only if you want to set `plat.find= /usr/bin/find` in the platform config. This goes with the current practice of setting variables, so the user specified `plat.find "/usr/bin/find"`

lib/vanagon/platform/windows.rb
melissa: Addressed in https://github.com/puppetlabs/vanagon/pull/259

lib/vanagon/platform/dsl.rb
mckern: :disappointed:

lib/vanagon/platform/windows.rb
```diff
@@ -0,0 +1,180 @@
...
+          erb_file(File.join(VANAGON_ROOT, "templates/windows/#{win_file}.erb"), File.join(workdir, win_file), false, { :binding => binding })
+        end
+      end
+
+      # The specific bits used to generate a windows nuget package for a given project
+      # Nexus expects packages to be named #{name}-#{version}.nupkg. However, chocolatey
+      # will generate them to be #{name}.#{version}.nupkg. So, we have to rename the
+      # package after we build it. We also take advantage of the rename to sneak the
+      # architecture in the package name. Neither chocolatey nor nexus know how to deal
+      # with architecture, so we are just pretending it's part of the package name.
```
mckern: yey

lib/vanagon/platform/windows.rb
```diff
@@ -0,0 +1,180 @@
...
+        "(cd $(tempdir)/#{project.name} ; C:/ProgramData/chocolatey/bin/choco.exe pack #{project.name}.nuspec)",
+        "cp $(tempdir)/#{project.name}/#{project.name}-#{@architecture}.#{nuget_package_version(project.version, project.release)}.nupkg ./output/#{target_dir}/#{nuget_package_name(project)}"]
+      end
+
+      # Method to derive the package name for the project.
+      # Nexus expects packages to be named #{name}-#{version}.nupkg. However, chocolatey
+      # will generate them to be #{name}.#{version}.nupkg. So, we have to rename the
+      # package after we build it. We also take advantage of the rename to sneak the
+      # architecture in the package name. Neither chocolatey nor nexus know how to deal
+      # with architecture, so we are just pretending it's part of the package name.
```
mckern: This sort of duplicates the comment from #49-54; We should probably consolidate that?

melissa: Addressed in https://github.com/puppetlabs/vanagon/pull/260

lib/vanagon/platform.rb
stahnma: Double bang was pretty common in ruby books from circa 2007. I found that to be more idiomatic ruby than `!thing.nil?`. Also, in lock_manager, we totes disabled that check.

lib/vanagon/component/source/http.rb
stahnma: I'd probably vote for `/bin/true` as it's easier to read.

lib/vanagon/platform.rb
mckern: Like I said, mixed feelings. But 2007 was 9 years ago; I probably wouldn't cite anything Ruby related from 9 years ago. Note that I do not totally disagree with you, though after thinking about it I find the intent in `!thing.nil?` a little easier to parse at a glance.

lib/vanagon/component/source/http.rb
haus: True isn't always in /bin

mckern: Right. I'd prefer `:` because

```shell
mckern@flexo ~ $ type :
: is a shell builtin
```

stahnma: true is builtin as well. I guess the /bin is the problem, that tells it not to use the built in. 

    stahnma@stahnma-pro ~> type true
   true is a shell builtin

mckern: Flip a coin, I guess?

```
mckern@flexo ~ $ help :
:: :
    No effect; the command does nothing.  A zero exit code is returned.
mckern@flexo ~ $ help true
true: true
    Return a successful result.
```

lib/vanagon/platform.rb
melissa: I'm not changing this here. That should be something we change in the main branch and not a feature branch.

lib/vanagon/platform.rb
melissa: Agreed out of scope for this PR

lib/vanagon/component/source/http.rb
melissa: soooo, it looks like we're divided... I'm just going to leave this as is until someone tells me definitely to change it.

mckern: Works for me. We'll hammer it out somewhere else and fix it in `MASTER` when we decide.

templates/windows/chocolateyInstall.ps1.erb
```diff
@@ -0,0 +1,28 @@
...
+    }
+    Copy-Item -path "$origin" -destination "$destination" -Force
+  } else {
+    # If the item we are trying to copy over does not exist in our source directory,
+    # we assume it is an empty directory and simply create one in its place. There is
+    # a possibility that this will hide an error where there is actually a missing
+    # file. However, this is such a slim possibity, this action was deemed safe.
+    New-Item -ItemType directory -Path "$destination"
+  }
+}
```
underscorgan: Why is this an erb? IT doesn't look like it's using any ruby elements?

templates/windows/chocolateyUninstall.ps1.erb
```diff
@@ -0,0 +1,24 @@
...
+    # directory structure. It doesn't seem worth it ATM. As is, only remove
+    # things that are files, and do not remove any directories.
+    if (!((Get-Item "$file") -is [System.IO.DirectoryInfo])) {
+      Write-Debug "Removing '$file'"
+      remove-item -Path "$file" -Force
+    }
+  } else {
+    Write-Debug "Skipping missing file '$file'"
+  }
+}
```
underscorgan: Why is this an erb? It doesn't look like it's using any ruby elements?

templates/windows/chocolateyInstall.ps1.erb
melissa: It was originally, but then I changed it and forgot to make it not an erb...  good point. I should probably change that

templates/windows/chocolateyUninstall.ps1.erb
melissa: https://github.com/puppetlabs/vanagon/pull/264

templates/windows/chocolateyInstall.ps1.erb
melissa: https://github.com/puppetlabs/vanagon/pull/264

Iristyle:
Totally agree with @MikaelSmith comments from 2 weeks ago re: merging down.

We're now at 70 commits in this branch, and there are currently merge conflicts.  Is there any way we can start to move parts of this into the master branch?  I'd like to see us integrate these bits more frequently, rather than continuing a parallel long-lived branch.

melissa:
@Iristyle want to merge https://github.com/puppetlabs/vanagon/pull/265 so I can rebase this and resolve the merge conflicts and then merge this?

melissa:
https://github.com/puppetlabs/vanagon/pull/266 will fix the breakages the last update from master caused

templates/windows/chocolateyUninstall.ps1.erb
melissa: https://github.com/puppetlabs/vanagon/pull/266

Iristyle:
You beat me @underscorgan ... so @melissa should we cut a new `0.5.x` gem soon?

PR #253, opened 2016-01-08
(maint) Switch from expand_path to pathname to clean up paths

melissa:
expand_path can't handle windows paths, so we need to not use it.

### DIFF
--- a/lib/vanagon/common/pathname.rb
+++ b/lib/vanagon/common/pathname.rb
@@ -1,3 +1,5 @@
+require 'pathname'
+
 class Vanagon
   class Common
     class Pathname
@@ -28,7 +30,7 @@ class Vanagon
       #   and exposed through the {#configfile?} method.
       # @return [Vanagon::Common::Pathname] Returns a new Pathname instance.
       def initialize(path, mode: nil, owner: nil, group: nil, config: false)
-        @path = File.expand_path(path)
+        @path = ::Pathname.new(path).realpath.to_s
         @mode ||= mode
         @owner ||= owner
         @group ||= group


lib/vanagon/common/pathname.rb
```diff
@@ -28,7 +30,7 @@ class Pathname
       #   and exposed through the {#configfile?} method.
       # @return [Vanagon::Common::Pathname] Returns a new Pathname instance.
       def initialize(path, mode: nil, owner: nil, group: nil, config: false)
-        @path = File.expand_path(path)
+        @path = ::Pathname.new(path).cleanpath
```
mckern: Did you determine if this needs to be coerced to a string or if the use of the bare `::Pathname` object was OK?

melissa: It does need to be normalized

mckern: I think we'll have to revisit this some day or write good doc on it, just because of the joys of interpolated strings vs. literal ones:

```ruby
>> path = Pathname.new('C:\some thing\some where')
=> #<Pathname:C:\some thing\some where>
>> path = Pathname.new("C:\some thing\some where")
=> #<Pathname:C: ome thing ome where>
```

mckern:
I think we'll have to revisit this some day or write good doc on it, just because of the joys of interpolated strings vs. literal ones:

```ruby
>> path = Pathname.new('C:\some thing\some where')
=> #<Pathname:C:\some thing\some where>
>> path = Pathname.new("C:\some thing\some where")
=> #<Pathname:C: ome thing ome where>
```

mckern:
Oh glob, Ruby is such a pain in the ****s.

```ruby
>> path = Pathname.new('C:\some thing\some where')
=> #<Pathname:C:\some thing\ some where>
>> path.cleanpath
=> #<Pathname:C:\some thing\ some where>
>> path.to_s
=> "C:\\some thing\\ some where"
>> path.absolute?
=> false
>> path.relative?
=> true
```

lib/vanagon/common/pathname.rb
```diff
@@ -28,7 +30,7 @@ class Pathname
       #   and exposed through the {#configfile?} method.
       # @return [Vanagon::Common::Pathname] Returns a new Pathname instance.
       def initialize(path, mode: nil, owner: nil, group: nil, config: false)
-        @path = File.expand_path(path)
+        @path = ::Pathname.new(path).cleanpath.to_s
```
mckern: It looks like we should probably use `#realpath` instead of `#cleanpath` as `#cleanpath` runs the risk of containing `..`.

http://ruby-doc.org/stdlib-2.1.0/libdoc/pathname/rdoc/Pathname.html#method-i-cleanpath
http://ruby-doc.org/stdlib-2.1.0/libdoc/pathname/rdoc/Pathname.html#method-i-realpath

Yes, I know I'm the one who said "use `#cleanpath`" and I am sorry.

mckern:
Yuck. Apparently the convention for paths in Ruby is "forward slash regardless of platform". So you get very different results for `#absolute?` and `#relative?` for `C:\path\to\places` vs. `C:/path/to/places`.

programming

melissa:
Closed in favor of https://github.com/puppetlabs/vanagon/pull/254

melissa:
/code irb(main):016:0> puts Pathname.new(win_path).realpath
Errno::ENOENT: No such file or directory @ realpath_rec - /home/melissa/vanagon/C:
	from (irb):16:in `realpath'
	from (irb):16:in `realpath'
	from (irb):16
	from /home/melissa/.rbenv/versions/2.1.0/bin/irb:11:in `<main>'

melissa:
@mckern pathname.realpath gives us the same problems as expand_path

melissa:
I think we still want cleanpath for this
```
irb(main):008:0> puts Pathname.new("win//nsde/path/../stis/gsdf").cleanpath
win/nsde/stis/gsdf
=> nil
irb(main):009:0> puts Pathname.new("/win//nsde/path/../stis/gsdf").cleanpath
/win/nsde/stis/gsdf
=> nil
irb(main):010:0> puts Pathname.new("/win//nsde/path/../stis/gsdf").realpath
Errno::ENOENT: No such file or directory @ realpath_rec - /win
	from (irb):10:in `realpath'
	from (irb):10:in `realpath'
	from (irb):10
	from /home/melissa/.rbenv/versions/2.1.0/bin/irb:11:in `<main>'
```

mckern:
@melissa if the Path won't exist, then yes, `#cleanpath` is the thing to use.

PR #255, opened 2016-01-11
(maint) Normalize path shortcuts

melissa:
On windows, we need the ability to deal with spaces in path names. This
commit does two things to handle that. First, it adds double quotes
around all the path names when we query the filesystem. Then, it adds
cygpath to normalize the path. This lets us change Progra~1 to Program
Files. Super gross, I know. I didn't know what else to do. There are
enough assumptions on cygwin currently, I figured it was safe to add one
more.

### DIFF
--- a/lib/vanagon/project.rb
+++ b/lib/vanagon/project.rb
@@ -232,6 +232,11 @@ class Vanagon
       files = ['file-list', 'bill-of-materials']
       files.push get_files.map(&:path)
       files.push get_configfiles.map(&:path)
+      if @platform.is_windows?
+        files.flatten.map { |f| "$$(cygpath --mixed --long-name '#{f}')" }
+      else
+        files.flatten
+      end
     end
 
     # Generate a bill-of-materials: a listing of the components and their
--- a/lib/vanagon/project.rb
+++ b/lib/vanagon/project.rb
@@ -248,7 +248,7 @@ class Vanagon
     def pack_tarball_command
       tar_root = "#{@name}-#{@version}"
       ["mkdir -p '#{tar_root}'",
-       %('#{@platform.tar}' -cf - -T #{get_tarball_files.join(" ")} | ( cd '#{tar_root}/'; '#{@platform.tar}' xfp -)),
+       %('#{@platform.tar}' -cf - -T "#{get_tarball_files.join('" "')}" | ( cd '#{tar_root}/'; '#{@platform.tar}' xfp -)),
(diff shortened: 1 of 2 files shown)


McdonaldSeanp:
:+1:

lib/vanagon/project.rb
```diff
@@ -232,6 +232,11 @@ def get_tarball_files
       files = ['file-list', 'bill-of-materials']
       files.push get_files.map(&:path)
       files.push get_configfiles.map(&:path)
+      if @platform.is_windows?
+        files.flatten.map { |f| "$$(cygpath -ml #{f})" }
```
haus: how about using long flags here?

melissa: Good call. I'll update in a moment.

melissa: @haus updated

lib/vanagon/project.rb
```diff
@@ -232,6 +232,11 @@ def get_tarball_files
       files = ['file-list', 'bill-of-materials']
       files.push get_files.map(&:path)
       files.push get_configfiles.map(&:path)
+      if @platform.is_windows?
+        files.flatten.map { |f| "$$(cygpath --mixed --long-name #{f})" }
```
haus: do you want to wrap `#{f}` in single quotes in case there are spaces?

melissa: oooh, good call! Updated

haus:
as long as this still works on all the platforms it seems sane to me.

melissa:
@haus I tested with an el-7 build, and there were no issues

underscorgan:
:+1:

McdonaldSeanp:
:+1:

MikaelSmith:
Vanagon has other assumptions of Cygwin? I guess we'll abstract that out later :(

melissa:
@MikaelSmith https://github.com/puppetlabs/vanagon/blob/vanagon_for_the_win/lib/vanagon/platform/windows.rb#L192-L198 and probably all of https://github.com/puppetlabs/vanagon/blob/vanagon_for_the_win/lib/vanagon/platform/windows.rb#L68-L75

MikaelSmith:
Ah, excepting https://github.com/puppetlabs/vanagon/blob/vanagon_for_the_win/lib/vanagon/platform/windows.rb#L173 it could've also been MSYS sshd.

PR #256, opened 2016-01-11
(RE-5863) - Teach Vanagon harvest files for Wix

jcoconnor:
Outline package definition scripts for MSI generation.
First push up for discussion purposes only.

### DIFF
--- a/lib/vanagon/platform/windows.rb
+++ b/lib/vanagon/platform/windows.rb
@@ -83,18 +83,15 @@ class Vanagon
         ["mkdir -p output/#{target_dir}",
         "mkdir -p $(tempdir)/#{project.name}/staging",
         "gunzip -c #{project.name}-#{project.version}.tar.gz | '#{@tar}' -C '$(tempdir)/#{project.name}/staging' --strip-components 1 -xf -",
-        # Run the Heat command in a single pass - use the filter.xslt to filter out components for "special" treatment (i.e. services)
+        # Run the Heat command in a single pass
         # Heat command documentation at: http://wixtoolset.org/documentation/manual/v3/overview/heat.html
         #   dir <directory> - Traverse directory to find all sub-files and directories.
-        #   -v              - verbose
         #   -ke             - Keep Empty directories
-        #   -indent <n>     - Indentation multiple (overrides default of 4)
         #   -cg             - Component Group Name
         #   -gg             - Generate GUIDS now
         #   -dr             - Directory reference to root directories (cannot contains spaces e.g. -dr MyAppDirRef)
         #   -sreg           - Suppress registry harvesting.
         #   -var <variable> - Substitute File/@Source="SourceDir" with a preprocessor or a wix variable
-        #   -out <file>     - Output File
         "cd $(tempdir)/pl-libffi/staging; \"$$WIX/bin/heat.exe\" dir tools -v -ke -indent 2 -cg #{cg_name} -gg -dr #{dir_ref} -sreg -var var.StageDir -out wix/#{project.name}-harvest.wxs",
         ]
       end
--- a/lib/vanagon/platform/windows.rb
+++ b/lib/vanagon/platform/windows.rb
@@ -85,6 +85,17 @@ class Vanagon
         "mkdir -p $(tempdir)/#{project.name}/staging",
         "gunzip -c #{project.name}-#{project.version}.tar.gz | '#{@tar}' -C '$(tempdir)/#{project.name}/staging' --strip-components 1 -xf -",
         # Run the Heat command in a single pass - use the filter.xslt to filter out components for "special" treatment (i.e. services)
+        # Heat command documentation at: http://wixtoolset.org/documentation/manual/v3/overview/heat.html
+        #   dir <directory> - Traverse directory to find all sub-files and directories.
+        #   -v              - verbose
+        #   -ke             -
+        #   -indent <n>     -
+        #   -cg             - Component Group Name
+        #   -gg             - Generate GUIDS now
+        #   -dr             -
+        #   -sreg           -
+        #   -var <variable> -
+        #   -out <file>     -
         "cd $(tempdir)/pl-libffi/staging; \"$$WIX/bin/heat.exe\" dir tools -v -ke -indent 2 -cg #{cg_name} -gg -dr #{dir_ref} -sreg -var var.StageDir -out wix/#{project.name}-harvest.wxs",
         ]
       end
--- a/lib/vanagon/platform/windows.rb
+++ b/lib/vanagon/platform/windows.rb
@@ -92,7 +92,7 @@ class Vanagon
         #   -dr             - Directory reference to root directories (cannot contains spaces e.g. -dr MyAppDirRef)
         #   -sreg           - Suppress registry harvesting.
         #   -var <variable> - Substitute File/@Source="SourceDir" with a preprocessor or a wix variable
-        "cd $(tempdir)/pl-libffi/staging; \"$$WIX/bin/heat.exe\" dir tools -v -ke -indent 2 -cg #{cg_name} -gg -dr #{dir_ref} -sreg -var var.StageDir -out wix/#{project.name}-harvest.wxs",
+        "cd $(tempdir)/#{project.name}; \"$$WIX/bin/heat.exe\" dir staging -v -ke -indent 2 -cg #{cg_name} -gg -dr #{dir_ref} -sreg -var var.StageDir -out wix/#{project.name}-harvest.wxs",
         ]
       end
 
--- a/lib/vanagon/platform/windows.rb
+++ b/lib/vanagon/platform/windows.rb
@@ -6,9 +6,9 @@ class Vanagon
       # @param project [Vanagon::Project] project to build a windows package of
       # @return [Array] list of commands required to build a windows package for the given project from a tarball
       def generate_package(project)
-        # If nothing is passed in as platform type, default to building a nuget package
-        # We should default to building an MSI once that code has been implemented
         case project.platform.package_type
+        when "msi"
+          return generate_msi_package(project)
         when "nuget"
           return generate_nuget_package(project)
         else
@@ -21,9 +21,9 @@ class Vanagon
       # @param project [Vanagon::Project] project to name
       # @return [String] name of the windows package for this project
       def package_name(project)
-        # If nothing is passed in as platform type, default to a nuget package
-        # We should default to an MSI once that code has been implemented
         case project.platform.package_type
+        when "msi"
+          return msi_package_name(project)
         when "nuget"
           return nuget_package_name(project)
         else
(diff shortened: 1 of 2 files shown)


lib/vanagon/platform/windows.rb
```diff
@@ -8,10 +8,13 @@ class Windows < Vanagon::Platform
       def generate_package(project)
         # If nothing is passed in as platform type, default to building a nuget package
         # We should default to building an MSI once that code has been implimented
+        # SUGGEST WE TAKE DEFAULT OUT ALTOGETHER - WE MUST HAVE A package_type DEFINED on Windows
```
melissa: Addressed in https://github.com/puppetlabs/vanagon/pull/258

lib/vanagon/platform/windows.rb
```diff
@@ -75,7 +80,39 @@ def generate_nuget_package(project)
...
+      # [TBD - More to add]
+      #
+      # @param project [Vanagon::Project] project to build a msi package of
+      # @return [Array] list of commands required to build an msi package for the given project from a tarball
+      def generate_msi_package(project)
+        target_dir = project.repo ? output_dir(project.repo) : output_dir
+        ["mkdir -p output/#{target_dir}",
+        "mkdir -p $(tempdir)/STAGING",
+        # need to do things here.
+        "cp file-list $(tempdir)/#{project.name}/tools/file-list.txt", # May not need this (nuget only)
```
melissa: we only need this if we need a list of files for the packaging process. This file looks something like:
```
C:/Program Files/Puppet Labs
C:/Program Files/Puppet Labs/Puppet
C:/Program Files/Puppet Labs/Puppet/bin
C:/Program Files/Puppet Labs/Puppet/bin/ruby.exe
C:/ProgramData/PuppetLabs
```
It includes a listing of files, directories, and empty directories.

lib/vanagon/platform/windows.rb
```diff
@@ -75,7 +80,39 @@ def generate_nuget_package(project)
...
+        ]
+      end
+
+      # Method to derive the msi (Windows Installer) package name for the project.
+      #
+      # @param project [Vanagon::Project] project to name
+      # @return [String] name of the windows package for this project
+      def msi_package_name(project)
+        # Decided to use native project version in hope msi versioning doesn't have same resrictions as nuget
+        "#{project.name}-#{@architecture}-#{project.version, project.release}.msi"
```
melissa: May I suggest `"#{project.name}-#{project.version}.{project.release}-#{@architecture}.msi"`

melissa: I would be very surprised if we ran into similar restrictions to what we has with nuget versioning

lib/vanagon/platform/windows.rb
jcoconnor: Have rebased to changes in #258 - Had to resolve a conflict over the generate_package/package_name methods.
Also removed comments w.r.t. default as they are no longer relevant here.

lib/vanagon/platform/windows.rb
jpogran: The filename of the msi can be named what we want, there isn't a version restriction to the filename. The version in the wxs file is something that has rules to it that we need to worry about.

lib/vanagon/platform/windows.rb
```diff
@@ -67,7 +67,39 @@ def generate_nuget_package(project)
...
+        ]
+      end
+
+      # Method to derive the msi (Windows Installer) package name for the project.
+      #
+      # @param project [Vanagon::Project] project to name
+      # @return [String] name of the windows package for this project
+      def msi_package_name(project)
+        # Decided to use native project version in hope msi versioning doesn't have same resrictions as nuget
+        "#{project.name}-#{@architecture}-#{project.version}.msi"
```
melissa: Again, we should keep this as close to the current versioning scheme as we can. This should be `"#{project.name}-#{project.version}.#{project.release}-#{@architecture}.msi"`, which corresponds with what we already are shipping `puppet-agent-1.3.2-x86.msi`

lib/vanagon/platform/windows.rb
```diff
@@ -67,7 +67,39 @@ def generate_nuget_package(project)
...
+        ]
+      end
+
+      # Method to derive the msi (Windows Installer) package name for the project.
+      #
+      # @param project [Vanagon::Project] project to name
+      # @return [String] name of the windows package for this project
+      def msi_package_name(project)
+        # Decided to use native project version in hope msi versioning doesn't have same resrictions as nuget
+        "#{project.name}-#{project.version}.#{project.release}-#{@architecture}.msi"
```
melissa: :+1:

lib/vanagon/platform/windows.rb
```diff
@@ -21,9 +21,9 @@ def generate_package(project)
       # @param project [Vanagon::Project] project to name
       # @return [String] name of the windows package for this project
       def package_name(project)
-        # If nothing is passed in as platform type, default to a nuget package
-        # We should default to an MSI once that code has been implimented
```
melissa: :+1: with the new vanagon updates, we're setting the default to MSI. So if people want to build nuget packages, or build packages while we are still working on the MSI stuff, they have to explicitely set package_type to nuget.

lib/vanagon/platform/windows.rb
```diff
@@ -67,7 +67,39 @@ def generate_nuget_package(project)
         "cp $(tempdir)/#{project.name}/#{project.name}-#{@architecture}.#{nuget_package_version(project.version, project.release)}.nupkg ./output/#{target_dir}/#{nuget_package_name(project)}"]
       end
 
-      # Method to derive the package name for the project.
+      # The specific bits used to generate a windows msi package for a given project
+      # [TBD - More to add]
+      #
+      # @param project [Vanagon::Project] project to build a msi package of
+      # @return [Array] list of commands required to build an msi package for the given project from a tarball
+      def generate_msi_package(project)
```
melissa: So, what is this going to look like? I'm assuming this is where we're going to call harvest, heat, light, etc? What is the order that's going to happen in? Do we know what WXS files we have to copy over, and where we have to copy them to get the commands to run correctly? Do we just have to unpack the tarball into the stagedir, or somewhere else?

jcoconnor:
@melissa and @jpogran This PR is ready for submission.
I have limited the scope purely to the harvest operation.
Want to progress to RE-6237 next (service files). This will involve some further modification of the heat command (i.e. utilise the xslt file).

Also have tidied the commit history to remove a lot of the extraneous comments etc. and make it clearer.

lib/vanagon/platform/windows.rb
```diff
@@ -66,6 +66,38 @@ def generate_nuget_package(project)
...
+        target_dir = project.repo ? output_dir(project.repo) : output_dir
+        cg_name = "compfiles"
+        dir_ref = "INSTALLDIR"
+        filters_xslt = ""  # will set this when we have setup erb for xslt file (RE-6236)
+        # Actual array of commands to be written to the Makefile
+        ["mkdir -p output/#{target_dir}",
+        "mkdir -p $(tempdir)/#{project.name}/staging",
+        "gunzip -c #{project.name}-#{project.version}.tar.gz | '#{@tar}' -C '$(tempdir)/#{project.name}/staging' --strip-components 1 -xf -",
+        # Run the Heat command in a single pass - use the filter.xslt to filter out components for "special" treatment (i.e. services)
+        "cd $(tempdir)/pl-libffi/staging; \"$$WIX/bin/heat.exe\" dir tools -v -ke -indent 2 -cg #{cg_name} -gg -dr #{dir_ref} -sreg -var var.StageDir -out wix/#{project.name}-harvest.wxs",
```
melissa: Why have pl-libffi? Do we need it for building the package? In my opinion, it'd be easier to leave out project.name/pl-libffi completely from $(tempdir). We're only building one project at a time, so ideally there shouldn't be a collision. Even if someone is kicking off multiple vanagon runs on the same host, we should be safe given we're building in a temp directory

melissa: Also, there are a lot of flags here. Can we expand them out for clarity?

lib/vanagon/platform/windows.rb
```diff
@@ -66,6 +66,38 @@ def generate_nuget_package(project)
...
+      # 2. Compile (candle) all the wxs files into wixobj files
+      # 3. Run light to produce the final MSI
+      #
+      # @param project [Vanagon::Project] project to build a msi package of
+      # @return [Array] list of commands required to build an msi package for the given project from a tarball
+      def generate_msi_package(project)
+        target_dir = project.repo ? output_dir(project.repo) : output_dir
+        cg_name = "compfiles"
+        dir_ref = "INSTALLDIR"
+        filters_xslt = ""  # will set this when we have setup erb for xslt file (RE-6236)
```
melissa: It looks like this isn't used anywhere. I'm wondering if we should leave it out until we're ready to implement this?

lib/vanagon/platform/windows.rb
melissa: Where does `wix/#{project.name}-harvest.wxs` come from? Is that automatically generated by running heat?

jcoconnor: Change tools to '.' if that works - otherwise move up a directory and parse staging.

lib/vanagon/platform/windows.rb
```diff
@@ -66,6 +66,38 @@ def generate_nuget_package(project)
...
+      # Question - should we break this down into some simpler Make tasks ?
+      # 1. Heat the directory tree to produce the file list
+      # 2. Compile (candle) all the wxs files into wixobj files
+      # 3. Run light to produce the final MSI
+      #
+      # @param project [Vanagon::Project] project to build a msi package of
+      # @return [Array] list of commands required to build an msi package for the given project from a tarball
+      def generate_msi_package(project)
+        target_dir = project.repo ? output_dir(project.repo) : output_dir
+        cg_name = "compfiles"
```
melissa: Will this always be `compfiles` for eveything, or will this change?

lib/vanagon/platform/windows.rb
```diff
@@ -66,6 +66,38 @@ def generate_nuget_package(project)
...
+        dir_ref = "INSTALLDIR"
```
melissa: What does this correspond to? It looks like in puppet_for_the_win it's the source directory for each component? Am I misinterpreting that?

lib/vanagon/platform/windows.rb
melissa: It looks like `-sreg` is the only new flag you added. Why?

melissa: Also, I assume you are planning to add `-t \"#{filters_xslt}\"` when we have the service file stuff working. This is probably a question for then, but what if that variable is empty?

lib/vanagon/platform/windows.rb
melissa: Oh wait, yeah, I was. According to https://github.com/puppetlabs/puppet_for_the_win/blob/b515d41c89ade27d8b68d02ccbf24cf7ebed5a71/tasks/windows/windows.rake#L102, this makes sense

lib/vanagon/platform/windows.rb
jcoconnor: @Melissa At the moment yes unless we decide to break it up to sub-components.
This is what we were discussing with @jpogran

lib/vanagon/platform/windows.rb
jcoconnor: Ok

lib/vanagon/platform/windows.rb
jcoconnor: -sreg suppresses registry harvesting - I need to check this with @jpogran, @ferventcoder and @Iristyle, but I don't think we need to harvest registry entries from any of the dlls.

lib/vanagon/platform/windows.rb
```diff
@@ -66,6 +66,48 @@ def generate_nuget_package(project)
...
+        #   -v              - verbose
+        #   -ke             - Keep Empty directories
+        #   -indent <n>     - Indentation multiple (overrides default of 4)
+        #   -cg             - Component Group Name
+        #   -gg             - Generate GUIDS now
+        #   -dr             - Directory reference to root directories (cannot contains spaces e.g. -dr MyAppDirRef)
+        #   -sreg           - Suppress registry harvesting.
+        #   -var <variable> - Substitute File/@Source="SourceDir" with a preprocessor or a wix variable
+        #   -out <file>     - Output File
+        "cd $(tempdir)/pl-libffi/staging; \"$$WIX/bin/heat.exe\" dir tools -v -ke -indent 2 -cg #{cg_name} -gg -dr #{dir_ref} -sreg -var var.StageDir -out wix/#{project.name}-harvest.wxs",
```
melissa: Are there not expanded versions of these flags? (i.e., `--verbose` instead of `-v`)

jcoconnor: Not according to the [documentation](http://wixtoolset.org/documentation/manual/v3/overview/heat.html)

melissa: Darn. That sucks. That would have been a lot nicer to look at. Since we're including a link to this documentation, it might make more sense to only have specific explanations for special/confusing commands rather then copying the command definitions from the link here?

I still think you should remove referenced to filter.xslt until we actually implement that aspect of this code though.

jcoconnor: @melissa I have only copied the options we are using, but I guess we could remove some of the obvious ones. Tidy up coming when I get home.
Will also remove the xslt bit

melissa:
:+1: I think this is a good start. If we decide to alter the approach, we can do that. I just want to make sure everyone looks over the comments and understands what's happening

PR #257, opened 2016-01-12
(CPR-264) Rewrite chocolatey packaging scripts

melissa:
Our current nuget packaging scripts make a few assumptions that have
turned out to be false. Namely, these assumptions surround the path
specifications for where we are installing. Previously, we were using a
list of files and directories picked up from the project and component
configs. This was fine, but we have a better source for this
information. We have a file-list.txt file that details all of the files
and directories (even if they are empty) that are to be included in the
package. These paths have already been normalized to deal with spaces
and shortcuts (if any are used). So any use of `Progra~1` will already
have been expanded to `Program Files`. This is especially important
because we cannot guarentee that `Progra~1` will expand to what we want
it to be on the system the package is installed on. We have to expand
all of these shortcuts prior to packaging anything up. We use these
powershell scripts to then copy over files and directories from the
chocolatey install location to the final install location on the system.

The uninstall script only cleans up files. It does leave an empty
directory system in place.

### DIFF
--- a/templates/windows/chocolateyInstall.ps1.erb
+++ b/templates/windows/chocolateyInstall.ps1.erb
@@ -1,26 +1,28 @@
 $toolsDir = "$(Split-Path -parent $MyInvocation.MyCommand.Definition)"
+$fileList = Join-Path -path $toolsDir -childpath "file-list.txt"
 
-<%- dirnames = get_directories.map {|d| d.path } -%>
-<%- dirnames.each do |dir| -%>
-  <%- windows_path = platform.convert_to_windows_path(dir) -%>
-if (Test-Path -path "<%= windows_path %>") {
-  Write-Debug "Warning: copying over '<%= windows_path %>'"
-} else {
-  New-Item -ItemType directory -Path "<%= windows_path %>"
+if (!(Test-Path -path "$fileList")) {
+  Throw "Unable to find file '$fileList', cannot proceed with install"
 }
-$origin = Join-Path -path "$toolsDir" -childpath "<%= dir.gsub('/', '\\') %>"
-Copy-Item -path "$origin\*" -destination "<%= windows_path.gsub('/', '\\') %>\" -Recurse -Force
-<%- end -%>
 
-<%- filenames = get_files.map {|f| f.path} -%>
-<%- configfilenames = get_configfiles.map {|cf| cf.path} -%>
-<%- (filenames + configfilenames).each do |file| -%>
-  <%- windows_path = platform.convert_to_windows_path(file) -%>
-if (Test-Path -path "<%= windows_path %>") {
-  Write-Debug "Warning: copying over '<%= windows_path %>'"
-} elseif (!(Test-Path -path "<%= File.dirname windows_path %>")) {
-  New-Item -ItemType directory -Path "<%= File.dirname windows_path %>"
+$lines = Get-Content "$fileList"
+foreach ($destination in $lines) {
+  $originFile = "$destination" -replace '^[a-zA-Z]:/',''
+  $origin = Join-Path -path "$toolsDir" -childpath "$originFile"
+  if (Test-Path -path "$origin") {
+    $parent = Split-Path -path "$destination" -parent
+    if (!(Test-Path -path "$parent")) {
+      New-Item -ItemType directory -Path "$parent"
+    }
+    if (Test-Path -path "$destination") {
+      Write-Debug "Overwriting '$destination'"
+    }
+    Copy-Item -path "$origin" -destination "$destination" -Force
+  } else {
+    # If the item we are trying to copy over does not exist in our source directory,
+    # we assume it is an empty directory and simply create one in its place. There is
+    # a possibility that this will hide an error where there is actually a missing
+    # file. However, this is such a slim possibity, this action was deemed safe.
+    New-Item -ItemType directory -Path "$destination"
+  }
 }
-$origin = Join-Path -path "$toolsDir" -childpath "<%= file.gsub('/', '\\') %>"
-Copy-Item -path "$origin" -destination "<%= windows_path.gsub('/', '\\') %>" -Force
-<%- end -%>
--- a/templates/windows/chocolateyUninstall.ps1.erb
+++ b/templates/windows/chocolateyUninstall.ps1.erb
@@ -1,54 +1,24 @@
 $toolsDir = "$(Split-Path -parent $MyInvocation.MyCommand.Definition)"
+$fileList = Join-Path -path $toolsDir -childpath "file-list.txt"
 
-$contentFile=(join-path -path "$toolsDir" -childpath "file-list") + ".txt"
-if (Test-Path -path "$contentFile") {
-  Write-Debug "Removing files defiend in `'$contentFile`'"
-  $contents=get-content "$contentFile"
-  foreach ($fileInstalled in $contents) {
-    $fileInstalled=$fileInstalled.Replace("/cygdrive/c", "C:")
-    $fileInstalled=$fileInstalled.Replace("/", "\")
-    if (Test-Path -path "$fileInstalled" -PathType Container) {
-      Write-Debug "Not removing directory $fileInstalled"
-    } elseif (Test-Path -path "$fileInstalled") {
-      remove-item -Path "$fileInstalled" -Force
-    } else {
-      Write-Debug "Skipping missing file $fileInstalled"
+if (!(Test-Path -path "$fileList")) {
+  Throw "Unable to find file '$fileList', cannot proceed with uninstall"
+}
+
+$lines = Get-Content "$fileList"
+foreach ($file in $lines) {
+  if (Test-Path -path "$file") {
+    # We cannot guarentee a directory is only populated with files from
+    # this package, so we cannot whole-sale remove directories. We could
+    # check to see if a directory is empty after we remove all the files,
+    # but that would still end up with some reminantes of our skeletal
+    # directory structure. It doesn't seem worth it ATM. As is, only remove
+    # things that are files, and do not remove any directories.
+    if (!((Get-Item "$file") -is [System.IO.DirectoryInfo])) {
+      Write-Debug "Removing '$file'"
+      remove-item -Path "$file" -Force
     }
-  }
-} else {
-  Write-Debug "'$contentFile' not found, not removing any files in non-standard chocolatey location"
-}
-
-<%- dirnames = get_directories.map {|d| d.path } -%>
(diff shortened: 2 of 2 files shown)


ferventcoder:
So crazy story - anything in the package folder is going to get removed on uninstall.

templates/windows/chocolateyInstall.ps1.erb
```diff
@@ -1,26 +1,40 @@
...
-} elseif (!(Test-Path -path "<%= File.dirname windows_path %>")) {
-  New-Item -ItemType directory -Path "<%= File.dirname windows_path %>"
+$reader = [System.IO.File]::OpenText("$fileList")
+try {
+  for (;;) {
+    $destination = $reader.ReadLine()
+    if ($destination -eq $null) { break }
+    $originFile = "$destination" -replace '^[a-zA-Z]:/',''
+    $origin = Join-Path -path "$toolsDir" -childpath "$originFile"
+    if (Test-Path -path "$origin") {
```
ferventcoder: If I follow this logic correctly - If the `$origin` location exists, then do the following. Else if the `$destination` doesn't exist, create it.

templates/windows/chocolateyInstall.ps1.erb
```diff
@@ -1,26 +1,40 @@
...
-<%- configfilenames = get_configfiles.map {|cf| cf.path} -%>
-<%- (filenames + configfilenames).each do |file| -%>
-  <%- windows_path = platform.convert_to_windows_path(file) -%>
-if (Test-Path -path "<%= windows_path %>") {
-  Write-Debug "Warning: copying over '<%= windows_path %>'"
-} elseif (!(Test-Path -path "<%= File.dirname windows_path %>")) {
-  New-Item -ItemType directory -Path "<%= File.dirname windows_path %>"
+$reader = [System.IO.File]::OpenText("$fileList")
+try {
+  for (;;) {
```
ferventcoder: what is this short for?

templates/windows/chocolateyInstall.ps1.erb
```diff
@@ -1,26 +1,40 @@
...
+      }
+    } else {
+      if (!(Test-Path -path "$destination")) {
+        New-Item -ItemType directory -Path "$destination"
+      }
+    }
+  }
+}
+finally {
+  $reader.Close()
```
ferventcoder: Let's not hold the file open - instead just use `[System.IO.File]::ReadAllText("$fileList")` on line 9 and then loop over lines in the string.

templates/windows/chocolateyInstall.ps1.erb
ferventcoder: I get the check to ensure you have the file.

ferventcoder: But I see a small issue in here that you will have something in the file list that should be a file and not a directory, but a directory will be created. I would hope for an error instead.

templates/windows/chocolateyUninstall.ps1.erb
```diff
@@ -1,54 +1,27 @@
...
-    } elseif (Test-Path -path "$fileInstalled") {
-      remove-item -Path "$fileInstalled" -Force
-    } else {
-      Write-Debug "Skipping missing file $fileInstalled"
-    }
-  }
-} else {
-  Write-Debug "'$contentFile' not found, not removing any files in non-standard chocolatey location"
+if (!(Test-Path -path "$fileList")) {
+  Write-Host "Unable to find file '$fileList', cannot proceed with uninstall"
```
ferventcoder: Write-Warning

templates/windows/chocolateyInstall.ps1.erb
melissa: I'm not sure. I'm assuming it corresponds to `while true`. I copied it from a script I found... Seems to work..

templates/windows/chocolateyUninstall.ps1.erb
```diff
@@ -1,54 +1,27 @@
...
+  exit
```
ferventcoder: Are there reasons why the file list wouldn't exist? Should this be an error or merely a warning?

templates/windows/chocolateyUninstall.ps1.erb
```diff
@@ -1,54 +1,27 @@
...
-    Write-Debug "Removing any empty directories in '<%= windows_path %>'"
-    # Remove any empty subdirectories
-    $SearchRoot = "<%= windows_path %>"
-    Get-ChildItem -Path $SearchRoot -Recurse |
-      Where-Object {$_.PSIsContainer -eq $true -and (Get-ChildItem -Path $_.FullName) -eq $null} |
-        Remove-Item
-    # Remove the main directory if it's now empty after removing all empty subdirectories
-    Where-Object {(Get-ChildItem -Path "<%= windows_path %>" -Recurse) -eq $null} | Remove-Item
+
+$reader = [System.IO.File]::OpenText("$fileList")
```
ferventcoder: Same comments on just reading in the whole file at once and closing the stream.

templates/windows/chocolateyInstall.ps1.erb
ferventcoder: Any reason why we switched?

templates/windows/chocolateyInstall.ps1.erb
melissa: Hmm, that's a possibility. My thought process behind this was dealing with empty directories. Currently, if we have an empty directory, chocolatey/nuget seem to ignore it. So, this first checks if the origin exists (in the chocolatey install directory) and if it doesn't, it assumes it's an empty directory. I'm not sure of a better way to handle the possibility the missing element is a file... Any ideas?

templates/windows/chocolateyUninstall.ps1.erb
melissa: Yup, probably should be an error. The only reasons I could think of that the file would be missing is if we created a bad package that was missing the file or if the user manually deleted the file

templates/windows/chocolateyInstall.ps1.erb
melissa: None in particular. I forgot we were doing something similar in the uninstall script, and this was the first thing I found on the internet that seemed to work. Is there a better way of doing this?

melissa: Should I use `Get-Content "$fileList"` rather then `System.IO.File]::ReadAllText("$fileList")`?

templates/windows/chocolateyUninstall.ps1.erb
melissa: What about `Throw` to terminate the script?

templates/windows/chocolateyInstall.ps1.erb
```diff
@@ -1,26 +1,40 @@
...
+try {
+  for (;;) {
+    $destination = $reader.ReadLine()
+    if ($destination -eq $null) { break }
+    $originFile = "$destination" -replace '^[a-zA-Z]:/',''
+    $origin = Join-Path -path "$toolsDir" -childpath "$originFile"
+    if (Test-Path -path "$origin") {
+      if ((Get-Item "$origin") -is [System.IO.DirectoryInfo]) {
+        if (!( Test-Path -path "$destination")) {
+          New-Item -ItemType directory -Path "$destination"
```
melissa: I just realized this isn't going to preserve any directory permissions we set...

MikaelSmith: I'm not sure we set any that make sense when packaging the source. I think creating them as the current user is a reasonable behavior.

templates/windows/chocolateyInstall.ps1.erb
MikaelSmith: That seems like an ok approach, a file that's missing but in the file list seems like a pretty unlikely bug in packaging. I'd like to see a comment about that assumption in the script.

templates/windows/chocolateyInstall.ps1.erb
```diff
@@ -1,26 +1,40 @@
...
+        }
+      } else {
+        $parent = Split-Path -path $destination -parent
+        if (!(Test-Path -path "$parent")) {
+          New-Item -ItemType directory -Path "$parent"
+        }
+        if (Test-Path -path "$destination") {
+          Write-Host "Overwriting '$destination'"
```
MikaelSmith: Did this happen in practice?

templates/windows/chocolateyInstall.ps1.erb
MikaelSmith: Why not hold the file open? Streaming potentially large files is usually a good practice.

templates/windows/chocolateyUninstall.ps1.erb
```diff
@@ -1,54 +1,27 @@
...
+try {
+  for (;;) {
+    $file = $reader.ReadLine()
+    if ($file -eq $null) { break }
+    if (Test-Path -path "$file") {
+      if (!((Get-Item "$file") -is [System.IO.DirectoryInfo])) {
```
MikaelSmith: Could we remove directories? If we know that directories will always be listed first (or last), we could iterate so that the directory is the last thing we delete, and if it's empty delete it.

melissa: I tried that in the last iteration (see the old code), and I just feel like it's a lot of work for not a lot of gain. We have to make sure we start at the leaves and work down. That's too much for me to want to code. We can't guarantee any of the directories are solely populated by the package we are removing (see `C:/Program Files/Puppet Labs`), so we need to validate the directories are empty before attempting to remove them.

templates/windows/chocolateyInstall.ps1.erb
melissa: If you're asking if I've seen this output, then no, I haven't. I figured it'd be nice to know if your files are being overwritten though...

templates/windows/chocolateyInstall.ps1.erb
melissa: Okay, I know for linux we have the ability to specify directory owner/permissions. I know those are different on windows, but I wasn't sure if that would carry over or not once we've updated the directory method to deal with windows specific setting...

templates/windows/chocolateyInstall.ps1.erb
melissa: comment added

MikaelSmith:
:+1:

templates/windows/chocolateyInstall.ps1.erb
ferventcoder: :+1:

PR #258, opened 2016-01-13
Maint/vanagon for the win/nitpick fixes

melissa:


McdonaldSeanp:
:+1: pending travis

shrug:
:+1:

lib/vanagon/platform.rb
```diff
@@ -230,7 +230,7 @@ def is_unix?
     #
     # @return [true, false] true if it is a windows variety, false otherwise
     def is_windows?
-      return !!@name.match(/^win-.*$/)
+      return !!@name.match(/^windows-.*$/)
```
melissa: cc/ @jcoconnor and @MikaelSmith this will impact current vanagon projects that are building for windows

MikaelSmith: Ok, they'll have to change their platform names?

melissa: Yeah, so rather than `win-2012-x64`, they'll need to be `windows-2012-x64`

lib/vanagon/platform/windows.rb
```diff
@@ -170,7 +162,7 @@ def output_dir(target_repo = "")
       #
       # @return [String] the cygwin drive root
       def drive_root
-        "/cygdrive/c"
+        "C:"
```
melissa: We need cygpath to get rid shortcuts (`Progra~1`) and to specify `PATH` in the cygwin build environment

lib/vanagon/platform/windows.rb
```diff
@@ -71,7 +63,7 @@ def generate_nuget_package(project)
         "cp chocolateyInstall.ps1 chocolateyUninstall.ps1 $(tempdir)/#{project.name}/tools/",
         "cp file-list $(tempdir)/#{project.name}/tools/file-list.txt",
         "gunzip -c #{project.name}-#{project.version}.tar.gz | '#{@tar}' -C '$(tempdir)/#{project.name}/tools' --strip-components 1 -xf -",
-        "(cd $(tempdir)/#{project.name} ; #{self.drive_root}/ProgramData/chocolatey/bin/choco.exe pack #{project.name}.nuspec)",
+        "(cd $(tempdir)/#{project.name} ; C:/ProgramData/chocolatey/bin/choco.exe pack #{project.name}.nuspec)",
```
haus: Why move away from using drive_root here?

lib/vanagon/platform/windows.rb
MikaelSmith: I guess eventually we'll come up with a better way.

lib/vanagon/platform/windows.rb
melissa: We're on the builder machine. It's safe to assume this will always be in this location. That, and I want to get rid of `drive_root`. This was the only place we were using it in vanagon

haus: Is removing drive_root going to be part of the PR too?

melissa: I'd really like it to be.... But I was waiting to get the folks who originally proposed we use it on board

lib/vanagon/platform/windows.rb
melissa: Does that mean your on board with removing those two methods?

MikaelSmith: The only reason I would see to keep them is if they can be overridden from a platforms file, and someone wants to setup a build environment on a drive besides `C:`. I suspect that would take more refactoring to make possible.

melissa: Hmm, yeah, you're right. That would take more refactoring to make possible. But that is a very compelling point you make. I'm thinking we do that if someone opens a ticket or asks for it specifically. There are currently only two places in vanagon that reference `C:`. The build environment we're depending on is so strange, I have a feeling this will be the least of our problems when the public starts to take advantage of this. But, if you feel strongly about it, I can try to make it happen.

lib/vanagon/platform/windows.rb
```diff
@@ -196,6 +172,7 @@ def initialize(name)
         @num_cores = "/usr/bin/nproc"
         @install = "/usr/bin/install"
         @copy = "/usr/bin/cp"
+        @package_type = "msi"
```
melissa: Hmm, this probably should be nuget for now... Eventually I want this to be MSI, but since we don't have support for building msis yet, projects will have to explicitily set package_type to nuget if they want to start testing this out

lib/vanagon/platform/windows.rb
MikaelSmith: I don't feel strongly about it.

lib/vanagon/platform/windows.rb
melissa: Nope, I think we should leave it like this. People should be deliberate about the type of package they are building and testing. They can remove the line once we have the MSI work done.

lib/vanagon/platform/windows.rb
melissa: So you're okay if I remove `drive_root`?

MikaelSmith: Yes.

lib/vanagon/platform/windows.rb
melissa: And I've gotten the okay to remove it!

underscorgan:
:+1:

PR #261, opened 2016-01-14
(RE-6106) Allow for putting items in /usr/local on debian

stahnma:
Prior to this, dh_usrlocal ran and would cause package builds to fail if
things were placed in /usr/local. This patch simply declares an override
of the dh_userlocal target that is a noop.


ScottGarman:
:+1:

PR #262, opened 2016-01-18
Add upstream URL to component sources

stahnma:
When trying to work with puppet-agent and other PL software outside of
Puppet Labs, things are difficult. This is an attempt to make it
slightly easier. This commit adds a new DSL method called 'upstream_url'
to a component source. This URL is used for fetching content if the
'url' specified fails.

The advantage of this is that we'll now have the upstream URL in the
component, and you won't have to spend your first 25 minutes download
sources and making a mirror on a webserver to use vanagon's url rewrite
engine.

Basically, this just means you need something like:

    pkg.upstream_url "http://ftp.gnu.org/gnu/#{pkg.get_name}/#{pkg.get_name}-#{pkg.get_version}.tar.gz"

Upstream URL is always the second choice behind pkg.url.

This patch has no affect on git urls.

### DIFF
--- a/lib/vanagon/component.rb
+++ b/lib/vanagon/component.rb
@@ -12,7 +12,7 @@ class Vanagon
     attr_accessor :settings, :platform, :patches, :requires, :service, :options
     attr_accessor :directories, :replaces, :provides, :cleanup_source
     attr_accessor :sources, :preinstall_actions, :postinstall_actions
-    attr_accessor :preremove_actions, :postremove_actions, :license
+    attr_accessor :preremove_actions, :postremove_actions, :license, :upstream_url
 
     # Loads a given component from the configdir
     #
@@ -105,7 +105,7 @@ class Vanagon
     #
     # @param workdir [String] working directory to put the source into
     def get_source(workdir)
-      if @url
+      if @url || @options[:upstream_url]
         @source = Vanagon::Component::Source.source(@url, @options, workdir)
         @source.fetch
         @source.verify
--- a/lib/vanagon/component/dsl.rb
+++ b/lib/vanagon/component/dsl.rb
@@ -270,6 +270,10 @@ class Vanagon
         @component.url = the_url
       end
 
+      def upstream_url(the_url)
+        @component.options[:upstream_url] = the_url
+      end
+
       # Sets the md5 sum to verify the sum of the source
       #
       # @param md5 [String] md5 sum of the source for verification
@@ -310,8 +314,8 @@ class Vanagon
       # @param url [String] url of the source
       # @param ref [String] Used for git sources, must be a git ref of some sort
       # @param sum [String] sum used to validate http and file sources
-      def add_source(url, ref: nil, sum: nil)
-        @component.sources << OpenStruct.new(:url => url, :ref => ref, :sum => sum)
+      def add_source(url, ref: nil, sum: nil, upstream_url: nil)
+        @component.sources << OpenStruct.new(:url => url, :ref => ref, :sum => sum, :upsteam_url => upstream_url)
       end
 
       # Adds a directory to the list of directories provided by the project, to be included in any packages of the project
--- a/lib/vanagon/component/source.rb
+++ b/lib/vanagon/component/source.rb
@@ -62,9 +62,9 @@ class Vanagon
         uri_scheme = url_match[1] if url_match
         local_source =  case uri_scheme
                         when /^http/
-                          Vanagon::Component::Source::Http.new(self.rewrite(url, 'http'), options[:sum], workdir)
+                          Vanagon::Component::Source::Http.new(self.rewrite(url, 'http'), options[:sum], workdir, options[:upstream_url])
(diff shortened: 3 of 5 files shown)


stahnma:
Working on tests, mostly put this up right now for comments.

stahnma:
Things I'm most questioning:

1. Should this also work for git urls?
2. How should ERRORs be bubbled up?

stahnma:
Found a bug with file:// sources, fixing coming.

haus:
I think a more generic approach should be considered here. If the general problem is: "sometimes the url doesn't work", then adding 1 more may fix it today, until that url also fails. It seems like if this is the road to making url downloads more stable, then an array of urls would be better, with each being tried before a failure declared. That also avoids needing to make a series of `_url` variables (`url`, `upstream_url`, `backup_upstream_url`, ...).

stahnma:
I think having an upstream_url makes a lot of sense in the DSL. Right now, you can't tell where we got the software. Both Deb and RPM have this in their metadata. We should have it in ours. The point wasn't to have a fallback, the point was to have it this build able off our VPN. The method of falling back was just an idea.

haus:
I always thought the url was going to be the upstream url. The only reason in puppet-agent that the urls aren't the upstream urls is because some upstream urls still don't work with vanagon's downloader.

Otherwise all of the urls would be upstream urls and the rewrite_engine would repoint them to buildsources for PL builds. If all of the urls were upstream urls, and the rewrite_engine were turned on for PL builds, would this work still be needed?

stahnma:
I think having the concept of upstream vs mirror is easier than the rewrite as an option, because you can modify per component vs whole thing.

mckern:
I hunted through this and here's my thoughts:

This specific implementation bolts a lot of complexity into existing methods, which sould probably be broken up into some new methods. I am interested in the idea of "yo that URL should be the upstream URL and we got this rewriter engine right here."

hqdefault

I **would** like to replace all of our `http://buildsources` specific URLs with the upstream ones but I'd also like the ability to specify an array of URLs as the `:url` instead of a single string (like some of the attributes you'd find in a Gemspec, as an example). What does lighting up rewriting look like?

stahnma:
I like the idea of URL being multi-valued. I don't like the idea that upstream URL isn't a separate entry all together and not just one of something in an array. 

Ideally, we'd have 

upstream_url "gnu.org"
url  [ 'place1', 'place2' ]

use upstream unless there is a url. If there is a url, try all array items before falling back to upstream. 

(I'd be in favor of dropping the rewrite facility that exists)

mckern:
@stahnma does this currently implement your proposal to try everything in `url` before trying `upstream_url`? And what if **only** `upstream_url` is defined (should it default to that?)?

stahnma:
@mckern it needs a little work. This does not handle multi-valued URL. I actually think that should be another PR all together. I started playing with it, and it's not a small changeset. 

My proposal here would be:

If url is not empty, use it. 
Otherwise fallback to upstream_url
Otherwise fail.

mckern:
If time permits and you want to hack on this after hours next week, I'll hack on it with you.

mckern:
@stahnma Rebase this, and let's hack on it in the working session this week.

lib/vanagon/component.rb
```diff
@@ -12,7 +12,7 @@ class Component
     attr_accessor :settings, :platform, :patches, :requires, :service, :options
     attr_accessor :directories, :replaces, :provides, :cleanup_source
     attr_accessor :sources, :preinstall_actions, :postinstall_actions
-    attr_accessor :preremove_actions, :postremove_actions, :license
+    attr_accessor :preremove_actions, :postremove_actions, :license, :upstream_url
```
mckern: nitpick, but I'd like to see the `:upstream_url` alongside the `:url`. I like logical groupings.

lib/vanagon/component/dsl.rb
```diff
@@ -270,6 +270,10 @@ def url(the_url)
         @component.url = the_url
       end
 
+      def upstream_url(the_url)
+        @component.options[:upstream_url] = the_url
```
mckern: What's the benefit to using `@component.options[:upstream_url]` instead of `@component.upstream_url` since it's defined as an accessor?

mckern:
I've been working on an initial refactor of the `Vanagon::Component::Source::Http` class tonight, and one of the glaring things that makes **this** PR frustrating is that `file://` URIs are handled by the `HTTP` class. I think that we could clean this up some if we moved moved `file://` into its own class and then just let the `Http` class handle HTTP/HTTPS connections. But that's a lot of work. So I'm working on seeing if I can spin the existing `Http` class into something made of more manageable chunks and then add trying the Upstream URL if a given URL doesn't work.

stahnma:
@mckern  Technically, we could have a URI handling class and file:///, git:// and http(s):// could all be handled there maybe? It looks like libcurl supports file and https (and maybe git) so we could hook into that...maybe?

PR #264, opened 2016-01-20
(maint) Rename templates to resources

melissa:
We need to have the option to use static packaging files if we need
them. This commit moves the templates directory to the resources
directory. Though this combines both templates and static resources in
the same place, we are always explicit about which files to parse. This
allows us a home for non-templates files.

This also changes some of the windows resources from templates to static
files.


haus:
This seems better targeted at master

melissa:
Fair point.

PR #265, opened 2016-01-20
(maint) Move templates dir to resources

melissa:
When building packages with vanagon, we need more then just templates.
There are occasions where we need static packaging resources in addition
to or instead of templates. This is all dependent on the type of package
we are creating. However, it's misleading to have static resources in a
directory called `templates`. This commit renames the templates
directory to resources. This directory will contain both static and
dynamic resources to be used when creating packages.


melissa:
I build an rpm for el7 with this change

MikaelSmith:
:+1:

MikaelSmith:
This will require moving resources to templates in every project that uses Vanagon when it's released? Feels like this should have a ticket (or however we track significant feature changes for Vanagon).

melissa:
@MikaelSmith this is internal to vanagon. As far as I'm aware, this shouldn't impact projects that use vanagon. cc/ @haus for confirmation

MikaelSmith:
Ah, it is internal-only. I got confused by the description, and it's only to support https://github.com/puppetlabs/vanagon/pull/252/files#r50304942.

melissa:
Yup. Should I reword the description?

melissa:
P.s., this successfully built an el-7 package

MikaelSmith:
No, I think it's ok.

mckern:
I am :+1: on this. It's a pretty straightforwardchangeset.

PR #268, opened 2016-01-27
(CPR-273) Allow repeatable builds

MikaelSmith:
Vanagon builds should be repeatable over the same temporary directory
(via make). Currently attempting to repeat a link fails the build
because the symlink already exists. Only attempt to create the symlink
if it doesn't already point to the correct file.

### DIFF
--- a/lib/vanagon/component/dsl.rb
+++ b/lib/vanagon/component/dsl.rb
@@ -244,7 +244,10 @@ class Vanagon
       # @param target [String] path to the desired symlink
       def link(source, target)
         @component.install << "#{@component.platform.install} -d '#{File.dirname(target)}'"
-        @component.install << "ln -s '#{source}' '#{target}'"
+        # Use a bash conditional to only create the link if it doesn't already point to the correct source.
+        # This allows rerunning the install step to be idempotent, rather than failing because the link
+        # already exists.
+        @component.install << "([[ '#{target}' -ef '#{source}' ]] || ln -s '#{source}' '#{target}')"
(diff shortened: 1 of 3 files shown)


MikaelSmith:
This appears to work for my problem, tested on CentOS 7.

lib/vanagon/component/dsl.rb
```diff
@@ -244,7 +244,10 @@ def install_configfile(source, target)
       # @param target [String] path to the desired symlink
       def link(source, target)
         @component.install << "#{@component.platform.install} -d '#{File.dirname(target)}'"
-        @component.install << "ln -s '#{source}' '#{target}'"
+        # Use a bash conditional to only create the link if it doesn't already point to the correct source.
+        # This allows rerunning the install step to be idempotent, rather than failing because the link
+        # already exists.
+        @component.install << "([[ '#{target}' -ef '#{source}' ]] || ln -s '#{source}' '#{target}')"
```
haus: If the link exists but doesn't point to the source, this will fail the build. Is that desirable?

MikaelSmith: That's the current behavior, and I intentionally preserved it.

MikaelSmith: Oh, I guess some of the concern was if it was a directory. From a `make` perspective we just want to make it a new thing.

haus: cool.

stahnma:
:+1:

PR #269, opened 2016-01-27
(RE-6236) Add Service Handling Files to Vanagon

jcoconnor:
Add in configurable filter (XSLT) file to filter out service components
from the main heat run in the build and update heat command to use transformed xslt template.


lib/vanagon/platform/windows.rb
```diff
@@ -37,6 +37,25 @@ def package_name(project)
       # @param name [String] name of the project
       # @param binding [Binding] binding to use in evaluating the packaging templates
       def generate_packaging_artifacts(workdir, name, binding)
+        # HOW DO I WORK OUT the Package Type Here - Can I use the binding as per next statement ??
+        # Or should we be splitting out this into a nuget and a windows sub-class - i.e. too many package_type
+        package_type = binding.eval("@project.platform.package_type")
```
jcoconnor: @melissa Is this the best way to pickup the package type - is it even legal Ruby ?

Think we need to sub-class the windows.rb file into two separate msi and nuget sub-classes as we appear to be using this conditional rather too many times.

resources/windows/msi/service-filter.xslt.erb
```diff
@@ -0,0 +1,19 @@
...
+  <xsl:output method="xml" indent="yes" />
+  <!--<xsl:strip-space elements="*"/>-->
+  <xsl:template match="@*|node()">
+    <xsl:copy>
+      <xsl:apply-templates select="@*|node()"/>
+    </xsl:copy>
+  </xsl:template>
+
+  <!-- Loop through the list of services and add filters for each component requiring Service type special treatment -->
+  <%- service_files.each do |service| -%>
```
jcoconnor: @melissa service_files is not defined yet - can we add it as an array for the msi/windows platform initialisation and then add to the array as we add service entries.
This should probably be a project (and not component) array, but will needs set in the component files.

resources/windows/msi/service.wxs.erb
```diff
@@ -0,0 +1,11 @@
+<?xml version='1.0' encoding='windows-1252'?>
+<Wix xmlns='http://schemas.microsoft.com/wix/2006/wi' xmlns:util="http://schemas.microsoft.com/wix/UtilExtension">
+
+<!-- include the set of service files -->
+
+  <%- service_files.each do |service| -%>
+    <?include $(sys.SOURCEFILEDIR)include/<%=  service.path %>' ?>
```
jpogran: I don't see the value of a WIX include file, if we need to have the same information in the erb templates as well. I had thought we were making the include file a yaml file for the template parts?

resources/windows/msi/service-filter.xslt.erb
jpogran: As noted in puppet-agent pr, I don't see the value in dynamically generating this now. We have to hardcode it in service.puppet.wxs right now until we can express that part in vanagon. Adding more complexity here IMHO is a maintenance burden.

lib/vanagon/platform/windows.rb
```diff
@@ -46,6 +65,20 @@ def generate_packaging_artifacts(workdir, name, binding)
...
 
+      # Method to generate the files required to build a windows msi package for the project
+      #
+      # @param workdir [String] working directory to stage the evaluated templates in
+      # @param name [String] name of the project
+      # @param binding [Binding] binding to use in evaluating the packaging templates
+      def generate_packaging_artifacts_msi(workdir, name, binding)
+        # nuget templates that do require a name change
+        erb_file(File.join(VANAGON_ROOT, "resources/windows/msi/service-filter.xslt.erb"), File.join(workdir, "#{name}.service-filter.xslt"), false, { :binding => binding })
+        erb_file(File.join(VANAGON_ROOT, "resources/windows/msi/main.wxs.erb"), File.join(workdir, "#{name}.main.wxs"), false, { :binding => binding })
```
jpogran: I would change the order here and put main.wxs.erb first, as its the most important one and it has to work. As this list gets longer (feature, directory, customaction, misc components, etc) to have it mixed in confuses things.

resources/windows/msi/main.wxs.erb
```diff
@@ -0,0 +1,781 @@
+<?xml version='1.0' encoding='windows-1252'?>
+<Wix xmlns='http://schemas.microsoft.com/wix/2006/wi'
+  xmlns:util="http://schemas.microsoft.com/wix/UtilExtension">
+  <!--
+  This is how to include wxi files.  We expect this file to define variables
+  like the version numbers and the upgrade UUID.
+  -->
+  <?include $(sys.SOURCEFILEDIR)include/puppet.wxi ?>
```
jpogran: This should be removed, as we are using erb templates

resources/windows/msi/main.wxs.erb
```diff
@@ -0,0 +1,781 @@
...
+      WIX_UPGRADE_DETECTED
+      -->
+    </InstallExecuteSequence>
+
+    <!--
+      The Features to install.  We're using ComponentGroupRef instead of ComponentRef since we're
+      generating the ruby wxs file using heat
+    -->
+    <Feature Id="$(var.OurProductNameWord)Runtime" Title="$(var.OurProductName) Runtime" Level="1">
+      <ComponentRef Id="RegistryEntries" />
```
jpogran: If we name the ComponentGroup purposefully in the heat generation, we can most likely reduce this list of comonentrefs to one.

resources/windows/msi/main.wxs.erb
```diff
@@ -0,0 +1,781 @@
...
+  <?include $(sys.SOURCEFILEDIR)include/puppet.wxi ?>
+
+  <!--
+  The Product Id should not be changed, the unique identifier enables
+  un-install of the MSI package easily.  More information about supporting
+  major upgrades is available at:
+  http://wix.sourceforge.net/manual-wix3/major_upgrade.htm
+  Name is made of localized product name and version number.
+  -->
+  <Product Id="$(var.ProductCode)" UpgradeCode="$(var.UpgradeCode)"
```
jpogran: Lots of variable replacement todo here

resources/windows/msi/service-filter.xslt.erb
jcoconnor: @jpogran This is the filter part - we are just populating a list of files that the heat stage ignores.
Its nothing more.

resources/windows/msi/service.wxs.erb
jcoconnor: @jpogran We could just use a wildcard include - the specific files will be copied from the product directories (i.e. puppet-agent), so we may need some associated code in the generate_package_articfacts method to copy the puppet-agent service files to the working area.

resources/windows/msi/main.wxs.erb
jcoconnor: Yes - quite a few of them can probably be directly replaced using the ERB operation.

resources/windows/msi/service.wxs.erb
jpogran: The `include` element is meant for `wxi` files, not `wxs` files. I haven't personally seen this work, although documentation implies it might. I haven't seen this as a common practice.

In the past I have never used this, instead passed the list of `wxs` files to light either with a `*.wxs` or a listing of each file.

resources/windows/msi/service-filter.xslt.erb
jpogran: I know what it is. My point is here its dynamic and based on a variable set somewhere else, in the service.puppet.wxs.erb its hardcoded. Two places we have to keep track.

resources/windows/msi/service.wxs.erb
jcoconnor: Ok - perhaps the file is superfluous so perhaps we should drop it altogether and just include the wxs files.

lib/vanagon/platform/windows.rb
```diff
@@ -37,6 +37,25 @@ def package_name(project)
       # @param name [String] name of the project
       # @param binding [Binding] binding to use in evaluating the packaging templates
       def generate_packaging_artifacts(workdir, name, binding)
+        # HOW DO I WORK OUT the Package Type Here - Can I use the binding as per next statement ??
+        # Or should we be splitting out this into a nuget and a windows sub-class - i.e. too many package_type
+        package_type = binding.eval("@platform.package_type")
+        case package_type
```
melissa: ummmmm, check out the `package_name` and `generate_package` methods above.... Lines 24 and 9 (`case project.platform.package_type`) have what you need. Notice it's contained in the project object, so you'll have to pass that in to access it.

Might I suggest not conditionally generating package artifacts. Generate all of them regardless. It doesn't take a lot of time to process templates, and they don't do any harm if we just leave them hanging around. We have the ability to only move in the ones we want when building the package in question.

i.e.,
```
"#{@copy} #{project.name}.nuspec $(tempdir)/#{project.name}/",
"#{@copy} chocolateyInstall.ps1 chocolateyUninstall.ps1 $(tempdir)/#{project.name}/tools/",
```

jcoconnor: @melissa Hmm trouble is that is generate artefacts is defined for each platform so would I need to fix it in multiple files?

Let's discuss tomorrow.

melissa: That's why you should generate all of the windows packaging artifacts (both nuget and msi) regardless. That will let us move on from this problem.

resources/windows/msi/main.wxs.erb
```diff
@@ -0,0 +1,781 @@
+<?xml version='1.0' encoding='windows-1252'?>
```
melissa: Let's keep this in line with other templates and name this file `project.wxs.erb`, and then rename it when we process it to be `puppet-agent.wxs` or `pe-client-tools.wxs`, whatever project we're building.

Notice we're doing this for nuget with `erb_file(File.join(VANAGON_ROOT, "resources/windows/nuget/project.nuspec.erb"), File.join(workdir, "#{name}.nuspec"), false, { :binding => binding })`

resources/windows/msi/main.wxs.erb
melissa: Is this going to be a per-project hard-coded file or something we can auto-generate in vanagon for each project?

lib/vanagon/platform/windows.rb
jcoconnor: Done

resources/windows/msi/main.wxs.erb
jcoconnor: Done

resources/windows/msi/service.wxs.erb
jcoconnor: Removed

melissa:
Sooo, there's 800 new lines in this PR. Can we only add the relevant lines and pull the rest out so we can review it? Otherwise, it's scary and overwhelming....

lib/vanagon/platform/windows.rb
```diff
@@ -32,11 +32,14 @@ def package_name(project)
       end
 
       # Method to generate the files required to build a windows package for the project
+      # Both the MSI and nuget generic articfacts are created for simplicity.
       #
       # @param workdir [String] working directory to stage the evaluated templates in
       # @param name [String] name of the project
       # @param binding [Binding] binding to use in evaluating the packaging templates
       def generate_packaging_artifacts(workdir, name, binding)
+        # First the nuget artefacts (MAINT ticket coming up here to fix the spelling of the function name)
```
melissa: the spelling of the function name? I think `generate_packaging_artifacts` is spelled correctly... it also need to stay as is because it's getting called at https://github.com/puppetlabs/vanagon/blob/master/lib/vanagon/driver.rb#L84

resources/windows/msi/project.wxs.erb
```diff
@@ -0,0 +1,775 @@
...
+    Language='1033'
+    Codepage='1252'
+    Version="$(var.ProductVersionNumber)"
+    Manufacturer="$(var.OurCompanyName)">
+
+    <Package
+      InstallerVersion='300'
+      InstallScope="perMachine"
+      Description="$(var.OurProductName)$(var.OurProductPlatformText) Installer"
+      Comments="http://www.puppetlabs.com/"
```
shermdog: We may want this as a variable also.

lib/vanagon/platform/windows.rb
jcoconnor: Indeed - as discovered, its the usual UK<-->US language variants. In this case the US wins.

resources/windows/msi/project.wxs.erb
jcoconnor: @shermdog Yeah - quite a few of the "Wix" variables can be replaced with ERB variables.

resources/windows/msi/service-filter.xslt.erb
melissa: Since this is service specific, would this be better to include in https://github.com/melissa/vanagon/blob/ticket/master/RE-6236-wix-services/resources/windows/wix/service.component.wxs.erb ? Or do we need to have a filter file? Will each service have an entry to add here? Or will it just be the service files, like ruby.exe, rubyw.exe, and maybe but maybe not nssm?

jcoconnor: This filter file is tied to the harvest operation - it lists the files that need to be screened out of the harvest operation.

melissa: But, what files need to be screened out of the harvest operation? How do we define them? Where do we get the list of files? Do we define them as part of the services? I assume from your comment we need a separate file to pass in to the harvest command

jcoconnor: This file should track the list of services, so if we add a service using a particular executable, this filter should automatically screen it out of the harvest.

jcoconnor:
I'm going to remove the proj.wxs.erb file completely from this PR as @jpogran is submitting this in PR #271

resources/windows/wix/service-filter.xslt.erb
```diff
@@ -0,0 +1,19 @@
...
+  <!--<xsl:strip-space elements="*"/>-->
+  <xsl:template match="@*|node()">
+    <xsl:copy>
+      <xsl:apply-templates select="@*|node()"/>
+    </xsl:copy>
+  </xsl:template>
+
+  <!-- Loop through the list of services and add filters for each component requiring Service type special treatment -->
+  <%- service_files.each do |service| -%>
+    <xsl:template match="wix:Component[wix:File[@Source='<%=  service.path %>']]" />
```
melissa: https://github.com/puppetlabs/puppet-agent/pull/519/files#diff-9ae79323449fab62c617785704463a77R76

jcoconnor: @melissa Have updated the xslt.erb file with what I think are the appropriate variables. Will probably have to fine tune this, but at the moment, the service filenames are making it into the template correctly, albeit, I'm not sure about the prefix.

resources/windows/wix/service-filter.xslt.erb
```diff
@@ -0,0 +1,21 @@
...
+    <xsl:copy>
+      <xsl:apply-templates select="@*|node()"/>
+    </xsl:copy>
+  </xsl:template>
+
+  <!-- Loop through the list of services and add filters for each component requiring Service type special treatment -->
+  <%- get_services.each do |service| -%>
+
+      <xsl:template match="wix:Component[wix:File[@Source='<%= service.options.first[:file][:source] %>"']]" />
+
```
melissa: Is this extra white space necessary? Also, I'm not sure about the comment. Maybe something with a little more clarity? Like, why we have to filter these? Don't worry about the prefix. From the conversation this morning, this will change.

jcoconnor: Will remove the white spaces.
The prefix is causing the compile to break, but at least its progressing that the xslt file is now arriving over, being picked up and used.

shermdog:
This is a lot of commits for such few changes overall.  Can you squash these together?

jcoconnor:
@shermdog Sqashed out some of the superfluous commits

lib/vanagon/platform/windows.rb
```diff
@@ -44,6 +47,16 @@ def generate_packaging_artifacts(workdir, name, binding)
         ["chocolateyInstall.ps1", "chocolateyUninstall.ps1"].each do |win_file|
           FileUtils.copy(File.join(VANAGON_ROOT, "resources/windows/nuget/#{win_file}"), File.join(workdir, win_file))
         end
+
+        # Now for the msi (Wix) artefacts
```
melissa: Spelling

lib/vanagon/platform/windows.rb
```diff
@@ -81,8 +94,9 @@ def generate_msi_package(project)
         dir_ref = "INSTALLDIR"
         # Actual array of commands to be written to the Makefile
         ["mkdir -p output/#{target_dir}",
-        "mkdir -p $(tempdir)/#{project.name}/staging",
-        "gunzip -c #{project.name}-#{project.version}.tar.gz | '#{@tar}' -C '$(tempdir)/#{project.name}/staging' --strip-components 1 -xf -",
+        "mkdir -p $(tempdir)/{staging,wix}",
+        "cp -v wix/* $(tempdir)/wix",
```
melissa: :+1:

melissa:
Needs a rebase

lib/vanagon/platform/windows.rb
melissa: Wait, you probably want `-r`, not `-v`

jcoconnor: The -v was deliberate - just wanted to log what we were copying.
But we will need an -r anyways - we can add that when we need to copy sub-directories (e.g. UI artefacts).

lib/vanagon/platform/windows.rb
jcoconnor: Lets compromise - keep the UK spelling for the comment and the US one for the function name :)

melissa: but it hurts my eyeballs...

resources/windows/wix/service-filter.xslt.erb
```diff
@@ -0,0 +1,18 @@
...
+  <xsl:output method="xml" indent="yes" />
+  <!--<xsl:strip-space elements="*"/>-->
+  <xsl:template match="@*|node()">
+    <xsl:copy>
+      <xsl:apply-templates select="@*|node()"/>
+    </xsl:copy>
+  </xsl:template>
+  <!-- Filter out all Service File Executables from the Harvest (heat) run as these are specified in the transformed service.components.wxs file -->
+  <%- get_services.each do |service| -%>
+      <xsl:template match="wix:Component[wix:File[@Source='<%= service.options.first[:file][:source] %>']]" />
```
melissa: We probably want to check for duplicate entries

jcoconnor: Can you give a use case of what you are thinking of here @melissa ?

The same check would also apply to the service.components file.
Suspect if we did have duplicates here, the wix build would complain (but not certain about this).

jcoconnor:
Have rebased this and dealt with other comments above - except for duplicate detection.

resources/windows/wix/service-filter.xslt.erb
melissa: If we have multiple entries of one source
```
stagedir\sys\ruby\bin\ruby.exe
stagedir\sys\ruby\bin\ruby.exe
stagedir\sys\ruby\bin\ruby.exe
stagedir\sys\ruby\bin\ruby.exe
stagedir\sys\ruby\bin\ruby.exe
```
we want to specify it only once.

This may happen if we have multiple services using the same source. I.e., the puppet services all use `stagedir\sys\ruby\bin\ruby.exe`. They all have unique identifiers, but if we are passing that path to this file, then we will have multiple entries of stagedir\sys\ruby\bin\ruby.exe. Or if we have different components that have services that use the same source. That isn't currently the case, but if mco used `stagedir\sys\ruby\bin\ruby.exe` rather than `stagedir\sys\ruby\bin\rubyw.exe` we would have duplicate sources.

Ruby has the ability to remove any duplicates from an array (http://docs.ruby-lang.org/en/2.1.0/Array.html#method-i-uniq). The method is called `uniq`.

Have you tested this with https://github.com/puppetlabs/puppet-agent/pull/519 ? If you have multiple service entries for one component, you're going to miss the additional ones. We have no guarentee that multiple services in a component will use the same source. In this current example, you're only taking the first service entry and ignoring the rest.

lib/vanagon/platform/windows.rb
```diff
@@ -32,6 +32,7 @@ def package_name(project)
       end
 
       # Method to generate the files required to build a windows package for the project
+      # Both the MSI and nuget generic articfacts are created for simplicity.
```
melissa: This comment isn't necessary. It restates what that code already shows and adds no additional information.

lib/vanagon/platform/windows.rb
```diff
@@ -55,6 +56,7 @@ def generate_packaging_artifacts(workdir, name, binding)
       def generate_msi_packaging_artifacts(workdir, name, binding)
         FileUtils.mkdir_p(File.join(workdir, "wix"))
         erb_file(File.join(VANAGON_ROOT, "resources/windows/wix/service.component.wxs.erb"), File.join(workdir, "wix", "service.#{name}.wxs"), false, { :binding => binding })
+        erb_file(File.join(VANAGON_ROOT, "resources/windows/wix/service-filter.xslt.erb"), File.join(workdir, "wix", "#{name}.service-filter.xslt"), false, { :binding => binding })
```
melissa: If we're renaming this file, we should change it from `project.service-filter.xslt.erb` to `#{name}.service-filter.xslt`. Rather than adding a new value to the name, we should be replacing a value.

lib/vanagon/platform/windows.rb
```diff
@@ -55,6 +56,7 @@ def generate_packaging_artifacts(workdir, name, binding)
       def generate_msi_packaging_artifacts(workdir, name, binding)
         FileUtils.mkdir_p(File.join(workdir, "wix"))
         erb_file(File.join(VANAGON_ROOT, "resources/windows/wix/service.component.wxs.erb"), File.join(workdir, "wix", "service.#{name}.wxs"), false, { :binding => binding })
```
melissa: Not for this PR, but I should have named this file `service.project.wxs.erb` since it contains all the services for the project, not services for individual components.

lib/vanagon/platform/windows.rb
```diff
@@ -107,8 +109,9 @@ def generate_msi_package(project)
         dir_ref = "INSTALLDIR"
         # Actual array of commands to be written to the Makefile
         ["mkdir -p output/#{target_dir}",
-        "mkdir -p $(tempdir)/#{project.name}/staging",
```
melissa: Since this change is made in a separate PR, it should be removed from this one.

lib/vanagon/platform/windows.rb
```diff
@@ -107,8 +109,9 @@ def generate_msi_package(project)
...
-        "gunzip -c #{project.name}-#{project.version}.tar.gz | '#{@tar}' -C '$(tempdir)/#{project.name}/staging' --strip-components 1 -xf -",
+        "mkdir -p $(tempdir)/{staging,wix}",
+        "cp -v -r wix/* $(tempdir)/wix",
```
melissa: You should be able to use `-vr`

Also, you should use `#{@copy}`

Also, I'd rather not have `-v` here because it adds verbose output we don't need. It's fine for debugging purposes. I don't want it in production.

resources/windows/wix/service-filter.xslt.erb
melissa: I disagree that the same check would apply to the services.wxs file. The user should be specifying unique id's for each service that uses the same source.

jcoconnor: Thanks for the use case. We should certainly check and filter out multiple components - will apply a fix for this shortly.

However, the case of multiple services using the same source is quite complex as you can only install a single instance of a source. See [How to install multiple windows services with one executable](http://stackoverflow.com/questions/9116870/wix-how-to-install-multiple-windows-services-with-one-executable) & [Using Wix to install process with multiple windows services](http://stackoverflow.com/questions/19414695/using-wix-to-install-process-with-multiple-windows-services).

lib/vanagon/platform/windows.rb
jcoconnor: Agreed - can do this here if you wish.

lib/vanagon/platform/windows.rb
jcoconnor: Good point - done.

lib/vanagon/platform/windows.rb
jcoconnor: Ok - my preference would be to leave the -v in as it doesn't add a lot extra, but since we can easily check the result of the copy command I will remove it.

Did think about using -vr, but since the heat/light commands have a lot -xx -yy type directives (which are odd and not gnu-ish), I though separate options here were better.

lib/vanagon/platform/windows.rb
jcoconnor: I think it appear here due to the rebasing I did when #270 was merged and the use of the ```"mkdir -p $(tempdir)/{staging,wix}"``` construct I have finally used - will see if I can make it clearer.

jcoconnor: Have cherry-picked your change 620db16b2b5414fdd008784de4ffc9bcea35a187 into the head of my changelist.
Notice that your change hasn't been merged up yet though.

lib/vanagon/platform/windows.rb
jcoconnor: LOL - I Think its disappeared anyway...

lib/vanagon/platform/windows.rb
jcoconnor: Should be gone now.

lib/vanagon/platform/windows.rb
melissa: Naw, let's ignore this for now. I'll change it later when it bothers me too much

resources/windows/wix/service-filter.xslt.erb
jcoconnor: Done - Unhappy with the extra lines in the transformed files, so will look at that.

lib/vanagon/platform/windows.rb
```diff
@@ -55,6 +55,7 @@ def generate_packaging_artifacts(workdir, name, binding)
       def generate_msi_packaging_artifacts(workdir, name, binding)
         FileUtils.mkdir_p(File.join(workdir, "wix"))
         erb_file(File.join(VANAGON_ROOT, "resources/windows/wix/service.component.wxs.erb"), File.join(workdir, "wix", "service.#{name}.wxs"), false, { :binding => binding })
+        erb_file(File.join(VANAGON_ROOT, "resources/windows/wix/project.service-filter.xslt.erb"), File.join(workdir, "wix", "#{name}.service-filter.xslt"), false, { :binding => binding })
```
melissa: Did we decide to leave this as `filter` rather then `service-filter` to deal with the potential we may at some point need to filter things not associated with the services?

jcoconnor: Ok - will change - push coming up in a few mins.

resources/windows/wix/project.service-filter.xslt.erb
```diff
@@ -0,0 +1,23 @@
...
+  <xsl:template match="@*|node()">
+    <xsl:copy>
+      <xsl:apply-templates select="@*|node()"/>
+    </xsl:copy>
+  </xsl:template>
+  <!-- Filter out all Service File Executables from the Harvest (heat) run as these are specified in the transformed service.components.wxs file -->
+  <!-- The list of component service files are collected in an array so that a set of unique names can be extracted -->
+  <%- service_files = Array.new -%>
+  <%- get_services.each do |service| -%>
+    <%- service_files << service.options.first[:file][:source] -%>
```
melissa: Wait, don't we have to iterate through the different definitions still?

jcoconnor: Ugh - yes - you are right :(

lib/vanagon/platform/windows.rb
```diff
@@ -107,7 +108,8 @@ def generate_msi_package(project)
         dir_ref = "INSTALLDIR"
         # Actual array of commands to be written to the Makefile
         ["mkdir -p output/#{target_dir}",
-        "mkdir -p $(tempdir)/staging",
+        "mkdir -p $(tempdir)/{staging,wix}",
+        "#{@copy} -r wix/* $(tempdir)/wix",
```
melissa: Wanna add an extra `/` after `$(tempdir)/wix` so we're extra clear we're moving stuff into a directory? It doesn't impact functionality at all, since this already does the right thing. It would just add extra clarity

resources/windows/wix/project.service-filter.xslt.erb
jcoconnor: Fixed

lib/vanagon/platform/windows.rb
jcoconnor: Done

lib/vanagon/platform/windows.rb
jcoconnor: Done

melissa:
:+1: this looks ready to me

melissa:
shoot, those commits needed to get squashed. I didn't notice that until after the thumbs up

McdonaldSeanp:
Woah, I didn't even notice, whooops

melissa:
Oh well. @jcoconnor can you have small changes like this in single commits next time?

jcoconnor:
Ok @melissa @McdonaldSeanp will do that in future.

PR #270, opened 2016-01-28
(RE-6236) Add wix service template

melissa:


resources/windows/wix/service.component.wxs.erb
```diff
@@ -0,0 +1,58 @@
+<?xml version='1.0' encoding='windows-1252'?>
+<Wix xmlns='http://schemas.microsoft.com/wix/2006/wi' xmlns:util="http://schemas.microsoft.com/wix/UtilExtension">
+  <Fragment>
+  <%- get_services.each do |service| -%>
+    <%- service.options.each do |component| -%>
+    <DirectoryRef Id="<%= component[:directory_ref] %>">
+      <Component Id='<%= component[:id] %>' Guid="<%= component[:guid] %>" <%= "Win64=\"$(var.Win64)\"" unless component[:win64].nil? %>>
+        <%- if component[:file][:id] == "NSSM" -%>
+        <CreateFolder />
```
melissa: Also, I have no idea why this is getting called for the pxp-agent service, but not for the others. The conditional is rather arbitrary, but it'd be nice to have something more concrete to define if we add this line or not

jpogran: CreateFolder is only added if we expect the folder to not have contents at install time, but later on. MSI won't create an empty folder without it. Will need to look at original hierarchy to decide if it is really necessary.

resources/windows/wix/service.component.wxs.erb
```diff
@@ -0,0 +1,58 @@
+<?xml version='1.0' encoding='windows-1252'?>
+<Wix xmlns='http://schemas.microsoft.com/wix/2006/wi' xmlns:util="http://schemas.microsoft.com/wix/UtilExtension">
+  <Fragment>
+  <%- get_services.each do |service| -%>
+    <%- service.options.each do |component| -%>
+    <DirectoryRef Id="<%= component[:directory_ref] %>">
+      <Component Id='<%= component[:id] %>' Guid="<%= component[:guid] %>" <%= "Win64=\"$(var.Win64)\"" unless component[:win64].nil? %>>
+        <%- if component[:file][:id] == "NSSM" -%>
```
jcoconnor: We shouldn't be putting in service specific code in here.
So if a <CreateFolder> is needed (bearing in mind James' comment below), we should do that with a specific tag like :create_folder => "true".

melissa: I'm going to wait until I hear back from James so we know why we're creating the folder.

jcoconnor: @melissa As an approach, this looks very good.
I still need to try it in practice against build so that I can ascertain the the syntax is correct.
(unfortunately there was rather too much disruption today to do it).
Will do this first thing on Monday.

In the meantime would value comments from @Iristyle @ferventcoder and @MikaelSmith

resources/windows/wix/service.component.wxs.erb
```diff
@@ -0,0 +1,58 @@
+<?xml version='1.0' encoding='windows-1252'?>
+<Wix xmlns='http://schemas.microsoft.com/wix/2006/wi' xmlns:util="http://schemas.microsoft.com/wix/UtilExtension">
+  <Fragment>
+  <%- get_services.each do |service| -%>
+    <%- service.options.each do |component| -%>
+    <DirectoryRef Id="<%= component[:directory_ref] %>">
+      <Component Id='<%= component[:id] %>' Guid="<%= component[:guid] %>" <%= "Win64=\"#{component[:win64]}\"" unless component[:win64].nil? %>>
+        <%- if component[:file][:id] == "NSSM" -%>
+        <CreateFolder />
```
melissa: We should still figure out a better way to decide if we need to create the directory or not

cc/ @jpogran @jcoconnor

melissa:
cc/ @ferventcoder @Iristyle @glennsarti @jpogran @jcoconnor @shrug @McdonaldSeanp @mckern @underscorgan could you review this?

lib/vanagon/platform/windows.rb
```diff
@@ -44,6 +44,9 @@ def generate_packaging_artifacts(workdir, name, binding)
         ["chocolateyInstall.ps1", "chocolateyUninstall.ps1"].each do |win_file|
           FileUtils.copy(File.join(VANAGON_ROOT, "resources/windows/nuget/#{win_file}"), File.join(workdir, win_file))
         end
+
+        # MSI artifacts
+        erb_file(File.join(VANAGON_ROOT, "resources/windows/wix/service.component.wxs.erb"), File.join(workdir, "service.#{name}.wxs"), false, { :binding => binding })
```
MikaelSmith: Why are we generating nuget and wix files at the same time?

melissa: @MikaelSmith we don't have access to the package_type at this moment in time. We could finagle it out of the binding that gets passed in, but we figured it wouldn't do any hard to generate both. The only thing that might be an issue would be missing variables... but I'm not sure if that would even throw an issue unless we explicitely said to

resources/windows/wix/service.component.wxs.erb
jcoconnor: @melissa Holding directory needs to be changes to resources/windows/wix

jcoconnor:
@melissa It won't be possible to test these definitively until we have at least a basic set of candle/light commands in the makefile. So - should we hold off submission of this until we can verify it, or push these through subject to review with a ticket later to verify their "Wix" correctness ?

melissa:
@jcoconnor can you review this manually then? Can you generate the files and review the contents manually?

jcoconnor:
Yes - was going to try doing a comparison some extracts from the current pxp-agent build tomorrow.

lib/vanagon/component/dsl.rb
```diff
@@ -184,7 +185,7 @@ def install_service(service_file, default_file = nil, service_name = @component.
           # Return here because there is no file to install, just a string read in
           return
         when "windows"
-          # Placeholder for Windows
+          @component.service = OpenStruct.new(:name => service_name, :options => service_hash)
```
mckern: New-style declarations?

```ruby
OpenStruct.new(name: service_name, options: service_hash)
```

resources/windows/wix/service.component.wxs.erb
melissa: @jcoconnor are you telling me this isn't a thing we need in this file? What are you saying with that comment?

jcoconnor: Ugh - Either I was talking through my butt, or you have already changed the directory, i.e. the file should sit in: resources/windows/**wix**/service.component.wxs.erb

melissa: Oohh, that's what you were referring to. This file has always been in `resources/windows/wix`

jcoconnor: note to self - make clearer comments
I see github has a bit of an upgrade and markdown options are available above - cool.

melissa: So, under which circumstances do we call `<CreateFolder />`? cc/ @jcoconnor @jpogran

lib/vanagon/platform/windows.rb
```diff
@@ -37,12 +37,19 @@ def package_name(project)
       # @param name [String] name of the project
       # @param binding [Binding] binding to use in evaluating the packaging templates
       def generate_packaging_artifacts(workdir, name, binding)
-        # nuget templates that do require a name change
-        erb_file(File.join(VANAGON_ROOT, "resources/windows/nuget/project.nuspec.erb"), File.join(workdir, "#{name}.nuspec"), false, { :binding => binding })
+        if @package_type == "nuget"
```
melissa: @jcoconnor I figured it out

lib/vanagon/component/dsl.rb
melissa: done

lib/vanagon/platform/windows.rb
melissa: Okay, I looked into it more. It's actually way more do-able than I thought.

lib/vanagon/platform/windows.rb
MikaelSmith: Ah; I was willing to accept the other answer, but this looks good.

melissa: :) I figured it was worth another look. We're adding a lot of wix specific variables, which would suck if we had to fail building nuget packages for all of them

jcoconnor: :+1: Cool

lib/vanagon/platform/windows.rb
```diff
@@ -37,12 +37,18 @@ def package_name(project)
...
-        ["chocolateyInstall.ps1", "chocolateyUninstall.ps1"].each do |win_file|
-          FileUtils.copy(File.join(VANAGON_ROOT, "resources/windows/nuget/#{win_file}"), File.join(workdir, win_file))
+          # nuget static resources to be copied into place
+          ["chocolateyInstall.ps1", "chocolateyUninstall.ps1"].each do |win_file|
+            FileUtils.copy(File.join(VANAGON_ROOT, "resources/windows/nuget/#{win_file}"), File.join(workdir, win_file))
+          end
+        end
+
+        if @package_type == "msi"
+          erb_file(File.join(VANAGON_ROOT, "resources/windows/wix/service.component.wxs.erb"), File.join(workdir, "service.#{name}.wxs"), false, { :binding => binding })
```
jcoconnor: @melissa Can you change ```"service.#{name}.wxs"``` to ```"wix/service.#{name}.wxs"```
This means that all the Wix files are placed in a "wix" sub-directory allowing a single copy command to move them to the ```$(tempdir)/wix``` directory.

jcoconnor:
@melissa I have compared the produced ```service.puppet-agent.wxs``` that gets dropped on the build machine.
The transformation is 100% correct for the puppet-agent services.
There are some minor issues with the registry entries for pxp-agent and the arguments for the MCO Service, but these are issues with the actual data coming in from their respective component files.
I will put comments on those individual lines in the [Puppet-Agent PR](https://github.com/puppetlabs/puppet-agent/pull/519) 

So :+1: LGTM

resources/windows/wix/service.component.wxs.erb
jpogran: CreateFolder is only used if the folder has no contents in it, otherwise MSI would not create the folder on install, so I am not sure why it is here, when we know that the executable will be. In order to ensure we can remove it, we have to examine the directory specified in line 6 and see where that is located in our wxs files. If it is present in our wxs with files associated with it, then we can remove createfolder

McdonaldSeanp:
I'm fully :+1:

lib/vanagon/platform/windows.rb
melissa: What?!?!?! That puts the wix files into a wix subdirectory?!?! I'm shocked! That's pure magic. I had no idea that would happen.

On a serious note, I agree this will make it easier to ensure we copy all the wix files into the correct location.

lib/vanagon/platform/windows.rb
melissa: @jcoconnor good point. Done.

lib/vanagon/platform/windows.rb
melissa: Done.

jpogran:
:+1:

melissa:
@jpogran @jcoconnor the one thing I want to check on is missing variables. We should go back and make sure the user sets everything that's required for these, and fail explicitely if they're missing.

resources/windows/wix/service.component.wxs.erb
```diff
@@ -0,0 +1,55 @@
...
+          <%- component[:conditions].each do |condition| -%>
+        <Condition>
+          <%= condition %>
+        </Condition>
+          <%- end -%>
+        <%- end -%>
+        <File Id="<%= component[:file][:id] %>" KeyPath="yes" Source="<%= component[:file][:source] %>" />
+        <ServiceInstall Id="<%= component[:serviceinstall][:id] %>"
+          Account="[PUPPET_AGENT_ACCOUNT_DOMAIN]\[PUPPET_AGENT_ACCOUNT_USER]"
+          Password="[PUPPET_AGENT_ACCOUNT_PASSWORD]"
```
ferventcoder: A little late to the conversation, but most of this looks good and I wouldn't consider it an abuse of Ruby at all.

These particular bits though, I can possibly foresee folks using different accounts for different services and would want to move to maybe this being more open to that.

resources/windows/wix/service.component.wxs.erb
```diff
@@ -0,0 +1,55 @@
+<?xml version='1.0' encoding='windows-1252'?>
+<Wix xmlns='http://schemas.microsoft.com/wix/2006/wi' xmlns:util="http://schemas.microsoft.com/wix/UtilExtension">
+  <Fragment>
+  <%- get_services.each do |service| -%>
+    <%- service.options.each do |component| -%>
```
ferventcoder: I think I know what this is doing, but I would love to see an example of this

ferventcoder: I'm thinking this is accounting for the manual startup, the disabled startup and the automatic startup for EACH service.

melissa: Yes, it's taking into account the multiple service types provided for puppet. Each service type is a component of the main service argument.

resources/windows/wix/service.component.wxs.erb
melissa: yes, this should be changed.

PR #271, opened 2016-01-29
(WIP) (RE-6236) Add wix main template

jpogran:
Provide a main WiX template that will be the entry point for creating MSI files. This file will contain

- Product and Package elements
- Top level Feature and Directory elements
- Any properties and custom actions for the Remembered Installation directory pattern

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,27 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 This changelog adheres to [Keep a CHANGELOG](http://keepachangelog.com/).
 
 ## [Unreleased]
-## [0.5.0] - 2015-12-22
+
+## [0.5.1] - 2016-2-04
+### Changed
+- Vanagon now assumes anything that is not listed as an archive is just
+  a file, and thus NON-ARCHIVE-EXTENSIONS no longer needs to be updated
+  for new extentions.
+
+### Fixed
+- Builds will halt if any git command fails
+- Hardware engine will release locks on failures
+- Builds no longer fail if things items were in /usr/local on debian
+- on RPM builds, if there are no directories to package builds no longer
+  fail
+- Repeated builds no longer fail due to a symlink issue
+
+### Added
+- Added functionality to enable SSH forwarding in to vanagon pooler hosts
+- Added functionality to build nuget archives
+- Added support for patching gems
+
+## [0.5.0] - 2016-1-06
 ### Changed
 - Replaced the ERB template based Makefile generation with a procedural
   Makefile backend
@@ -149,7 +169,8 @@ This changelog adheres to [Keep a CHANGELOG](http://keepachangelog.com/).
 
 ## Versions <= 0.3.9 do not have a change log entry
 
-[Unreleased]: https://github.com/puppetlabs/vanagon/compare/0.5.0...HEAD
+[Unreleased]: https://github.com/puppetlabs/vanagon/compare/0.5.1...HEAD
+[0.5.1]: https://github.com/puppetlabs/vanagon/compare/0.5.0...0.5.1
 [0.5.0]: https://github.com/puppetlabs/vanagon/compare/0.4.1...0.5.0
 [0.4.1]: https://github.com/puppetlabs/vanagon/compare/0.4.0...0.4.1
 [0.4.0]: https://github.com/puppetlabs/vanagon/compare/0.3.19...0.4.0


melissa:
We'll also need to process the template in https://github.com/puppetlabs/vanagon/blob/master/lib/vanagon/platform/windows.rb#L41

I suggest something like
```
# MSI artifacts
erb_file(File.join(VANAGON_ROOT, "resources/windows/wix/project.wxs.erb"), File.join(workdir, "#{name}.wxs"), false, { :binding => binding })
```
I also suggest, just because this is what is done for templates for other platforms, that we name this file `project.wxs.erb` and rename it when we process the template to `puppet-agent.wxs` (or whatever the project name we're building is).

jpogran:
No problem with changing name of file, so long as its not puppet. ;)

Also, I noticed today that i have `resources/windows/wix` because I branched off your PR, but PR #269 has `resources/windows/msi` instead. I don't know if we chose a folder name. This isn't a technical choice, so we should choose the name that works best to describe the thing that we are doing. We are working with wix stuff, but msi is our output. I lean toward wix.

melissa:
I'm inclined to go with wix, but my opinion isn't very strong

resources/windows/wix/main.wxs.erb
```diff
@@ -0,0 +1,91 @@
...
+    <Media Id="1" Cabinet="<%= Product %>.cab" EmbedCab="yes" CompressionLevel="high" />
+
+    <Feature Id="<%= ProductName %>Runtime" Title="<%= OurProductName %> Runtime" Level="1">
+      <ComponentRef Id="ProductComponentGroup" />
+    </Feature>
+
+    <Directory Id='TARGETDIR' Name='SourceDir'>
+      <Directory Id="<%= PlatformProgramFilesFolder %>" >
+        <Directory Id='PuppetLabs' Name="<%= CompanyName %>">
+          <Directory Id='INSTALLDIR' Name="<%= OurProductName %>"/>
```
melissa: What about if different projects aren't installing here? Or have a different number of nested directories for their `INSTALLDIR`?

resources/windows/wix/project.wxs.erb
```diff
@@ -0,0 +1,70 @@
+<?xml version="1.0" encoding="windows-1252"?>
+  <Wix xmlns="http://schemas.microsoft.com/wix/2006/wi" xmlns:util="http://schemas.microsoft.com/wix/UtilExtension">
+    <Product
+      Id="*"
+      UpgradeCode="<%= settings[:upgrade_code] %>"
+      Name="<%= settings[:product_name] %>"
+      Language="1033"
+      Codepage="1252"
+      Version="<%= @version.sub(/\.g[0-9a-z]{7}/, '') %>"
```
melissa: Oops, this isn't quite safe yet. It should be something closer to `/\.g[0-9a-z]{7}$/` to ensure we're only taking out the last bit. It should be fine, but it's an easy thing to change to prevent possible issues later on

lib/vanagon/platform/windows.rb
```diff
@@ -82,6 +85,8 @@ def generate_msi_package(project)
         # Actual array of commands to be written to the Makefile
         ["mkdir -p output/#{target_dir}",
         "mkdir -p $(tempdir)/#{project.name}/staging",
+        "mkdir -p $(tempdir)/wix",
+        "#{@copy} *.wxs $(tempdir)/wix",
```
jcoconnor: @jpogran and @melissa Just need to be careful on conflicts on these lines as all 3 of us have made slightly different changes (e.g. Melissa has dropped the project.name from the staging tree and I think we might a cp -r to copy the UI sub-directories (with icons/gifs)) over.

However - we can probably deal with this once we get the first set of PR's merged.

resources/windows/wix/project.wxs.erb
```diff
@@ -0,0 +1,70 @@
...
+      Comments="<%= @homepage %>"
+      Compressed="yes"
+      Platform="<%= @platform.architecture %>"
+    />
+    <MajorUpgrade AllowDowngrades="yes" />
+    <Media Id="1" Cabinet="<%= settings[:product_word] %>.cab" EmbedCab="yes" CompressionLevel="high" />
+
+    <Feature Id="<%= settings[:product_word] %>Runtime" Title="<%= settings[:product_word] %> Runtime" Level="1">
+      <!-- All heat runs and WiX Fragment files containing Component elements will reference this ComponentGroup -->
+      <ComponentGroup Id="ProductComponentGroup" />
```
jcoconnor: Are we sticking with a standard ProductComponentGroup or should we include project name or similar here.

resources/windows/wix/project.wxs.erb
```diff
@@ -0,0 +1,70 @@
...
+    </Feature>
+
+    <!-- We will use DirectoryRef at the project level to hook in the project directory structure -->
+    <Directory Id='TARGETDIR' Name='SourceDir' >
```
melissa: Is it possible that different projects are going to want to set this as a different name? So, if a project wanted `Name='sourcedir'` or `Name='builddir'` or something?

I'm thinking we could have `proj.setting(:targetdir, "SourceDir")`, which we could then use to build the directory structure up.

```
proj.setting(:company_name, "Puppet Labs")
proj.setting(:company_id, "PuppetLabs")
proj.setting(:product_name, "Puppet Agent (64-bit)")
proj.setting(:product_id, "Puppet")

proj.setting(:targetdir, "SourceDir")
proj.setting(:base_dir, File.join(proj.targetdir, "ProgramFilesFolder")
proj.setting(:install_root, File.join(proj.base_dir, proj.company_name)
proj.setting(:prefix, File.join(proj.install_root, proj.product_id)

proj.directory proj.base_dir
proj.directory proj.install_root, proj.company_id
proj.directory proj.prefix, "INSTALLDIR"
```

And then we do all the installation in a directory like `SourceDir/ProgramFilesFolder/Puppet Labs/Puppet` (ruby, openssl, facter, curl, puppet, etc).

And then @McdonaldSeanp does the directory work. `Name=<%= File.basename(directory.path) %>` and `Id=<%= directory.id %>` or something.

and we get all the directories we specify (notice I didn't include :targetdir in the list of directory declarations) which process all the subdirectories. In `project.wxs.erb`, we have this line be `<Directory Id='TARGETDIR' Name='<%= settings[:targetdir] %>' >`, and then in the directory erb we know the ID will be `TARGETDIR` and everything will be nested under it. Then things like `ruby.exe` will have a path set to  `SourceDir/ProgramFilesFolder/Puppet Labs/Puppet/bin/ruby.exe` which might make @jcoconnor's filter stuff work.

That still leaves the problem that vanagon is building packages with a space in the path name (https://github.com/puppetlabs/puppet-agent/blob/master/configs/components/openssl.rb#L81) and the possibility that some of these packages will hardcode the path into how they link, etc. But I need someone else to think about that problem for the moment.

lib/vanagon/platform/windows.rb
melissa: yeah, I'd suggest not worrying about that specific case until we implement it. We might think we're going to solve it one way, and then end up solving it a completely different way

resources/windows/wix/project.wxs.erb
jpogran: Changed

resources/windows/wix/project.wxs.erb
jpogran: IMHO we should stick with this. This has to be passed to the heat run as well as used in every other component.

An alternative is to remove this line entirely, and have some other file being responsible for hooking into `Feature`

jcoconnor: OK - lets run with it for the moment.

lib/vanagon/platform/windows.rb
```diff
@@ -70,6 +70,9 @@ def generate_nuget_packaging_artifacts(workdir, name, binding)
         ["chocolateyInstall.ps1", "chocolateyUninstall.ps1"].each do |win_file|
           FileUtils.copy(File.join(VANAGON_ROOT, "resources/windows/nuget/#{win_file}"), File.join(workdir, win_file))
         end
+
+        # MSI artifacts
+        erb_file(File.join(VANAGON_ROOT, "resources/windows/wix/project.wxs.erb"), File.join(workdir, "#{name}.wxs"), false, { :binding => binding })
```
melissa: You're going to want to move this up under line 56 :)

melissa: Also, the second argument should be `File.join(workdir, "wix", "#{name}.wxs")` now that we're utilizing a wix subdirectory

lib/vanagon/platform/windows.rb
melissa: ` "#{@copy} -r wix $(tempdir)/wix",`

resources/windows/wix/project.wxs.erb
melissa: Does this mean https://github.com/jpogran/vanagon/blob/ticket/master/RE-6236-wix-services/lib/vanagon/platform/windows.rb#L109 should be changed to correspond with this variable? Or, if not that, something else in the heat invocation?

melissa:
https://github.com/puppetlabs/puppet-agent/pull/523 provides a lot of the variables referenced here (for those of you reviewing this)

melissa:
@jpogran could you squash the first three commits together (including the one I provided)?

jpogran:
Done. I had wanted to do that, but didnt want to 'erase' your commit

melissa:
:+1: I'm happy with this

jcoconnor:
Just needs a rebase - will :+1: when done

jpogran:
Yah. Still working through that. windows.rb was modified and now there are conflicts that I have to resolve

jpogran:
@melissa @jcoconnor rebased

lib/vanagon/platform/windows.rb
```diff
@@ -103,12 +104,14 @@ def generate_nuget_package(project)
...
       def generate_msi_package(project)
         target_dir = project.repo ? output_dir(project.repo) : output_dir
-        cg_name = "compfiles"
+        cg_name = "ProductComponentGroup"
         dir_ref = "INSTALLDIR"
         # Actual array of commands to be written to the Makefile
         ["mkdir -p output/#{target_dir}",
-        "mkdir -p $(tempdir)/staging",
-        "gunzip -c #{project.name}-#{project.version}.tar.gz | '#{@tar}' -C '$(tempdir)/staging' --strip-components 1 -xf -",
+        "mkdir -p $(tempdir)/#{project.name}/staging",
```
melissa: It looks like you lost the changes merged in with https://github.com/puppetlabs/vanagon/pull/273!

jpogran: sigh. fixing.

melissa:
:+1: I'm into it. I think this is ready to merge pending travis

resources/windows/wix/project.wxs.erb
```diff
@@ -0,0 +1,70 @@
...
+      <Component Id="RegistryEntriesArchitectureDependent" Guid="E6D5AF4F-ACC4-4D11-AFCE-299A9ED2152C" Win64="<%= settings[:win64] %>" Permanent="yes">
```
melissa: Is this guid going to be the same for all projects? I.e., puppet-agent and also pe-client-tools?

jcoconnor: James - I would have thought the GUID for this should be specific to the product, i.e. template or wix variable ?

Also - looks like you need to rebase, but Travis is fine.

lib/vanagon/platform/windows.rb
```diff
@@ -103,7 +104,7 @@ def generate_nuget_package(project)
       # @return [Array] list of commands required to build an msi package for the given project from a tarball
       def generate_msi_package(project)
         target_dir = project.repo ? output_dir(project.repo) : output_dir
-        cg_name = "compfiles"
+        cg_name = "ProductComponentGroup"
```
jcoconnor: Just thinking aloud here, should this be a variable that is also used in the project template ? (see https://github.com/puppetlabs/vanagon/pull/271/files#r51787764). The downside is some extra complexity and that we would need to define the variable in the project.rb file.

resources/windows/wix/project.wxs.erb
```diff
@@ -0,0 +1,70 @@
...
+  <Wix xmlns="http://schemas.microsoft.com/wix/2006/wi" xmlns:util="http://schemas.microsoft.com/wix/UtilExtension">
+    <Product
+      Id="*"
+      UpgradeCode="<%= settings[:upgrade_code] %>"
+      Name="<%= settings[:product_name] %>"
+      Language="1033"
+      Codepage="1252"
+      Version="<%= @version.sub(/\.g[0-9a-z]{7}$/, '') %>"
+      Manufacturer="<%= settings[:company_name] %>"
+    />
```
jcoconnor: You have closed the ```<Product>``` element here, but you also have a ```</Product>``` at the end of the file.
Gives this error when I try to compile it.
```
puppet-agent.wxs
C:\cygwin64\var\tmp\tmp.avLTLHrNSr\wix\puppet-agent.wxs(61) : error CNDL0104 : Not a valid source file; detail: The 'Wix' start tag on line 2 position 4 does not match the end tag of 'Product'. Line 61, position 5.
Makefile:10: recipe for target 'puppet-agent-1.3.2.389.gaa23776.1-x64.msi' failed
```

lib/vanagon/platform/windows.rb
melissa: If we're only using it in two places, and it doesn't matter if multiple msis installed on the same system have this set the same, then I vote we keep it hard coded.

resources/windows/wix/project.wxs.erb
```diff
@@ -0,0 +1,73 @@
+<?xml version="1.0" encoding="windows-1252"?>
+  <Wix xmlns="http://schemas.microsoft.com/wix/2006/wi" xmlns:util="http://schemas.microsoft.com/wix/UtilExtension">
+    <Product
+      Id="*"
+      UpgradeCode="<%= settings[:upgrade_code] %>"
+      Name="<%= settings[:product_name] %>"
+      Language="1033"
+      Codepage="1252"
+      Version="<%= @version.sub(/\.g[0-9a-z]{7}$/, '') %>"
```
mckern: Maybe not in this PR but we should add a method that will return this value formatted already so we can pull the `sub` call out of this template.

resources/windows/wix/project.wxs.erb
```diff
@@ -0,0 +1,73 @@
...
+      Manufacturer="<%= settings[:company_name] %>"
+    >
+    <Package
+      InstallerVersion="300"
+      InstallScope="perMachine"
+      Description="<%= "#{settings[:product_word]}#{@platform.architecture == "x64" ? " (64-bit)" : ""}" %> Installer"
```
mckern: See previous comment about moving from inline `sub` to a helper function; inline ternary operator falls into the same bucket list.

(Non-blocking nitpick)

mckern:
I have no blocking concerns beyond "please rebase." A couple of nitpicks from the ERB, but we've got those anywhere we use ERB so that's nothing to worry about. :+1: pending rebase & travis re-run.

resources/windows/wix/project.wxs.erb
```diff
@@ -0,0 +1,73 @@
...
+    <!-- We can add all components by referencing this one thing -->
+      <ComponentGroupRef Id="ProductComponentGroup" />
+    </Feature>
+
+    <!-- All heat runs and WiX Fragment files containing Component elements will reference this ComponentGroup -->
+    <ComponentGroup Id="ProductComponentGroup" Directory="INSTALLDIR"/>
+
+    <!-- We will use DirectoryRef at the project level to hook in the project directory structure -->
+    <Directory Id='TARGETDIR' Name='SourceDir' >
+      <Component Id="RegistryEntriesArchitectureDependent" Guid="E6D5AF4F-ACC4-4D11-AFCE-299A9ED2152C" Win64="<%= settings[:win64] %>" Permanent="yes">
```
melissa: Wait, sorry, did we ever come to a conclusion on this guid?

lib/vanagon/platform/windows.rb
jcoconnor: Np

resources/windows/wix/project.wxs.erb
jcoconnor: I think it's safer to put the GUID outside of here.
Just concerned that two products installing with the same GUID value here might cause issues.

resources/windows/wix/project.wxs.erb
```diff
@@ -0,0 +1,73 @@
...
+        <RegistryKey Root="HKLM" Key="SOFTWARE\<%= settings[:company_name] %>\<%= settings[:product_name] %>" >
```
jcoconnor: This section is still causing a compile error:
```
C:\cygwin64\var\tmp\tmp.eB9RDfCbwk\wix\puppet-agent.wxs(34) : error CNDL0042 : The RegistryValue element has multiple key paths set.  
The key path may only be set to 'yes' in extension elements that support it or one of the following locations: RegistryValue/@KeyPath, File/@KeyPath, RegistryValue/@KeyPath, or ODBCDataSource/@KeyPath.
```

jpogran: This was due to the attempt to inline the if/else here. Some duplication is necessary to get this to work.

PR #274, opened 2016-02-02
(maint) Hardware engine should release lock upon failure

stahnma:
Prior to this, the hardware engine attempted to release the lock it had
on a system when regardless of the outcome of the vanagon build.
However, due to an if statement that could never evaluate to true, that
wasn't happening.

In debugging the problem, there was no way to figure out which engine
was being invoked. This commit adds a @name attribute to all engines so
other parts of the codebase can know what engine is being invoked.

This now means that if the hardware engine is being used, the ensure
block in driver.rb will release the lock regardless of pass/fail on the
build.

### DIFF
--- a/lib/vanagon/driver.rb
+++ b/lib/vanagon/driver.rb
@@ -92,7 +92,7 @@ class Vanagon
       puts e.backtrace.join("\n")
       raise e
     ensure
-      if @engine == "hardware"
+      if @engine.name == "hardware"
         @engine.teardown
       end
   end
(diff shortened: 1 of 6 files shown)


shrug:
:+1:

MikaelSmith:
This kind of makes me sad.

lib/vanagon/driver.rb
```diff
@@ -92,7 +92,7 @@ def run
       puts e.backtrace.join("\n")
       raise e
     ensure
-      if @engine == "hardware"
+      if @engine.name == "hardware"
```
MikaelSmith: This could've been @engine.is_a? Vanagon::Engine::Hardware, or a new method teardown_on_failure that no-ops for everything except Hardware. Adding a redundant name makes me sad.

PR #275, opened 2016-02-03
(CPR-290) Only create directory if path doesn't exist

MikaelSmith:
Previously if the destination existed when trying to create an empty
directory, installation would fail. Now installation will skip creating
the directory if it already exists, but will still fail if a
non-directory exists at that path.


MikaelSmith:
ping @melissa

melissa:
:+1: this seems great to me. Does it do what you need it to do?

MikaelSmith:
I believe so, I tested it out.

stahnma:
@MikaelSmith rebase and we can merge.

MikaelSmith:
Rebased.

PR #276, opened 2016-02-04
(RE-6476) Add option to enable SSH agent forwarding

heathseals:
This commit adds support for enabling SSH agent forwarding by checking
for the existence of an environment variable, VANAGON_SSH_AGENT.  If
set, Vanagon will construct the ssh command with the option to forward
the authentication agent connection.


shermdog:
:+1:  I had included this in my ssh config.  This is smarter :)

PR #278, opened 2016-02-05
(maint) Allow zip archives to be unpacked with 7za

melissa:
On windows, we need to use 7za to unpack zip archives. We don't have
`unzip` available and `gunzip` throws an error. The only problem is we
don't have any guarantees 7za will be on the path. It is installed as a
part of chocolatey, but we need to set path in the component in order to
ensure we have access to it.

### DIFF
--- a/lib/vanagon/component/source/http.rb
+++ b/lib/vanagon/component/source/http.rb
@@ -101,7 +101,7 @@ class Vanagon
             when ".tar.gz", ".tgz"
               return "gunzip -c '#{@file}' | '#{tar}' xf -"
             when ".zip"
-              return "unzip '#{@file}'"
+              return "unzip '#{@file}' || 7za x -r -tzip -o'#{File.basename(@file, ".zip")}' '#{@file}'"
             end
           else
             # Extension does not appear to be an archive


stahnma:
Doesn't windows have a built in unzip program since Vista?

melissa:
We've been using 7zip in puppet_for_the_win. I'm not sure what else we could use...

ScottGarman:
:+1:

shermdog:
Are we only dealing with .zip's on windows or also .tar.gz?

shermdog:
@stahnma I was thinking the same thing, turns out it's not available via command line.  You can do some stuff with powershell or the resource toolkit.  I'm in favor of using 7zip.

melissa:
@shermdog `.tar.gz` archives are dealt with in the conditional above this. That works fine since we have both `gunzip` and `tar` available. `gunzip` doesn't work with `.zip` archives :(

shermdog:
@melissa cool.  I was thinking we'd use 7zip for those also, but clearly not needed.  :+1:

MikaelSmith:
Native unzipping on Windows at http://stackoverflow.com/questions/17546016/how-can-you-zip-or-unzip-from-the-command-prompt-using-only-windows-built-in-ca/26843122#26843122. The one-liner
```
powershell.exe -nologo -noprofile -command "& { Add-Type -A 'System.IO.Compression.FileSystem'; [IO.Compression.ZipFile]::ExtractToDirectory('foo.zip', 'bar'); }"
```

PR #280, opened 2016-02-05
(RE-6468) Add Candle.exe to Wix/Vanagon

jcoconnor:
Add Candle.exe (Wix Compiler) commands to generate_msi_package for Vanagon

### DIFF
--- a/lib/vanagon/platform/windows.rb
+++ b/lib/vanagon/platform/windows.rb
@@ -109,19 +109,19 @@ class Vanagon
         dir_ref = "INSTALLDIR"
         # Actual array of commands to be written to the Makefile
         ["mkdir -p output/#{target_dir}",
-        "mkdir -p $(tempdir)/{staging,wix}",
-        "#{@copy} -r wix/* $(tempdir)/wix/",
-        "gunzip -c #{project.name}-#{project.version}.tar.gz | '#{@tar}' -C '$(tempdir)/staging' --strip-components 1 -xf -",
-        # Run the Heat command in a single pass
-        # Heat command documentation at: http://wixtoolset.org/documentation/manual/v3/overview/heat.html
-        #   dir <directory> - Traverse directory to find all sub-files and directories.
-        #   -ke             - Keep Empty directories
-        #   -cg             - Component Group Name
-        #   -gg             - Generate GUIDS now
-        #   -dr             - Directory reference to root directories (cannot contains spaces e.g. -dr MyAppDirRef)
-        #   -sreg           - Suppress registry harvesting.
-        "cd $(tempdir); \"$$WIX/bin/heat.exe\" dir staging -v -ke -indent 2 -cg #{cg_name} -gg -dr #{dir_ref} -t wix/#{project.name}.filter.xslt -sreg -out wix/#{project.name}-harvest.wxs",
-        ]
+          "mkdir -p $(tempdir)/{staging,wix}",
+          "#{@copy} -r wix/* $(tempdir)/wix/",
+          "gunzip -c #{project.name}-#{project.version}.tar.gz | '#{@tar}' -C '$(tempdir)/staging' --strip-components 1 -xf -",
+          # Run the Heat command in a single pass
+          # Heat command documentation at: http://wixtoolset.org/documentation/manual/v3/overview/heat.html
+          #   dir <directory> - Traverse directory to find all sub-files and directories.
+          #   -ke             - Keep Empty directories
+          #   -cg             - Component Group Name
+          #   -gg             - Generate GUIDS now
+          #   -dr             - Directory reference to root directories (cannot contains spaces e.g. -dr MyAppDirRef)
+          #   -sreg           - Suppress registry harvesting.
+          "cd $(tempdir); \"$$WIX/bin/heat.exe\" dir staging -v -ke -indent 2 -cg #{cg_name} -gg -dr #{dir_ref} -t wix/#{project.name}.filter.xslt -sreg -out wix/#{project.name}-harvest.wxs",
+          ]
       end
 
       # Method to derive the msi (Windows Installer) package name for the project.
--- a/lib/vanagon/platform/windows.rb
+++ b/lib/vanagon/platform/windows.rb
@@ -107,9 +107,10 @@ class Vanagon
         target_dir = project.repo ? output_dir(project.repo) : output_dir
         cg_name = "ProductComponentGroup"
         dir_ref = "INSTALLDIR"
+        candle_flags =  "-dPlatform=#{@architecture} -arch #{@architecture} -ext WiXUtilExtension -ext WixUIExtension"
         # Actual array of commands to be written to the Makefile
         ["mkdir -p output/#{target_dir}",
-          "mkdir -p $(tempdir)/{staging,wix}",
+          "mkdir -p $(tempdir)/{staging,wix/wixobj}",
           "#{@copy} -r wix/* $(tempdir)/wix/",
           "gunzip -c #{project.name}-#{project.version}.tar.gz | '#{@tar}' -C '$(tempdir)/staging' --strip-components 1 -xf -",
           # Run the Heat command in a single pass
@@ -121,6 +122,9 @@ class Vanagon
           #   -dr             - Directory reference to root directories (cannot contains spaces e.g. -dr MyAppDirRef)
           #   -sreg           - Suppress registry harvesting.
           "cd $(tempdir); \"$$WIX/bin/heat.exe\" dir staging -v -ke -indent 2 -cg #{cg_name} -gg -dr #{dir_ref} -t wix/#{project.name}.filter.xslt -sreg -out wix/#{project.name}-harvest.wxs",
+          # Apply Candle command to all *.wxs files - generates .wixobj files in wix directory.
+          # cygpath conversion is necessary as candle is unable to handle posix path specs
+          "cd $(tempdir)/wix/wixobj; for wix_file in `find $(tempdir)/wix -name \'*.wxs\'`; do \"$$WIX/bin/candle.exe\" #{candle_flags} $$(cygpath -aw $$wix_file) ; done",
           ]
       end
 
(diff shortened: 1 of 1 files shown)


lib/vanagon/platform/windows.rb
```diff
@@ -119,8 +119,9 @@ def generate_msi_package(project)
         #   -gg             - Generate GUIDS now
         #   -dr             - Directory reference to root directories (cannot contains spaces e.g. -dr MyAppDirRef)
         #   -sreg           - Suppress registry harvesting.
-        #   -var <variable> - Substitute File/@Source="SourceDir" with a preprocessor or a wix variable
-        "cd $(tempdir); \"$$WIX/bin/heat.exe\" dir staging -v -ke -indent 2 -cg #{cg_name} -gg -dr #{dir_ref} -t wix/#{project.name}.filter.xslt -sreg -var var.StageDir -out wix/#{project.name}-harvest.wxs",
+        "cd $(tempdir); \"$$WIX/bin/heat.exe\" dir staging -v -ke -indent 2 -cg #{cg_name} -gg -dr #{dir_ref} -t wix/#{project.name}.filter.xslt -sreg -out wix/#{project.name}-harvest.wxs",
+        # candle individual files to generate the wxo files.
+        "cd $(tempdir); \"$$WIX/bin/candle.exe\"  -arch #{@architecture} -out wix/#{project.name}-harvest.wixobj   wix/#{project.name}-harvest.wxs  ",
```
melissa: Do we need `-dPlatform=#{@architecture}` or any of the version information passed in with https://github.com/puppetlabs/puppet_for_the_win/blob/master/tasks/windows/windows.rake#L23 ? What do `-ext WiXUtilExtension -ext WixUIExtension` do any why don't we need them anymore?

It also looks like candle is called on all the wxs files (https://github.com/puppetlabs/puppet_for_the_win/blob/master/tasks/windows/windows.rake#L271-L285). Are we also going to have to do that? Should we glob for all the files in the `wxs` directory and run candle on all of them?

jcoconnor: @melissa Have taken the command directly from the Rakefile so need to determine exactly what options are actually necessary. Same applies to the WixUtil and UI extension options.

Yes - was thinking either of a glob, or specifying the filenames as a list so we can generalise the command at least in the Ruby code (even if we have to express it out as individual Makefile commands).

The "candle" command is not well formed yet - need to work on that.

Think we do need the .out.

jcoconnor: @melissa Had a look at the candle command in the Rakefile - The extra arguments appear to be to allow this command generalise all of the various input files (e.g. UI, Custom actions). This set of options is sufficient for the Harvest File only. 

I will probably leave this command as-is and then put in a more generalised command that works off a glob or array of filesnames (my preference) to be processed.

Would also prefer to keep the -out as it allows us to put the files in one place with the wixojb extension (don't know if it defaults to this, but could test it).

joshcooper:
`WiXUtilExtension` is needed for things like https://github.com/puppetlabs/puppet_for_the_win/blob/master/wix/puppet.wxs#L33 and https://github.com/puppetlabs/puppet_for_the_win/blob/master/wix/puppet.wxs#L455-L464. Note how the wix file assigns its to the `util` xml namespace: https://github.com/puppetlabs/puppet_for_the_win/blob/master/wix/puppet.wxs#L3

McdonaldSeanp:
:+1: but this needs a rebase

lib/vanagon/platform/windows.rb
```diff
@@ -107,21 +107,22 @@ def generate_msi_package(project)
...
+          # Run the Heat command in a single pass
+          # Heat command documentation at: http://wixtoolset.org/documentation/manual/v3/overview/heat.html
+          #   dir <directory> - Traverse directory to find all sub-files and directories.
+          #   -ke             - Keep Empty directories
+          #   -cg             - Component Group Name
+          #   -gg             - Generate GUIDS now
+          #   -dr             - Directory reference to root directories (cannot contains spaces e.g. -dr MyAppDirRef)
+          #   -sreg           - Suppress registry harvesting.
+          "cd $(tempdir); \"$$WIX/bin/heat.exe\" dir staging -v -ke -indent 2 -cg #{cg_name} -gg -dr #{dir_ref} -t wix/#{project.name}.filter.xslt -sreg -out wix/#{project.name}-harvest.wxs",
+          "cd $(tempdir)/wix; echo *.wxs | xargs -n 1 \"$WIX/bin/candle.exe\" -dPlatform=x64 -arch x64 -ext WiXUtilExtension -ext WixUIExtension"
```
melissa: Are we going to need to parse the `bill-of-materials` to pass in all the component versions (puppet, facter, openssl, hiera, etc) to candle, like what's currently happening in pftw?

melissa: Also the version pieces (major, minor, etc), it looks like we should be passing those in as well

melissa: For some of the wix files in pftw, `-dStageDir=` is getting passed in. Do we need that anywhere here?

jcoconnor: @melissa Need to check bill-of-materials
Correct - version pieces missing - will pull them in.
StageDir is not needed.

lib/vanagon/platform/windows.rb
```diff
@@ -107,21 +107,22 @@ def generate_msi_package(project)
         target_dir = project.repo ? output_dir(project.repo) : output_dir
         cg_name = "ProductComponentGroup"
         dir_ref = "INSTALLDIR"
+        candle_flags =  "-dPlatform=#{@architecture} -arch #{@architecture} -ext WiXUtilExtension -ext WixUIExtension"
```
melissa: It looks like this isn't used anymore. Was that on purpose?

lib/vanagon/platform/windows.rb
melissa: bill of materials is a file we include in the tarball that gets unpacked on line 115. It's generated with https://github.com/puppetlabs/vanagon/blob/021bfd97c1855da7566269839f43e7eb4b760168/lib/vanagon/project.rb#L246

lib/vanagon/platform/windows.rb
jcoconnor: Nope - Typo/edit error.... coming back in,

jcoconnor: @melissa and @McdonaldSeanp I don't think the version strings are needed as they are pulled in the ERB file translation in  [product.wxs.erb](https://github.com/puppetlabs/vanagon/blob/master/resources/windows/wix/project.wxs.erb#L9)

@jpogran can you verify and concur?

jcoconnor: @melissa As discussed, I think the best approach here is to ticket a separate task to ensure that the version payload is included correctly in the MSI.

I think the existing bill-of-materials code is also missing component version numbers, so will note that in the ticket.

melissa: Does your separate ticket take into account both the project version and the component versions?

jcoconnor: I have gone through the code (including history) and have determined that although the following variables are defined in windows.rake, the values are not used by the installer:
```
  flags['PuppetDescTag'] = describe 'downloads/puppet'
  flags['FacterDescTag'] = describe 'downloads/facter'
  flags['HieraDescTag']  = describe 'downloads/hiera'
  flags['MCODescTag']  = describe 'downloads/mcollective'
```
```puppet.wxi``` does reference two of them (?define), but again, doesn't use them.
So I think we can safely ignore these.

/cc @rob @iristyle @jpogran

Comment (jcoconnor) [inline lib/vanagon/platform/windows.rb]: I have created ticket [RE-6505](https://tickets.puppetlabs.com/browse/RE-6505) to pick up the issue of where the bill-of-materials goes. Have also created [RE-6506](https://tickets.puppetlabs.com/browse/RE-6506) to ensure that the project version is set correctly.

**Comment by melissa** [comment]
Will you break out the whitespace changes into a separate commit?

File: lib/vanagon/platform/windows.rb [hunk 4]
```diff
@@ -107,21 +107,24 @@ def generate_msi_package(project)
...
+          #   dir <directory> - Traverse directory to find all sub-files and directories.
+          #   -ke             - Keep Empty directories
+          #   -cg             - Component Group Name
+          #   -gg             - Generate GUIDS now
+          #   -dr             - Directory reference to root directories (cannot contains spaces e.g. -dr MyAppDirRef)
+          #   -sreg           - Suppress registry harvesting.
+          "cd $(tempdir); \"$$WIX/bin/heat.exe\" dir staging -v -ke -indent 2 -cg #{cg_name} -gg -dr #{dir_ref} -t wix/#{project.name}.filter.xslt -sreg -out wix/#{project.name}-harvest.wxs",
+          # Apply Candle command to all *.wxs files - generates .wixobj files in wix directory.
+          # cygpath conversion is necessary as candle is unable to handle posix path specs
+          "cd $(tempdir)/wix/wixobj; for wix_file in `find $(tempdir)/wix -name \'*.wxs\'`; do \"$$WIX/bin/candle.exe\" #{candle_flags} $$(cygpath -aw $$wix_file) ; done",
```
jcoconnor: @melissa Have incorporated your suggestions here.
Had to use the $(cygwin -aw) command to fixup the file path for Candle.

jcoconnor:
White-space change has also been broken out to separate commit.

lib/vanagon/platform/windows.rb
```diff
@@ -107,21 +107,24 @@ def generate_msi_package(project)
...
         ]
```
melissa: If we're using the trailing ']' syntax, we should have symmetrical syntax with the opening `[`.

jcoconnor: That better?

melissa: I wasn't referring to changing this line at all. I was referring to line 112.

shrug:
:+1:  lgtm, but I'm not really evaluating the specifics of the candle/heat calls

PR #282, opened 2016-02-08
(RE-6492) Add ability to choose user

jpogran:
Add ability to choose the ssh user that connects to the target VM. This allows
choosing different usernames than the default provided in the platform
files.

### DIFF
--- a/lib/vanagon/platform/dsl.rb
+++ b/lib/vanagon/platform/dsl.rb
@@ -228,6 +228,13 @@ class Vanagon
         @platform.ssh_port = port
       end
 
+      # Set the targetuser to use to connect
+      #
+      # @param type [String] user for the platform ('vagrant' for example)
+      def targetuser(user)
+        @platform.targetuser = user
+      end
+
       # Set the platform_triple for the platform
       #
       # @param triple[String] platform_triple for use in building out compiled


lib/vanagon/platform/dsl.rb
```diff
@@ -228,6 +228,13 @@ def ssh_port(port = 22)
         @platform.ssh_port = port
       end
 
+      # Set the targetuser to use to connect
+      #
+      # @param type [String] user for the platform ('vagrant' for example)
+      def targetuser(user)
+        @platform.targetuser = user
+      end
+
```
jcoconnor: @jpogran Will defer to the Vanagon experts, but if this for for vagrant, it should be specified as a new engine under lib/vanagon/engine

stahnma: I think this is ok where it is. Several engines have the option of using a specific user.

stahnma:
Rebase (and add a spec test).

jcoconnor:
@jpogran This has been sitting around for a while - If you want to rebase it we can test it and merge.

jcoconnor:
@jpogran Just having a look at this ticket.
It seems the functionality has already been added to Vanagon at: https://github.com/puppetlabs/vanagon/blob/master/lib/vanagon/platform/dsl.rb#L287-L292 added in bab55a46ff25f1406da91467b5d8ef9266c78d92

PR #283, opened 2016-02-09
(RE-6455) Create directorylist.wxs

McdonaldSeanp:
This commit will add functionality to use the existing
get_directories functionality for a project to generate
a directorylist.wxs file that contains the wix structure
of directories for the project.

As part of this feature this commit also extends the current
VANAGON::COMMON::PATHNAME structure to include "id" for
use in the wix structure. it does not specifically extend
the actual COMMON::PATHNAME class but instead creates a
new class to extend pathname under the namespace
VANAGON::WINDOWS:PATHNAME

finally this commit adds rspec tests for the new functionality
and updates any old rspec that would break with the new
functionality

Note that in order to get everything working in a way that doesn't
look confusing we had to update params on one specific rubocop
rule

### DIFF
--- a/lib/vanagon/component/dsl.rb
+++ b/lib/vanagon/component/dsl.rb
@@ -315,8 +315,12 @@ class Vanagon
       # @param mode [String] octal mode to apply to the directory
       # @param owner [String] owner of the directory
       # @param group [String] group of the directory
-      def directory(dir, mode: nil, owner: nil, group: nil)
-        @component.directories << Vanagon::Common::Pathname.new(dir, mode: mode, owner: owner, group: group)
+      def directory(dir, mode: nil, owner: nil, group: nil, wix_id: nil)
+        if @component.platform.is_windows?
+          @component.directories << Vanagon::Windows::Pathname.new(dir, mode: mode, owner: owner, group: group, wix_id: wix_id)
+        else
+          @component.directories << Vanagon::Common::Pathname.new(dir, mode: mode, owner: owner, group: group)
+        end
       end
 
       # Adds a set of environment overrides to the environment for a component.
--- a/lib/vanagon/platform/windows.rb
+++ b/lib/vanagon/platform/windows.rb
@@ -56,6 +56,7 @@ class Vanagon
         FileUtils.mkdir_p(File.join(workdir, "wix"))
         erb_file(File.join(VANAGON_ROOT, "resources/windows/wix/project.wxs.erb"), File.join(workdir, "wix",  "#{name}.wxs"), false, { :binding => binding })
         erb_file(File.join(VANAGON_ROOT, "resources/windows/wix/project.filter.xslt.erb"), File.join(workdir, "wix", "#{name}.filter.xslt"), false, { :binding => binding })
+        erb_file(File.join(VANAGON_ROOT, "resources/windows/wix/directorylist.wxs.erb"), File.join(workdir, "wix", "directorylist.#{name}.wxs"), false, { :binding => binding })
       end
 
       # Method to generate the files required to build a nuget package for the project
@@ -221,6 +222,91 @@ class Vanagon
       # @return [String] relative path to where windows packages should be staged
       def output_dir(target_repo = "")
         File.join("windows", target_repo, @architecture)
+      end
+
+      # Generate correctly formatted wix elements that match the
+      # structure of the directory input
+      #
+      # @param directories, Array of hashes including at least :path
+      # and optionally:
+      # @return [string] correctly formatted wix element string
+      def generate_wix_dirs(project)
+        if project.get_directories.empty?
+          raise Vanagon::Error, 'ERROR No directories specified!'
+        else
+          directories = []
+          project.get_directories.map { |dir| directories.push({ :path => dir.path, :id => dir.wix_id }) }
+
+          root = { :children => [] }
+
+          # iterate over all paths specified and break each one
+          # in to its specific directories. This will generate_wix_dirs
+          # an n-ary tree structure matching the specs from the input
+          directories.each do |dir|
+            # Always start at the beginning
+            curr = root
+            names = strip_path(dir[:path])
+            names.each do |name|
+              #The Id field will default to name, but be overridden later
+              new_obj = { :name => name, :id => name, :children => [] }
+              if (child_index = includes_child(new_obj, curr[:children]))
+                curr = curr[:children][child_index]
+              else
+                curr[:children].push(new_obj)
+                curr = new_obj
+              end
+            end
+            # at this point, curr will be the top dir, override the id if
+            # id exists
+            if dir[:id]
+              curr[:id] = dir[:id]
+            end
+          end
+          return generate_wix_from_graph(root)
+        end
+      end
+
+      # strip and split the directory path into single names
+      # @param [string] path string of directory
+      def strip_path(path)
+        if path.include?("/") || path.include?("\\")
+          # The regex in the last part of this if warrants some
+          # explanation. Specifically it matches any combinations
+          # of any letters, then the : char, then finally either
+          # the char / or the char \. it's menat to parse out drive
+          # roots on windows
+          if path.start_with?("/") || path.start_with?("\\") || path =~ (/([A-Za-z])*\:(\/|\\)/)
+            path = path.sub(/\/|\\|([A-Za-z])*\:(\/|\\)/, '')
+          end
+          names = path.split(/\/|\\/)
+        end
+        return names
+      end
+
+      # Find if child element is the same as one of
+      # the old_children elements, return that child
+      def includes_child(new_child, old_children)
+        old_children.each_with_index do |curr_old_child, index|
+          return index if curr_old_child[:name] == new_child[:name]
+        end unless old_children.empty?
+        return nil
+      end
+
+      # Recursively generate wix element structure
+      #
+      # @param root, the (empty) root of an n-ary tree containing the
+      # structure of directories
+      def generate_wix_from_graph(root)
+        string = ''
+        unless root[:children].empty?
+          root[:children].each do |child|
+            string += ("<Directory Name=\"#{child[:name]}\" Id=\"#{child[:id]}\">\n")
+            string += generate_wix_from_graph(child)
+            string += ("</Directory>\n")
+          end
+          return string
+        end
+        return ''
       end
 
       # Constructor. Sets up some defaults for the windows platform and calls the parent constructor
--- a/lib/vanagon/windows/pathname.rb
+++ b/lib/vanagon/windows/pathname.rb
@@ -1,0 +1,32 @@
+require 'pathname'
+
+class Vanagon
+  class Windows
+    class Pathname < Vanagon::Common::Pathname
+
+      # @!attribute wix_id
+      #   @return [String] Returns the wix_id representing the id representing
+      #   a file or directory in wix
+      attr_accessor :wix_id
+
+      # Extension of the VANAGON::COMMON::PATHNAME class specifially for windows
+      # @param [String, Integer] mode the UNIX Octal permission string to use when this file is archived
+      # @param [String, Integer] owner the username or UID to use when this file is archived
+      # @param [String, Integer] group the groupname or GID to use when this file is archived
+      # @param [Boolean] options => {config} mark this file as a configuration file, stored as private state
+      #   and exposed through the {#configfile?} method.
+      # @param [string] options => {id} the id to pass to wix objects during directory generation
+      # @return [Vanagon::Common::Pathname] Returns a new Pathname instance.
+      def initialize(path, mode: nil, owner: nil, group: nil, config: false, wix_id: nil)
+        super(path, mode: nil, owner: nil, group: nil, config: false)
+        # When the wix actually gets generated the "id" field in the wix elements
+        # will already default to whatever the name field indicates. this is only
+        # if you want to override that
+        if wix_id
+          @wix_id ||= wix_id
+        end
+      end
+
+    end
+  end
+end
--- a/resources/windows/wix/directorylist.wxs.erb
+++ b/resources/windows/wix/directorylist.wxs.erb
@@ -1,0 +1,9 @@
+<?xml version='1.0' encoding='windows-1252'?>
+<Wix xmlns='http://schemas.microsoft.com/wix/2006/wi' xmlns:util="http://schemas.microsoft.com/wix/UtilExtension">
+
+  <Fragment>
+    <DirectoryRef Id='TARGETDIR'>
+      <%= @platform.generate_wix_dirs(self) %>
+    </DirectoryRef>
+  </Fragment>
+</Wix>
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -489,6 +489,7 @@ Lint/UnderscorePrefixedVariableName:
 
 Metrics/ParameterLists:
   Enabled: true
+  CountKeywordArgs: false
 
 Lint/RequireParentheses:
   Enabled: true
--- a/lib/vanagon/driver.rb
+++ b/lib/vanagon/driver.rb
@@ -4,6 +4,7 @@ require 'vanagon/component'
 require 'vanagon/utilities'
 require 'vanagon/common'
 require 'vanagon/errors'
+require 'vanagon/windows'
 require 'tmpdir'
 require 'logger'
 
--- a/lib/vanagon/project/dsl.rb
+++ b/lib/vanagon/project/dsl.rb
@@ -143,8 +143,12 @@ class Vanagon
       # @param mode [String] octal mode to apply to the directory
       # @param owner [String] owner of the directory
       # @param group [String] group of the directory
-      def directory(dir, mode: nil, owner: nil, group: nil)
-        @project.directories << Vanagon::Common::Pathname.new(dir, mode: mode, owner: owner, group: group)
+      def directory(dir, mode: nil, owner: nil, group: nil, wix_id: nil)
+        if @project.platform.is_windows?
+          @project.directories << Vanagon::Windows::Pathname.new(dir, mode: mode, owner: owner, group: group, wix_id: wix_id)
+        else
+          @project.directories << Vanagon::Common::Pathname.new(dir, mode: mode, owner: owner, group: group)
+        end
       end
 
       # Add a user to the project
--- a/lib/vanagon/windows.rb
+++ b/lib/vanagon/windows.rb
@@ -1,0 +1,1 @@
+require 'vanagon/windows/pathname.rb'
--- a/spec/lib/vanagon/component/dsl_spec.rb
+++ b/spec/lib/vanagon/component/dsl_spec.rb
@@ -1,5 +1,6 @@
 require 'vanagon/component/dsl'
 require 'vanagon/common'
+require 'vanagon/windows'
 require 'json'
 
 describe 'Vanagon::Component::DSL' do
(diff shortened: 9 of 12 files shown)


resources/windows/wix/directorylist.wxs.erb
```diff
@@ -0,0 +1,16 @@
+<?xml version='1.0' encoding='windows-1252'?>
+<Wix xmlns='http://schemas.microsoft.com/wix/2006/wi' xmlns:util="http://schemas.microsoft.com/wix/UtilExtension">
+
+ <!--
+  This is how to include wxi files.  We expect this file to define variables
+  like the version numbers and the upgrade UUID.
+  -->
+  <?include $(sys.SOURCEFILEDIR)include/puppet.wxi ?>
```
melissa: You don't need this line. It previously was defining variables to be used by wix, but we're supplanting that with ruby variables.

lib/vanagon/platform/windows.rb
```diff
@@ -221,6 +222,73 @@ def output_dir(target_repo = "")
...
+      #
+      # @param directories, Array of hashes including at least :path
+      # and optionally:
+      # @return [string] correctly formatted wix element string
+      def generate_wix_dirs(project)
+        if project.get_directories.empty?
+          raise 'ERROR No directories specified!'
+        else
+        directories = []
+        project.get_directories.map {|dir| directories.push({:path => dir.path, :id => dir.id})}
```
jcoconnor: These two lines need indentation as well

lib/vanagon/platform/windows.rb
```diff
@@ -221,6 +222,73 @@ def output_dir(target_repo = "")
...
 
+      # Generate correctly formatted wix elements that match the
+      # structure of the directory input
+      #
+      # @param directories, Array of hashes including at least :path
+      # and optionally:
+      # @return [string] correctly formatted wix element string
+      def generate_wix_dirs(project)
+        if project.get_directories.empty?
+          raise 'ERROR No directories specified!'
```
melissa: OOh, we should raise a vanagon error here!

`raise Vanagon::Error, 'ERROR No directories specified!'`

McdonaldSeanp: Alrighty

lib/vanagon/common/pathname.rb
```diff
@@ -28,15 +28,18 @@ class Pathname
       # @param [String, Integer] group the groupname or GID to use when this file is archived
       # @param [Boolean] config mark this file as a configuration file, stored as private state
       #   and exposed through the {#configfile?} method.
+      # @param [string] id the id to pass to wix objects during directory generation
       # @return [Vanagon::Common::Pathname] Returns a new Pathname instance.
-      def initialize(path, mode: nil, owner: nil, group: nil, config: false)
+      def initialize(path, mode: nil, owner: nil, group: nil, config: false, id: nil)
```
melissa: I think you're going to want to update https://github.com/puppetlabs/vanagon/blob/f7a2b4ccdf2edd5a28ef49a469f662a6252b2a60/lib/vanagon/project/dsl.rb#L146 as well to pass in `id`?

McdonaldSeanp: problem is rubocop does not like more than 5 params

boooooooo

melissa: that's a thing :( we could change this to an options hash?

McdonaldSeanp: Right, which is likely what i would do, but now to me rubocop is making things less clear yes? seems counterproductive to what rubocop is for.

melissa:
What's the logic behind creating a new windows::pathname class rather than extending the original pathname? Is there harm in passing in wix_id => nil for non windows platforms?

McdonaldSeanp:
The logic behind it is mostly forward looking. Instead of including windows specific functionality to the Common::Pathname namespace, instead we should create a Windows::Pathname namespace for the windows specific stuff

melissa:
We should probably break these changes into more than one commit. The directory list stuff should be a commit, but changing `config` to be a part of the `options` hash should be its own commit

lib/vanagon/platform/windows.rb
```diff
@@ -221,6 +222,73 @@ def output_dir(target_repo = "")
...
+        return nil
+      end
+
+      # Recursively generate wix element structure
+      #
+      # @param root, the (empty) root of an n-ary tree containing the
+      # structure of directories
+      def generate_wix_from_graph(root)
+        string = ''
+        unless root[:children].empty?
```
melissa: Why do we have to check if `root[:children]` is empty? We should just be able to loop through and return string. I guess this would fail if `root[:children]` were nil and there was no array to loop over... But on line 238 you're setting it explicitely. I feel like that's an assumption we should be safe making, in which case lines 277, 288, and 289 shouldn't be necessary.

resources/windows/wix/directorylist.wxs.erb
```diff
@@ -0,0 +1,9 @@
+<?xml version='1.0' encoding='windows-1252'?>
+<Wix xmlns='http://schemas.microsoft.com/wix/2006/wi' xmlns:util="http://schemas.microsoft.com/wix/UtilExtension">
+
+  <Fragment>
+
+  <%- generate_wix_dirs(project) -%>
```
melissa: Three things:
- You're going to want to start this with `<%=`
- to access the wix_dirs method, you need to call it like `@platform.generate_wix_dirs`
- from what I can tell, we don't actually have access to the project object at this point. We do, however, have access to `get_directories`

```
[4] pry(#<Vanagon::Project>)> get_directories
=> [#<Vanagon::Windows::Pathname:0x007fa596992a30 @config=nil, @group=nil, @mode=nil, @owner=nil, @path="C:/ProgramFiles64Folder/PuppetLabs", @wix_id=nil>,
 #<Vanagon::Windows::Pathname:0x007fa596992210 @config=nil, @group=nil, @mode=nil, @owner=nil, @path="C:/ProgramFiles64Folder/PuppetLabs/Puppet", @wix_id=nil>,
 #<Vanagon::Windows::Pathname:0x007fa596991928 @config=nil, @group=nil, @mode=nil, @owner=nil, @path="C:/CommonAppDataFolder/PuppetLabs", @wix_id=nil>,
 #<Vanagon::Windows::Pathname:0x007fa5969910e0 @config=nil, @group=nil, @mode=nil, @owner=nil, @path="C:/ProgramFiles64Folder/PuppetLabs/bin", @wix_id=nil>,
 #<Vanagon::Common::Pathname:0x007fa596ba3798 @config=nil, @group=nil, @mode="0750", @owner=nil, @path="C:/CommonAppDataFolder/PuppetLabs/puppet/cache">,
 #<Vanagon::Common::Pathname:0x007fa596ba2e38 @config=nil, @group=nil, @mode=nil, @owner=nil, @path="C:/CommonAppDataFolder/PuppetLabs/puppet/etc">,
 #<Vanagon::Common::Pathname:0x007fa596ba25c8 @config=nil, @group=nil, @mode=nil, @owner=nil, @path="C:/CommonAppDataFolder/PuppetLabs/code">,
 #<Vanagon::Common::Pathname:0x007fa596ba1da8 @config=nil, @group=nil, @mode=nil, @owner=nil, @path="C:/CommonAppDataFolder/PuppetLabs/code/modules">,
 #<Vanagon::Common::Pathname:0x007fa596ba14c0 @config=nil, @group=nil, @mode=nil, @owner=nil, @path="C:/ProgramFiles64Folder/PuppetLabs/Puppet/modules">,
 #<Vanagon::Common::Pathname:0x007fa596ba08e0 @config=nil, @group=nil, @mode=nil, @owner=nil, @path="C:/CommonAppDataFolder/PuppetLabs/code/environments">,
 #<Vanagon::Common::Pathname:0x007fa596b979c0 @config=nil, @group=nil, @mode=nil, @owner=nil, @path="C:/CommonAppDataFolder/PuppetLabs/code/environments/production">,
 #<Vanagon::Common::Pathname:0x007fa596b96ac0 @config=nil, @group=nil, @mode=nil, @owner=nil, @path="C:/CommonAppDataFolder/PuppetLabs/code/environments/production/manifests">,
 #<Vanagon::Common::Pathname:0x007fa596b95f08 @config=nil, @group=nil, @mode=nil, @owner=nil, @path="C:/CommonAppDataFolder/PuppetLabs/code/environments/production/modules">,
 #<Vanagon::Common::Pathname:0x007fa596b95378 @config=nil, @group=nil, @mode=nil, @owner=nil, @path="C:/CommonAppDataFolder/PuppetLabs/puppet/var/log">,
 #<Vanagon::Common::Pathname:0x007fa596b94860 @config=nil, @group=nil, @mode=nil, @owner=nil, @path="C:/CommonAppDataFolder/PuppetLabs/puppet/var/run">,
 #<Vanagon::Common::Pathname:0x007fa596b3dad8 @config=nil, @group=nil, @mode=nil, @owner=nil, @path="C:/CommonAppDataFolder/PuppetLabs/facter/facts.d">,
 #<Vanagon::Common::Pathname:0x007fa596b32cc8 @config=nil, @group=nil, @mode=nil, @owner=nil, @path="C:/CommonAppDataFolder/PuppetLabs/code/hieradata">,
 #<Vanagon::Common::Pathname:0x007fa596aefc70 @config=nil, @group=nil, @mode=nil, @owner=nil, @path="C:/CommonAppDataFolder/PuppetLabs/mcollective/etc">,
 #<Vanagon::Common::Pathname:0x007fa596aef270 @config=nil, @group=nil, @mode=nil, @owner=nil, @path="C:/CommonAppDataFolder/PuppetLabs/mcollective/plugins">,
 #<Vanagon::Common::Pathname:0x007fa596aee668 @config=nil, @group=nil, @mode=nil, @owner=nil, @path="C:/CommonAppDataFolder/PuppetLabs/mcollective/var/log">,
 #<Vanagon::Common::Pathname:0x007fa596a6c078 @config=nil, @group=nil, @mode=nil, @owner=nil, @path="C:/CommonAppDataFolder/PuppetLabs/pxp-agent">,
 #<Vanagon::Common::Pathname:0x007fa596a631d0 @config=nil, @group=nil, @mode=nil, @owner=nil, @path="C:/CommonAppDataFolder/PuppetLabs/pxp-agent/etc/modules">,
 #<Vanagon::Common::Pathname:0x007fa596a62168 @config=nil, @group=nil, @mode=nil, @owner=nil, @path="C:/CommonAppDataFolder/PuppetLabs/pxp-agent/var/spool">,
 #<Vanagon::Common::Pathname:0x007fa596a612b8 @config=nil, @group=nil, @mode=nil, @owner=nil, @path="C:/CommonAppDataFolder/PuppetLabs/pxp-agent/var/log">,
 #<Vanagon::Common::Pathname:0x007fa596a60278 @config=nil, @group=nil, @mode=nil, @owner=nil, @path="C:/CommonAppDataFolder/PuppetLabs/pxp-agent/var/run">]
```
Rather than passing in project, could we pass in an array of directories to parse?

lib/vanagon/platform/windows.rb
McdonaldSeanp: No, this is the base case for the recursive function (refer to the call of the function itself on line 284). Without this case the function would just call itself indefinately.

melissa: Oooh, I forgot about recursion as a thing... cool, thanks for explaining!

resources/windows/wix/directorylist.wxs.erb
McdonaldSeanp: So I can do that, but get_directories is a function inside the project object, so if we do not have access to the project object I'm fairly certain we would not have access to the get_directories functionality.

McdonaldSeanp:
Also I will create a new commit for the stuff you mentioned

lib/vanagon/platform/windows.rb
McdonaldSeanp: yep ;)

lib/vanagon/windows/pathname.rb
```diff
@@ -0,0 +1,27 @@
...
+      # @param [String, Integer] mode the UNIX Octal permission string to use when this file is archived
+      # @param [String, Integer] owner the username or UID to use when this file is archived
+      # @param [String, Integer] group the groupname or GID to use when this file is archived
+      # @param [Boolean] options => {config} mark this file as a configuration file, stored as private state
+      #   and exposed through the {#configfile?} method.
+      # @param [string] options => {id} the id to pass to wix objects during directory generation
+      # @return [Vanagon::Common::Pathname] Returns a new Pathname instance.
+      def initialize(path, mode: nil, owner: nil, group: nil, options: {})
+        super(path, mode: nil, owner: nil, group: nil, options: {})
+        @wix_id ||= options[:wix_id]
```
melissa: We should check if we have `:wix_id` set, and if not, we should default to `File.basename(path)`

McdonaldSeanp: Coo,

lib/vanagon/common/pathname.rb
```diff
@@ -26,17 +26,18 @@ class Pathname
       # @param [String, Integer] mode the UNIX Octal permission string to use when this file is archived
       # @param [String, Integer] owner the username or UID to use when this file is archived
       # @param [String, Integer] group the groupname or GID to use when this file is archived
-      # @param [Boolean] config mark this file as a configuration file, stored as private state
+      # @param [Boolean] options => {config} mark this file as a configuration file, stored as private state
       #   and exposed through the {#configfile?} method.
       # @return [Vanagon::Common::Pathname] Returns a new Pathname instance.
-      def initialize(path, mode: nil, owner: nil, group: nil, config: false)
+      def initialize(path, mode: nil, owner: nil, group: nil, options: {})
```
mckern: I'm weary of introducing the Options hash pattern into a series of named parameters. Was this done because `rubocop` hollered at you?

McdonaldSeanp: Yup

resources/windows/wix/directorylist.wxs.erb
```diff
@@ -0,0 +1,9 @@
+<?xml version='1.0' encoding='windows-1252'?>
+<Wix xmlns='http://schemas.microsoft.com/wix/2006/wi' xmlns:util="http://schemas.microsoft.com/wix/UtilExtension">
+
+  <Fragment>
+
+  <%- @project.generate_wix_dirs(self) -%>
```
melissa: p.s., this needs to be nested under `<Directory Id='TARGETDIR' Name='SourceDir' >`

Also, this still needs to start with `<%=` otherwise we aren't going to get any output

lib/vanagon/platform/windows.rb
```diff
@@ -221,6 +222,73 @@ def output_dir(target_repo = "")
...
+          root[:children].each do |child|
+            if child[:id]
+              string += ("<Directory Name=\"#{child[:name]}\" Id=\"#{child[:id]}\">\n")
+            else
+              string += ("<Directory Name=\"#{child[:name]}\">\n")
```
melissa: Everything always needs in ID regardless. For now, let's assume id=name if id isn't set

resources/windows/wix/directorylist.wxs.erb
melissa: Shit, sorry, not the directory definition. It needs to be nested under `<DirectoryRef Id='TARGETDIR' />` instead

lib/vanagon/common/pathname.rb
```diff
@@ -26,17 +26,18 @@ class Pathname
...
         @path = ::Pathname.new(path).cleanpath.to_s
         @mode ||= mode
         @owner ||= owner
         @group ||= group
-        @config ||= config
+        @config ||= options[:config]
       end
 
+
```
melissa: extra white space on purpose?

melissa:
1) component directory also needs to get this update
2) `Id`s need to be set for all directory elements in the wix file
3) these directories need to be nested between `<DirectoryRef Id='TARGETDIR'>` and `</DirectoryRef>`

I think those are the only changes that I made to get this to compile successfully.

melissa:
We could also add a check for illegal `Id`s, which could be helpful for people. Not a blocker to getting this in though.
`error CNDL0014 : The Directory/@Id attribute's value, 'pxp-agent', is not a legal identifier.  Identifiers may contain ASCII character A-Z, a-z, digits, underscores (_), or periods (.).  Every identifier must begin with either a letter or an underscore.`

McdonaldSeanp:
As per your comment about C: not being included, do you mean that we should never have the actual drive (like C: or D: or whatever) in this directory structure?

melissa:
Yeah, we should never include the actual drive. I know I've defined it as a part of the path, but we shouldn't include it. Sorry for not pointing that out earlier

McdonaldSeanp:
It's cool, I'll write that out right now

lib/vanagon/windows/pathname.rb
```diff
@@ -0,0 +1,31 @@
...
+      # @param [String, Integer] owner the username or UID to use when this file is archived
+      # @param [String, Integer] group the groupname or GID to use when this file is archived
+      # @param [Boolean] options => {config} mark this file as a configuration file, stored as private state
+      #   and exposed through the {#configfile?} method.
+      # @param [string] options => {id} the id to pass to wix objects during directory generation
+      # @return [Vanagon::Common::Pathname] Returns a new Pathname instance.
+      def initialize(path, mode: nil, owner: nil, group: nil, options: {})
+        super(path, mode: nil, owner: nil, group: nil, options: {})
+        if options[:wix_id]
+          @wix_id = options[:wix_id]
```
melissa: If you wanted to play with ternary operators, you could also set this to 
```
@wix_id ||= options[:wix_id].nil? ? File.basename(path) : options[:wix_id]
```

lib/vanagon/common/pathname.rb
McdonaldSeanp: Nope

lib/vanagon/windows/pathname.rb
McdonaldSeanp: It'll actually be repeated functionality, I already am going to set the wix_id to the name in the geenerate_wix_dir function, it's easiest to do it there and I think it makes the most sense.

melissa: Cool, whatever works best

resources/windows/wix/directorylist.wxs.erb
```diff
@@ -0,0 +1,9 @@
+<?xml version='1.0' encoding='windows-1252'?>
+<Wix xmlns='http://schemas.microsoft.com/wix/2006/wi' xmlns:util="http://schemas.microsoft.com/wix/UtilExtension">
+
+  <Fragment>
+    <DirectoryRef Id='TARGETDIR'>
+      <%= @project.generate_wix_dirs(self) %>
```
melissa: Should be `@platform`

lib/vanagon/platform/windows.rb
melissa: should this be a fail case instead?

McdonaldSeanp: No, it's just deprecated, it'll never get here

McdonaldSeanp: fixing now

melissa:
```
Windows Installer XML Toolset Compiler version 3.10.2.2516
Copyright (c) Outercurve Foundation. All rights reserved.

directorylist.wxs
C:\cygwin64\var\tmp\tmp.VV7jcZkjTg\wix\directorylist.wxs(58) : error CNDL0014 : The Directory/@Id attribute's value, 'pxp-agent', is not a legal identifier.  Identifiers may contain ASCII characters A-Z, a-z, digits, underscores (_), or periods (.).  Every identifier must begin with either a letter or an underscore.
Windows Installer XML Toolset Compiler version 3.10.2.2516
Copyright (c) Outercurve Foundation. All rights reserved.
```
That one failure is user error. I'm a hearty :+1:  on this.

jcoconnor:
@McdonaldSeanp Have tested this with Candle and it looks good.
Its generating some Identifier warnings, but these can be fixed up in the puppet-agent product file.
lgtm :+1: 

Will merge once Travis wakes up

mckern:
I like a lot of this but it *needs* to be broken into more than one commit. The rubocop change alone is a single commit.

lib/vanagon/platform/windows.rb
```diff
@@ -224,6 +225,91 @@ def output_dir(target_repo = "")
...
+          return index if curr_old_child[:name] == new_child[:name]
+        end unless old_children.empty?
+        return nil
+      end
+
+      # Recursively generate wix element structure
+      #
+      # @param root, the (empty) root of an n-ary tree containing the
+      # structure of directories
+      def generate_wix_from_graph(root)
```
jcoconnor: Suggest you add another argument here to track nesting level.
This will allow you to prefix the directory with the requisite number of spaces.

lib/vanagon/platform/windows.rb
```diff
@@ -224,6 +225,91 @@ def output_dir(target_repo = "")
...
+        string = ''
+        unless root[:children].empty?
+          root[:children].each do |child|
+            string += ("<Directory Name=\"#{child[:name]}\" Id=\"#{child[:id]}\">\n")
```
jcoconnor: Once you have the level of nesting, you should use it here to indent the <Directory> (and </Directory>) below.
This will be invaluable for debugging.

lib/vanagon/platform/windows.rb
mckern: @jcoconnor can you clarify your intent? I don't follow.

McdonaldSeanp: Correct me if i'm wrong @jcoconnor but the intent is so we can add correctly formatted indents in front of the elements so debugging this directorylist file would be easier

jcoconnor: Thats correct @McdonaldSeanp

mckern: That is an obtuse way to solve the problem; if we're worried about things like indented formatting, wouldn't it be easier to just run the "final" XML through a formatter/linter than adding a brittle parameter?

jcoconnor: Right on Sherlock :+1:

mckern: I don't know how to interpret that, but I'm choosing to read it as "I agree. Let's not add a brittle parameter that couples functional implementation to transitional output."

jcoconnor: That would be the correct interpretation - I was complimenting you on an excellent suggestion.

McdonaldSeanp:
I broke it out into three commits, hopefully that makes more sense. and it's been rebased. unless the commit structure should look different this is ready for merge again

lib/vanagon/component/dsl.rb
```diff
@@ -316,8 +316,12 @@ def add_source(url, ref: nil, sum: nil)
       # @param mode [String] octal mode to apply to the directory
       # @param owner [String] owner of the directory
       # @param group [String] group of the directory
-      def directory(dir, mode: nil, owner: nil, group: nil)
-        @component.directories << Vanagon::Common::Pathname.new(dir, mode: mode, owner: owner, group: group)
+      def directory(dir, mode: nil, owner: nil, group: nil, wix_id: nil)
```
mckern: is `wix_id` ultimately just a GUID value?

mckern:
Typo in `1f3bfb9d` (`RUbocop` vs. `Rubocop`). Probably just a nitpick.

lib/vanagon/component/dsl.rb
melissa: No, https://github.com/puppetlabs/puppet_for_the_win/blob/master/wix/puppet.wxs#L151 is a good example of what `wix_id` could be set to

mckern: Ah, I understand now that it's a value used as a reference for other things that might need to perform operations on it. What an overly complicated format.

McdonaldSeanp:
Typo is fixed

lib/vanagon/component/dsl.rb
jcoconnor: Yes - it is a bit overly complex - GUID's are sometimes used because they are unique, but you can use any string that is unique within the whole wix project. The string is of [Microsoft Type "Identifier"](https://msdn.microsoft.com/library/aa369212.aspx) therefore must consist of Alphanumeric characters or the "." and "_" characters and must begin with a letter or underscore.

McdonaldSeanp:
This will be opened in a new fresh PR with different changes

PR #286, opened 2016-02-11
(maint) Add license to components

melissa:
We should have the ability to specify what license each of our
components has. Although not all of the packaging formats we support
will be able to present all the component licenses, some do. For those
that do, we should make this piece of information available.


jcoconnor:
:+1:

mckern:
I want to merge this. Can you expand the commit message at all? Otherwise, it's ready to go.

PR #284, opened 2016-02-11
(CPR-293) update local engine

pyther:
Builds currently fail when using the local engine, because @remote_dir
is not defined. Additionally, gem install fails due RUBY environment
variables being set by 'bundle exec build.'

Changes
- engine::local inherits from engine::base
- local_command function emulates remote_ssh_command behavior
- unset environment variables set by 'bundle exec build'

PR #290, opened 2016-02-17
(CPR-293, CPR-294) Fix Local engine, unbundle local commands

mckern:
This PR replaces PR #284.

Builds currently fail when using the local engine, because `@remote_dir` is not defined. Additionally, `gem install` fails due to numerous `RUBY_` and `BUNDLER_` environment variables being set when running `bundle exec build`.

I worked with @pyther to bang his initial work into something solid, and wrote some spec tests to cover the initial scope of the work done. `Vanagon::Utilities#local_command` now runs in a clean, unbundled environment and the Local build engine produces usable builds. There's still room for improvement (test coverage is still not 100% on the new work) but this should be in sturdy, workable shape now.

Thanks to @MikaelSmith and @pyther for their work on the original PRs.

### DIFF
--- a/lib/vanagon/engine/local.rb
+++ b/lib/vanagon/engine/local.rb
@@ -1,40 +1,35 @@
+require 'vanagon/engine/base'
 require 'vanagon/utilities'
 require 'vanagon/errors'
-require 'benchmark'
 
 class Vanagon
   class Engine
-    class Local
-      attr_accessor :target
+    class Local < Base
 
       def initialize(platform, target = nil)
-        @platform = platform
-        @target = "local machine"
+        @target = target || "local machine"
         @name = 'local'
+        super
+
+        # We inherit a set of required attributes from Base,
+        # and rather than instantiate a new empty array for
+        # required attributes, we can just clear out the
+        # existing ones.
+        @required_attributes.clear
       end
 
       # Dispatches the command for execution
-      def dispatch(command)
-        puts Benchmark.measure { local_command(command, @workdir) }
-      end
-
-      # Steps needed to tear down or clean up the system after the build is
-      # complete
-      def teardown
-      end
-
-      # This method will take care of validation and target selection all at
-      # once as an easy shorthand to call from the driver
-      def startup(workdir)
-        @workdir = workdir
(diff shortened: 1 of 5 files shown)


ScottGarman:
:+1:

lib/vanagon/engine/local.rb
```diff
@@ -1,40 +1,35 @@
...
-
-      # This method will take care of validation and target selection all at
-      # once as an easy shorthand to call from the driver
-      def startup(workdir)
-        @workdir = workdir
-        script = @platform.provisioning.join(' && ')
-        dispatch(script)
+      # Dispatches the command for execution
+      def dispatch(command, return_output = false)
+        Vanagon::Utilities.local_command(command, return_command_output: return_output)
```
MikaelSmith: Just noticed local_command doesn't support return_command_output. I'm not sure how this change made it into the original PR.

MikaelSmith: Oh, some of the changes to local_command didn't make it to the end of this PR. @mckern

lib/vanagon/utilities.rb
```diff
@@ -267,13 +267,22 @@ def remote_ssh_command(target, command, port = 22, return_command_output: false)
...
+    # @param workdir [String] change to directory <workdir> before running <command>
     # @return [true] Returns true if the command was successful
     # @raise [RuntimeError] If the command fails an exception is raised
-    def local_command(command, workdir)
-      puts "Executing '#{command}' locally in #{workdir}"
-      Kernel.system(command, :chdir => workdir)
-      $?.success? or raise "Local command (#{command}) failed."
+    def local_command(command, workdir: Dir.pwd)
+      clean_environment do
+        puts "Executing '#{command}' locally in #{workdir}"
```
MikaelSmith: I think all the return_command_output handling was probably last in a merge.

lib/vanagon/engine/local.rb
MikaelSmith: Potential fix up at https://github.com/puppetlabs/vanagon/pull/295

PR #291, opened 2016-02-17
(PE-13450) AIX upgrade leaves services stopped

shrug:
Currently, due to the way RPM scriptlet ordering works, on AIX
we are stopping the service *after* we restart the service on upgrades.

This makes a couple of changes.

First, make the service stop in the %preun conditional on package
uninstall, like the rest of the platforms.

Second, moves the upgrade %post actions to %postun, as that's more likely where
we actually want those to happen.

Unfortunately, neither change will help users who already have a version
of P-A installed, as the preun/postun scripts are run from the *previous*/
already installed package during upgrades, but this should fix the problem
going forward.

### DIFF
--- a/resources/rpm/project.spec.erb
+++ b/resources/rpm/project.spec.erb
@@ -149,11 +149,6 @@ fi
 if [ $1 -eq 1 ] ; then
   <%= get_postinstall_actions("install") %>
 fi
-
-# Run postinstall scripts on upgrade if defined
-if [ $1 -eq 2 ] ; then
-  <%= get_postinstall_actions("upgrade") %>
-fi
 <%- end -%>
 <%- get_services.each do |service| -%>
   # switch based on systemd vs systemv vs smf vs aix
@@ -205,6 +200,10 @@ fi
       /usr/bin/rmssys -s <%= service.name -%> > /dev/null 2>&1 || :
       /usr/sbin/rmitab <%= service.name -%> > /dev/null 2>&1 || :
     fi
+    # Run postinstall scripts on upgrade if defined
+    if [ $1 -eq 1 ] ; then
+      <%= get_postinstall_actions("upgrade") %>
+    fi
   <%- end -%>
 <%- end -%>
 
@@ -233,7 +232,9 @@ fi
       chkconfig --del <%= service.name %> || :
     fi
   <%- elsif @platform.servicetype == "aix" -%>
-    /usr/bin/stopsrc -s <%= service.name -%> > /dev/null 2>&1 || :
+    if [ $1 = 0 ]; then
+      /usr/bin/stopsrc -s <%= service.name -%> > /dev/null 2>&1 || :
+    fi
   <%- end -%>
 <%- end -%>
 
--- a/resources/rpm/project.spec.erb
+++ b/resources/rpm/project.spec.erb
@@ -118,10 +118,10 @@ done
 %pre
 # Save state so we know later if this is an upgrade or an install
 mkdir -p %{_localstatedir}/lib/rpm-state/%{name}
-if [ $1 -eq 1 ] ; then
+if [ "$1" -eq 1 ] ; then
   touch %{_localstatedir}/lib/rpm-state/%{name}/install
 fi
-if [ $1 -eq 2 ] ; then
+if [ "$1" -gt 1 ] ; then
   touch %{_localstatedir}/lib/rpm-state/%{name}/upgrade
 fi
 
@@ -132,12 +132,12 @@ fi
 <%- end -%>
 
 # Run preinstall scripts on install if defined
-if [ $1 -eq 1 ] ; then
+if [ "$1" -eq 1 ] ; then
   <%= get_preinstall_actions("install") %>
 fi
 
 # Run preinstall scripts on upgrade if defined
-if [ $1 -eq 2 ] ; then
+if [ "$1" -gt 1 ] ; then
   <%= get_preinstall_actions("upgrade") %>
 fi
 
@@ -146,7 +146,7 @@ fi
 <%- if @platform.is_aix? || (@platform.is_el? && @platform.os_version.to_i == 4) -%>
 ## EL-4 and AIX RPM don't have %posttrans, so we'll put them here
 # Run postinstall scripts on install if defined
-if [ $1 -eq 1 ] ; then
+if [ "$1" -eq 1 ] ; then
   <%= get_postinstall_actions("install") %>
 fi
 <%- end -%>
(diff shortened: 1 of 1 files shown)


shrug:
After a puppet-agent build:

```
$ rpm -qp --scripts puppet-agent-1.3.5.47.g5d2b9c8-1.aix7.1.ppc.rpm
preinstall scriptlet (using /bin/sh):
# Save state so we know later if this is an upgrade or an install
mkdir -p /opt/freeware/var/lib/rpm-state/puppet-agent
if [ $1 -eq 1 ] ; then
  touch /opt/freeware/var/lib/rpm-state/puppet-agent/install
fi
if [ $1 -eq 2 ] ; then
  touch /opt/freeware/var/lib/rpm-state/puppet-agent/upgrade
fi


# Run preinstall scripts on install if defined
if [ $1 -eq 1 ] ; then
  :
fi

# Run preinstall scripts on upgrade if defined
if [ $1 -eq 2 ] ; then
  :
fi
postinstall scriptlet (using /bin/sh):
## EL-4 and AIX RPM don't have %posttrans, so we'll put them here
# Run postinstall scripts on install if defined
if [ $1 -eq 1 ] ; then
  :
fi
  # switch based on systemd vs systemv vs smf vs aix
  #
    /usr/bin/mkssys -p /opt/puppetlabs/puppet/bin/ruby -s puppet -u root -a '/opt/puppetlabs/bin/puppet agent --no-daemonize' -w 7 -S -n 15 -f 9 > /dev/null  2>&1 || :
    if /bin/grep -q "^puppet:2:once:startsrc$" /etc/inittab; then
    /usr/sbin/rmitab puppet > /dev/null 2>&1 || :
    fi
    /usr/sbin/mkitab "puppet:2:once:/usr/bin/startsrc -s puppet" > /dev/null 2>&1 || :
  # switch based on systemd vs systemv vs smf vs aix
  #
    /usr/bin/mkssys -p /opt/puppetlabs/puppet/bin/ruby -s mcollective -u root -a '/opt/puppetlabs/puppet/bin/mcollectived --config=/etc/puppetlabs/mcollective/server.cfg --no-daemonize' -w 7 -S -n 15 -f 9 > /dev/null  2>&1 || :
    if /bin/grep -q "^mcollective:2:once:startsrc$" /etc/inittab; then
    /usr/sbin/rmitab mcollective > /dev/null 2>&1 || :
    fi
    /usr/sbin/mkitab "mcollective:2:once:/usr/bin/startsrc -s mcollective" > /dev/null 2>&1 || :
  # switch based on systemd vs systemv vs smf vs aix
  #
    /usr/bin/mkssys -p '/opt/puppetlabs/puppet/bin/pxp-agent' -a '--foreground' -s pxp-agent -u root -w 7 -S -n 15 -f 9 > /dev/null  2>&1 || :
    if /bin/grep -q "^pxp-agent:2:once:startsrc$" /etc/inittab; then
    /usr/sbin/rmitab pxp-agent > /dev/null 2>&1 || :
    fi
    /usr/sbin/mkitab "pxp-agent:2:once:/usr/bin/startsrc -s pxp-agent" > /dev/null 2>&1 || :
preuninstall scriptlet (using /bin/sh):
# Run pre-uninstall scripts on upgrade if defined
if [ $1 -eq 1 ] ; then
  :
fi

# Run pre-uninstall scripts on removal if defined
if [ $1 -eq 0 ] ; then
  :
fi

    if [ $1 = 0 ]; then
      /usr/bin/stopsrc -s puppet > /dev/null 2>&1 || :
    fi
    if [ $1 = 0 ]; then
      /usr/bin/stopsrc -s mcollective > /dev/null 2>&1 || :
    fi
    if [ $1 = 0 ]; then
      /usr/bin/stopsrc -s pxp-agent > /dev/null 2>&1 || :
    fi
postuninstall scriptlet (using /bin/sh):
# Run post-uninstall scripts on upgrade if defined
if [ $1 -eq 1 ] ; then
  :
fi

# Run post-uninstall scripts on removal if defined
if [ $1 -eq 0 ] ; then
  :
fi

  # switch based on systemd vs systemv vs smf vs aix
  #
    if  [ $1 -eq 0 ]; then
      /usr/bin/rmssys -s puppet > /dev/null 2>&1 || :
      /usr/sbin/rmitab puppet > /dev/null 2>&1 || :
    fi
    # Run postinstall scripts on upgrade if defined
    if [ $1 -eq 1 ] ; then
      :
    fi
  # switch based on systemd vs systemv vs smf vs aix
  #
    if  [ $1 -eq 0 ]; then
      /usr/bin/rmssys -s mcollective > /dev/null 2>&1 || :
      /usr/sbin/rmitab mcollective > /dev/null 2>&1 || :
    fi
    # Run postinstall scripts on upgrade if defined
    if [ $1 -eq 1 ] ; then
      :
    fi
  # switch based on systemd vs systemv vs smf vs aix
  #
    if  [ $1 -eq 0 ]; then
      /usr/bin/rmssys -s pxp-agent > /dev/null 2>&1 || :
      /usr/sbin/rmitab pxp-agent > /dev/null 2>&1 || :
    fi
    # Run postinstall scripts on upgrade if defined
    if [ $1 -eq 1 ] ; then
      :
    fi
```

resources/rpm/project.spec.erb
```diff
@@ -233,7 +232,9 @@ fi
       chkconfig --del <%= service.name %> || :
     fi
   <%- elsif @platform.servicetype == "aix" -%>
-    /usr/bin/stopsrc -s <%= service.name -%> > /dev/null 2>&1 || :
+    if [ $1 = 0 ]; then
```
McdonaldSeanp: Why is this conditional "=" while the above is using -eq ?

shrug: Because consistency is the last refuge of the unimaginative.

Or, you know, because we're awesome :)

mckern: `=` is strict string equality. But in shell, basically everything is a string. No strong feeling for "correctness" but maybe consistency?

McdonaldSeanp: Yeah, consistency seems like an idea we should get behind

mckern: bowls

underscorgan:
:+1:

resources/rpm/project.spec.erb
```diff
@@ -118,10 +118,10 @@ done
 %pre
 # Save state so we know later if this is an upgrade or an install
 mkdir -p %{_localstatedir}/lib/rpm-state/%{name}
-if [ $1 -eq 1 ] ; then
+if [ "$1" -eq 1 ] ; then
```
stahnma: This seems odd to me to compare a string to the number. Is this the best way to do this?

mckern: @shrug and I had some back and forth about how to do this, and it worked out to:

```
if you're comparing numbers in bash, always use -eq
if you're comparing strings in bash, always use =
that should work F*** EVERYWHERE
```

As for comparing a quoted variable to an unquoted value, the bottom line is "quote the interpolated variable regardless of type" because if something goes sideways, `[ "" -eq 1 ]` is valid, whereas `[ -eq 1]` is not.

mckern:
Hey @shrug, this needs a rebase

shrug:
Done

PR #293, opened 2016-02-26
(MAINT) Specify ruby version in README

glennsarti:
The readme is vague in its description of what a recent version is. This commit
explicitly details the minimum version required of 2.1.x.  This is due to gem
 dependencies.

PR #297, opened 2016-03-01
(MAINT) Specify ruby version

stahnma:
The readme is vague in its description of what a recent version is. This commit
explicitly details the minimum version required of 2.1.x.  This is due to gem
 dependencies.

 Obsoletes #293


stahnma:
Merging as this is a copy of #293 but updated.

PR #294, opened 2016-02-26
(MAINT) Update gitignore to be more inline with other Puppetlabs projects

glennsarti:
This commit updates the gitignore to be more inline with other Puppetlabs
projects e.g. ignoring .bundle and Gemfile.local

PR #298, opened 2016-03-01
(MAINT) Update gitignore

stahnma:
This commit updates the gitignore to be more inline with other Puppetlabs
projects e.g. ignoring .bundle and Gemfile.local

Obsoletes: #294


stahnma:
Merging as this is PR #294 but rebased.

PR #299, opened 2016-03-02
(RE-6640) Filter out service files from harvest

jcoconnor:
This updates the service file handling to for product specific wix
service file fragments. The service .exe still needs to be filtered out
from the harvest, so the list of service .exe's are placed in the filter
file.

The project.filter.xslt file has also been renamed to simply
filter.xslt.  This file is given a generic name as there may be a need
in the future to filter out other files from the harvest.

The nssm.exe file is not being correctly matched in the harvest - this
appears to be an issue with the build itself (the nssm and some other
files are not being dropped in the correct folder). This issue will be
checked, ticketed if necessary and handled in seperate PR.

### DIFF
--- a/lib/vanagon/component/dsl.rb
+++ b/lib/vanagon/component/dsl.rb
@@ -184,7 +184,8 @@ class Vanagon
           # Return here because there is no file to install, just a string read in
           return
         when "windows"
-          # Placeholder for Windows
+          @component.service = OpenStruct.new(:name => service_name, :service_file => service_file)
+          # return here as we are just collecting the name of the service file to put into the harvest filter list.
           return
         else
           fail "Don't know how to install the #{@component.platform.servicetype}. Please teach #install_service how to do this."
--- a/resources/windows/wix/filter.xslt.erb
+++ b/resources/windows/wix/filter.xslt.erb
@@ -15,11 +15,9 @@
   <!-- The list of component service files are collected in an array so that a set of unique names can be extracted -->
   <%- service_files = Array.new -%>
   <%- get_services.each do |service| -%>
-    <%- service.options.each do |service_option| -%>
-      <%- service_files << service_option[:file][:source] -%>
-    <%- end -%>
+    <%- service_files << service.service_file -%>
   <%- end -%>
   <%- service_files.uniq.each do |service_file|  -%>
-        <xsl:template match="wix:Component[wix:File[@Source='<%= service_file %>']]" />
+    <xsl:template match="wix:Component[wix:File[@Source='<%= service_file %>']]" />
   <%- end -%>
 </xsl:stylesheet>
--- a/lib/vanagon/platform/windows.rb
+++ b/lib/vanagon/platform/windows.rb
@@ -183,7 +183,7 @@ class Vanagon
           #   -gg             - Generate GUIDS now
           #   -dr             - Directory reference to root directories (cannot contains spaces e.g. -dr MyAppDirRef)
           #   -sreg           - Suppress registry harvesting.
-          "cd $(tempdir); \"$$WIX/bin/heat.exe\" dir staging -v -ke -indent 2 -cg #{cg_name} -gg -dr #{dir_ref} -t wix/project.filter.xslt -sreg -out wix/#{project.name}-harvest.wxs",
+          "cd $(tempdir); \"$$WIX/bin/heat.exe\" dir staging -v -ke -indent 2 -cg #{cg_name} -gg -dr #{dir_ref} -t wix/filter.xslt -sreg -out wix/#{project.name}-harvest.wxs",
           # Apply Candle command to all *.wxs files - generates .wixobj files in wix directory.
           # cygpath conversion is necessary as candle is unable to handle posix path specs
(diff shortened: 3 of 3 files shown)


McdonaldSeanp:
:+1: looks good to me, has this been tested?

lib/vanagon/component/dsl.rb
```diff
@@ -184,7 +184,8 @@ def install_service(service_file, default_file = nil, service_name = @component.
           # Return here because there is no file to install, just a string read in
           return
         when "windows"
-          # Placeholder for Windows
+          @component.service = OpenStruct.new(:name => service_name, :service_file => service_file)
```
melissa: could you provide an example of what this will look like when a component uses this?

jcoconnor:
I have tested it manually against a set of product service files that I'm developing for the Puppet Agent side PR. It looks good as in it:
1. Creates the filter.xslt file with all 3 service entries.
2. The harvest file actually filters out 2 of the 3 service executables (see note about nssm.exe as to why this isn't being filtered just yet).
3. The harvested file compiles (heat) ok.

The service files don't compile yet, but I'm handling that on the Puppet-Agent side in a separate PR.

lib/vanagon/component/dsl.rb
jcoconnor: Yes sure - do you mean though - as in what do I need to provide ?

McdonaldSeanp:
Can you post a gist of the output of heat along with the names of the service files that were filtered out?

jcoconnor:
Wix Commands running: https://gist.github.com/jcoconnor/285b70615872f1e4e39d
Heat Output file: https://gist.github.com/b671c41dfeca96ca91b9
xslt file: https://gist.github.com/e68149af8d4cdb1caf80

The executables are (without pathnames):
1. ruby.exe
2. rubyw.exe
3. nssm.exe

As noted, you will see that nssm.exe is there - the reason for this is that its actual location in the staging area is incorrect as noted above. Need to check the build against the filelist to see which is the correct version (suspect the staging is incorrect), but thats a separate PR and possibly ticket.

McdonaldSeanp:
Excellent, thank you. I am fully :+1:

lib/vanagon/component/dsl.rb
melissa: like, what's the user going to put into their component file when then need to add a service? What does this look like when it's being used?

melissa: Also, I forget, is `:name` required? Where is it being used?

jcoconnor: ```:name``` isn't required, but I thought it a useful key to leave in there for debugging, etc. etc.
I can add that detail in as a comment here, but wondering if its better to put it elsewhere.
You can also take a look at PR https://github.com/puppetlabs/puppet-agent/pull/514/files

Are you thinking of a comment like:
```# pkg.install_service "SourceDir\\ProgramFiles64Folder\\PuppetLabs\\Puppet\\bin\\ruby.exe"```
and a note that a separate service wix file also needs to be provided.

Comment (McdonaldSeanp) [inline lib/vanagon/component/dsl.rb]: @melissa were you just wondering what the call would look like? there are some examples in the mcollective component of PA, i.e. `pkg.install_service "ext/aio/redhat/mcollective.init", "ext/aio/redhat/mcollective.sysconfig", "mcollective"` and john didn't change anything about this function so it'll still look like that

Comment (melissa) [inline lib/vanagon/component/dsl.rb]: I don't think we need a comment in the code. I was just wondering what it would look like

Comment (McdonaldSeanp) [inline lib/vanagon/component/dsl.rb]: It'll look the same. https://gist.github.com/McdonaldSeanp/0d1973b017161d78396b is a snippet from mcollective, and there will just be another case under windows with filepaths to .exe files

Comment (melissa) [inline lib/vanagon/component/dsl.rb]: So, in windows, it looks like the user, in order to install a service on windows for a component, will put
```
pkg.install_service "SourceDir\\ProgramFiles64Folder\\PuppetLabs\\Puppet\\bin\\ruby.exe"
```
into their component file.

Cool. That's all I was wondering.

Comment (McdonaldSeanp) [inline lib/vanagon/component/dsl.rb]: yup, exactly

Comment (jcoconnor) [inline lib/vanagon/component/dsl.rb]: Sound

File: resources/windows/wix/filter.xslt.erb [hunk 2]
```diff
@@ -15,11 +15,9 @@
...
   <%- service_files = Array.new -%>
   <%- get_services.each do |service| -%>
-    <%- service.options.each do |service_option| -%>
-      <%- service_files << service_option[:file][:source] -%>
-    <%- end -%>
+    <%- service_files << service.service_file -%>
   <%- end -%>
   <%- service_files.uniq.each do |service_file|  -%>
-        <xsl:template match="wix:Component[wix:File[@Source='<%= service_file %>']]" />
+    <xsl:template match="wix:Component[wix:File[@Source='<%= service_file %>']]" />
```
melissa: Oh shoot, the other thing I was thinking about, does this need to be a unique list? What will happen if we have duplicate entries here?

McdonaldSeanp: @melissa We don't have to worry because of the service_files.uniq line above this one correct?

melissa: Ah, cool. I must have skimmed over it just now. Thanks!

PR #300, opened 2016-03-03
(maint) update the WiX toolset

McdonaldSeanp:
This commit updates the light commands to do the following:

* rename "staging" to SourceDir, since light looks specifically
for a directory called SourceDir

* remove the -dr flag from heat.exe, as we want all the defaults
from heat (as we will use these defaults in light.exe)

* add the -srd flag to heat.exe to ignore the source directory
because again, we want to use the defaults in light

* add the -b flag to light.exe to point it to where we are
putting SourceDir (which is in the temp folder)

* remove INSTALLDIR, we are no longer using it as its set in
the defaults.

* remove the componentgroup element from the project .wxs file
as this is created in heat already

* Update the registrykey element to use ForceCreateOnInstall
instead of action, as action is deprecated now

With all these changes light will now successfully compile an
MSI

### DIFF
--- a/lib/vanagon/platform/windows.rb
+++ b/lib/vanagon/platform/windows.rb
@@ -164,7 +164,6 @@ class Vanagon
       def generate_msi_package(project)
         target_dir = project.repo ? output_dir(project.repo) : output_dir
         cg_name = "ProductComponentGroup"
-        dir_ref = "INSTALLDIR"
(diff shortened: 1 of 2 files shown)


shermdog:
I'm light on the background of this, but seems okay to me :)

melissa:
:+1:

PR #302, opened 2016-03-04
(maint) Redo project.wxs

McdonaldSeanp:
This commit refactors project.wxs to remove a component
that was causing problems with the WiX build. The component
doesn't actually need to be removed, only moved outside
the "product" element and put inside a "ComponentGroupRef"
element in another WiX fragment.

Following this commit the WiX compiler will successfully
compile and the sanitary checks done by WiX at the end
of compilation will not fail out. Meaning Vanagon now
successfully completes.


melissa:
:+1: pending travis

PR #303, opened 2016-03-07
(MAINT) Convert filter.wxs.erb file to Unix Mode

jcoconnor:
This fixes an irritant in the ERB translation which leads to multiple
blank lines in the filter file. While not affecting functionality, it
represented an untidyness that "made me sad" in the PDX vernacular.

Although it looks like all lines have changed, in fact all that was done
was to run the :set fileformat=unix in vi.

And yes - I also removed an unneeded space.


McdonaldSeanp:
:+1:

mckern:
Can't argue with sadness.

PR #304, opened 2016-03-07
(MAINT) Fix Directory/DirectoryRef tag error

jcoconnor:
Directory tag on line 13 should be DirectoryRef


McdonaldSeanp:
Trivial rule, this is very simple

PR #305, opened 2016-03-07
(MAINT) Ensure Candle command fails correctly

jcoconnor:
The for; do; done loop that executes the candle command fails to trap
errors. Add a check that causes the loop to exit on failure if any of
the wix files fail compilation.


McdonaldSeanp:
Trivial

PR #307, opened 2016-03-08
(RE-6455) Create directorylist.wxs

McdonaldSeanp:
This commit will add functionality to use the existing
get_services functionality for a project to generate
a directorylist.wxs file that contains the wix structure
of directories containing binaries for the project.

Because more than one service can point to the same binary
the directorylist may contain more than one directory wix
element at the top of the directory tree. In WiX this is
fine as long as any elements that are supposed to point
to the same directory have different IDs and no "name"
attribute.

### DIFF
--- a/lib/vanagon/platform/windows.rb
+++ b/lib/vanagon/platform/windows.rb
@@ -289,6 +289,105 @@ class Vanagon
         File.join("windows", target_repo, @architecture)
       end
 
+      # Generate correctly formatted wix elements that match the
+      # structure of the directory input
+      #
+      # @param services, Array of components services
+      # and optionally:
+      # @return [string] correctly formatted wix element string
+      def generate_wix_dirs(services)
+        directories = []
+        services.map { |svc| directories.push({ :path => svc.service_file, :id => "#{svc.id}BINDIR" }) }
+        # root refers to the root of an n-ary tree (which we are about to make)
+        root = { :children => [] }
+        # iterate over all paths specified and break each one
+        # in to its specific directories. This will generate_wix_dirs
+        # an n-ary tree structure matching the specs from the input
+        directories.each do |dir|
+          # Always start at the beginning
+          curr = root
+          names = strip_path(dir[:path])
+          # The last entry in this list will be the actual file,
+          # which we do not want, we only want it's base path
+          names.pop
+          names.each do |name|
+            curr = insert_child(curr, name)
+          end
+          # at this point, curr will be the top dir, override the id if
+          # id exists
+          curr[:bindir_ids].push(dir[:id])
+        end
+        return generate_wix_from_graph(root)
+      end
+
+      # insert a new object with the name "name" if it doesn't already
+      # exist. Then assign curr to either the new child or the one that
+      # already exists here
+      #
+      # @param [HASH] curr, current object we are on
+      # @param [string] name, name of new object we are to search for and
+      #                 create if necessary
+      def insert_child(curr, name)
+        #The Id field will default to name, but be overridden later
+        new_obj = { :name => name, :id => name, :bindir_ids => [], :children => [] }
+        if (child_index = includes_child(new_obj, curr[:children]))
+          curr = curr[:children][child_index]
+        else
+          curr[:children].push(new_obj)
+          curr = new_obj
+        end
+        return curr
+      end
+
+      # strip and split the directory path into single names
+      # @param [string] path string of directory
+      def strip_path(path)
+        if path.include?("/") || path.include?("\\")
+          # The regex in the last part of this if warrants some
+          # explanation. Specifically it matches any combinations
+          # of any letters, then the : char, then finally either
+          # the char / or the char \. it's meant to parse out drive
+          # roots on windows
+          if path.start_with?("/") || path.start_with?("\\") || path.start_with?("SourceDir") || path =~ (/([A-Za-z])*\:(\/|\\)/)
+            path = path.sub(/\/|\\|([A-Za-z])*\:(\/|\\)|(\/|\\)?(SourceDir)(\/|\\)?/, '')
+          end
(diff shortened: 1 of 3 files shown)


McdonaldSeanp:
This PR contains some extra work that I am using to help with some collaboration between John and I which is why it has the don't merge label. all of the work in the actual windows.rb and windows_spec.rb is ready to go however, and should be reviewed

jcoconnor:
@McdonaldSeanp Sean - this looks good - will discuss offline after I pull down these changes and check them through.

resources/windows/wix/project.wxs.erb
```diff
@@ -25,6 +25,7 @@
       <!-- We can add all components by referencing this one thing -->
       <ComponentGroupRef Id="ProductComponentGroup" />
       <ComponentGroupRef Id="RegistryComponentGroup" />
+      <ComponentRef Id="service_nssm" />
```
jcoconnor: As discussed in e-mail - suggest we add componentrefs.wxs.erb to pick these up from the services.

jcoconnor:
@McdonaldSeanp I have tested this against my branch, and while it compiles ok, I think it is going to the component level, which is incorrect.
This is the Wix that is generated (have added indentation - may not be quite correct):
```

  <Fragment>
    <DirectoryRef Id='TARGETDIR'>
      <Directory Name="ProgramFiles64Folder" Id="ProgramFiles64Folder">
            <Directory Name="PuppetLabs" Id="PuppetLabs">
                  <Directory Name="Puppet" Id="Puppet">
                        <Directory Name="puppet" Id="puppet">
                              <Directory Name="bin" Id="bin">
                                    <Directory Name="ruby.exe" Id="ruby.exe">
                                    <Directory Id="PUPPETBINDIR" />
                                    </Directory>
                                    <Directory Name="rubyw.exe" Id="rubyw.exe">
                                          <Directory Id="MARIONETTECOLLECTIVEBINDIR" />
                                          </Directory>
                                          <Directory Name="nssm.exe" Id="nssm.exe">
                                                <Directory Id="PXPAGENTBINDIR" />
                                                </Directory>
                                          </Directory>
                                    </Directory>
                              </Directory>
                        </Directory>
                  </Directory>

    </DirectoryRef>
  </Fragment>
</Wix>


```

McdonaldSeanp:
Changes have been pushed up

McdonaldSeanp:
There is no longer extra work in here for collaboration, this is ready for merge

lib/vanagon/platform/windows.rb
```diff
@@ -289,6 +289,105 @@ def output_dir(target_repo = "")
...
 
+      # Generate correctly formatted wix elements that match the
+      # structure of the directory input
+      #
+      # @param services, Array of components services
+      # and optionally:
+      # @return [string] correctly formatted wix element string
+      def generate_wix_dirs(services)
+        directories = []
+        services.map { |svc| directories.push({ :path => svc.service_file, :id => "#{svc.id}BINDIR" }) }
```
melissa: `service_file` corresponds to `"SourceDir\\#{settings[:base_dir]}\\#{settings[:company_id]}\\#{settings[:product_id]}\\puppet\\bin\\nssm.exe"`, right?

McdonaldSeanp: yes

melissa: So, well get `:path => "SourceDir\\#{settings[:base_dir]}\\#{settings[:company_id]}\\#{settings[:product_id]}\\puppet\\bin\\nssm.exe"` and `:id => PXPAGENTBINDER`?

If we pass it a file, will it just take the parent directory and assign the ID to that? How does it know which ID to put? It looks like were going to have `PXPAGENTBINDIR`, `PUPPETBINDIR`, and `MARIONETTECOLLECTIVEBINDIR` all correspond to the same directory, if I'm still following everything?

McdonaldSeanp: Yes, So what it does is strip off SourceDir, because SourceDir corresponds to TARGETDIR with is the directoryRef already existing in directorylist.wxs.erb, then it will strip off the last entry in the path (in this case nssm.exe) because we want it's source directory. Then you'll notice that https://github.com/McdonaldSeanp/vanagon/blob/feature/RE-6455/dirlist/lib/vanagon/platform/windows.rb#L379L381 puts several child elements (without "name" attributes) under the actual bin directory. Per the WiX documentation, if you add a directory element without a name under another directory element with a name, both elements will reference the same directory with different Ids, which is what we want.

jcoconnor:
@McdonaldSeanp Running the build at the moment - the directorylist.wxs file in the work directory looks correct this time.
After attempting an install, I can see the 3 service executables in ```"C:\Program Files\PuppetLabs\Puppet\puppet\bin"```
(services don't start, but this isn't a surprise as there are quite a few other things missing).
(The install also fails at this point and unwinds, which is actually correct behaviour).
I also checked the .msi using ``InstEdit``` and can see the 3 services there.

So this looks quite encouraging.

/cc @melissa

lib/vanagon/platform/windows.rb
jcoconnor: Thats correct Melissa - in this case, they all are in the same directory, but the structure allows for a service file to be placed outside of the main bin directory.
Having multiple ID's point to the same actual directory doesn't appear to be a problem for the MSI.

McdonaldSeanp: For example, if both mco and puppet had the same bin dir, at the point in the wix file where we have the bin directory it will look like: 

`<Directory Name="bin" Id="bin">
    <Directory Id="MARIONETTECOLLECTIVEBINDIR" />
    <Directory Id="PUPPETBINDIR" />
  </Directory>`

jcoconnor:
:+1: Pending Travis

melissa:
:+1:

PR #308, opened 2016-03-08
(RE-6440) Add Service Component References

jcoconnor:
The service component reference is generated from "service_name" using
the Wix rules, i.e. only legal characters are Alphanumeric and _.
This is used to generate a table of service references to ensure the
service components are loaded into the MSI.

### DIFF
--- a/lib/vanagon/component/dsl.rb
+++ b/lib/vanagon/component/dsl.rb
@@ -184,7 +184,7 @@ class Vanagon
           # Return here because there is no file to install, just a string read in
           return
         when "windows"
-          @component.service = OpenStruct.new(:name => service_name, :service_file => service_file)
+          @component.service = OpenStruct.new(:name => service_name, :id => "#{service_name.gsub(/([^A-Za-z0-9])/, '').upcase}", :service_file => target_service_file)
           # return here as we are just collecting the name of the service file to put into the harvest filter list.
           return
         else
--- a/resources/windows/wix/componentrefs.wxs.erb
+++ b/resources/windows/wix/componentrefs.wxs.erb
@@ -1,0 +1,11 @@
+<?xml version='1.0' encoding='windows-1252'?>
+<Wix xmlns='http://schemas.microsoft.com/wix/2006/wi' xmlns:util="http://schemas.microsoft.com/wix/UtilExtension">
+
+  <Fragment>
+    <ComponentGroup Id="ServiceListGroup">
+      <%- get_services.each do |service| -%>
+      <ComponentGroupRef Id="Service_<%= service.id %>" />
+      <%- end -%>
+    </ComponentGroup>
+  </Fragment>
+</Wix>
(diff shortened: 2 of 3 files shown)


resources/windows/wix/componentrefs.wxs.erb
```diff
@@ -0,0 +1,11 @@
+<?xml version='1.0' encoding='windows-1252'?>
+<Wix xmlns='http://schemas.microsoft.com/wix/2006/wi' xmlns:util="http://schemas.microsoft.com/wix/UtilExtension">
+
+  <Fragment>
+    <ComponentGroup Id="ServiceListGroup">
+      <%- get_services.each do |service| -%>
+      <ComponentGroupRef Id="Service_<%= service.id %>" />
```
McdonaldSeanp: This should be just ComponentRef, not ComponentGroupRef, since the service itself is not a component group, just a single component

jcoconnor: I have changed the service files - there is a bit of indirection, but it covers the multiple service definitions that we have in puppet.rb
So GroupRef is actually correct here.

McdonaldSeanp:
:+1:

lib/vanagon/component/dsl.rb
```diff
@@ -184,7 +184,7 @@ def install_service(service_file, default_file = nil, service_name = @component.
           # Return here because there is no file to install, just a string read in
           return
         when "windows"
-          @component.service = OpenStruct.new(:name => service_name, :service_file => service_file)
+          @component.service = OpenStruct.new(:name => service_name, :id => "#{service_name.gsub(/([^A-Za-z0-9])/, '').upcase}", :service_file => target_service_file)
```
underscorgan: Where is target_service_file defined for windows? I'm not seeing it anywhere?

jcoconnor: Uh - oh - so thats where the problem is. Stupid Typo.
It should be service_file - fix coming up quickly.
Tx @underscorgan

jcoconnor: Btw - I may drop :name out of this once @McdonaldSeanp and I get our two changes synced up together.

McdonaldSeanp: Oops, looks like I need to review a little harder

jcoconnor: I think the merge has gone in, so I'll do the target_XXX fix as MAINT once the windows/vanagon call is over.

mckern: When the fix goes in, you can drop the capture (`()`) around the square brackets in the regex pattern used in the `gsub`. Nothing is being captured, so dropping those will shave off a little work for the regex engine.

```ruby
>> "SpecialString-0945_ABC".gsub(/[^A-Za-z0-9]/, '').upcase
=> "SPECIALSTRING0945ABC"
```

jcoconnor: Dang - the PR is up - will check this tho and modify it.

PR #309, opened 2016-03-08
(MAINT) Fix service_file for windows service def

jcoconnor:
This corrects a Typo that was spotted in PR #308
The service_file argument is incorrect.
Have also removed :name as this is no longer needed for either service
table or associated directory list creation (see #307)

### DIFF
--- a/lib/vanagon/component/dsl.rb
+++ b/lib/vanagon/component/dsl.rb
@@ -184,7 +184,7 @@ class Vanagon
           # Return here because there is no file to install, just a string read in
           return
         when "windows"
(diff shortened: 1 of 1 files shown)


McdonaldSeanp:
As noted by @mckern in https://github.com/puppetlabs/vanagon/pull/308#discussion-diff-55396790 , drop the ( ) from the regex so it's just `/[^A-Za-z0-9]/`

jcoconnor:
Have pushed up edit recommended by @mckern

McdonaldSeanp:
I am :+1: on this after travis passes, I would merge myself on the trivial rule but the last time I wanted to merge too quickly I missed something and we had to make this PR.

underscorgan:
:+1: pending travis

PR #311, opened 2016-03-09
(RE-6551) Make vanagon builds retry on errors

shrug:


lib/vanagon/driver.rb
```diff
@@ -76,15 +76,16 @@ def run
...
+      Vanagon::Utilities.retry_with_timeout(3, 3600) do
+        install_build_dependencies
+        @project.fetch_sources(@workdir)
+        @project.make_makefile(@workdir)
+        @project.make_bill_of_materials(@workdir)
+        @project.generate_packaging_artifacts(@workdir)
+        @engine.ship_workdir(@workdir)
+        @engine.dispatch("(cd #{@engine.remote_workdir}; #{@platform.make})")
+        @engine.retrieve_built_artifact
+      end
```
melissa: should each of these things be in a retry block? Is it just going to try everything all over again if there's a failure? Like, if the `dispatch` step fails, are we going to re-fetch all the sources?

shrug: I was just writing a note to that effect, wanted to talk over with the team. Wrap each item? break them down into smaller chinks?

McdonaldSeanp: Yeah, It seems to me these should all have their own separate retries

shrug: also - should the retry count and timeout be set in the project?

McdonaldSeanp: I am +1 for those being set at the project level

melissa: I would ultimately prefer that things that are inclined to fail have a retry (like cloning the github repo or downloading the source archive), but that also would be a lot of work to track that whole thing down

mckern: I don't expect each item in this block to be idempotent, but what actually happens if `generate_packaging_artifacts` or `dispatch` fails halfway through? What happens if this block fails halfway through? This implementation would reset to the beginning of the block and rerun everything from the start.

stahnma:
What normally fails here? I would imagine fetch sources fails 10x more than anything else. Maybe just start with that and see if that offer relief? Retrying building the project doesn't make sense if nothing will have changed.

lib/vanagon/driver.rb
```diff
@@ -76,14 +76,17 @@ def run
...
       @project.fetch_sources(@workdir)
       @project.make_makefile(@workdir)
       @project.make_bill_of_materials(@workdir)
       @project.generate_packaging_artifacts(@workdir)
       @engine.ship_workdir(@workdir)
-      @engine.dispatch("(cd #{@engine.remote_workdir}; #{@platform.make})")
+      Vanagon::Utilities.retry_with_timeout(3, 3600) do
+        @engine.dispatch("(cd #{@engine.remote_workdir}; #{@platform.make})")
+      end
```
mckern: I had some thoughts about making retries a configurable thing that could be defined per project or component. Something like this (and this has NOT been tested)

```ruby
attr_accessor :timeout
attr_accessor :retry_count

def retry(&block)
  Vanagon::Utilities.retry_with_timeout(retry_count, timeout) { yield }
end
private_method: retry
#  Use as: retry { @engine.dispatch("(cd #{@engine.remote_workdir}; #{@platform.make})") }
```

McdonaldSeanp: I was actually thinking something similar but had no idea how to mock it up, I am very +1 to something like this

shrug:
Closing in favor of PR 316

PR #312, opened 2016-03-09
(maint) add get_service , update install_service

McdonaldSeanp:
this adds get_service funcitonality
to the project object to facilitate grabbing
specific values needed by the service files from
vanagon so the user can grab the values
without having to know anything about them

Also move around some of the ComponentGroupRef
stuff to remove some of that functionality from the user
side.


jcoconnor:
@McdonaldSeanp Have just tested this - looks good.
It also builds ok without problems.
:+1:

lib/vanagon/project.rb
```diff
@@ -101,6 +101,15 @@ def get_replaces
       replaces.flatten.uniq
     end
 
+    def get_component(name)
+      @components.each do |component|
+        if component.name == name
+          return component
+        end
+      end
+      return nil
```
stahnma: No need for return.

lib/vanagon/project.rb
```diff
@@ -101,6 +101,15 @@ def get_replaces
       replaces.flatten.uniq
     end
 
+    def get_service(name)
```
melissa: On line 224, we have
```
     def get_services
       @components.map(&:service).flatten.compact
     end
```

Would it be worth it to use this method rather than adding a new one? It probably makes for cleaner code down the road if we make a new method...

Could you add documentation to this?

McdonaldSeanp: I think adding the new method makes the calls in the service files way cleaner. And i'll add comments

resources/windows/wix/project.wxs.erb
```diff
@@ -25,7 +25,9 @@
       <!-- We can add all components by referencing this one thing -->
       <ComponentGroupRef Id="ProductComponentGroup" />
       <ComponentGroupRef Id="RegistryComponentGroup" />
-      <ComponentGroupRef Id="ServiceListGroup" />
+      <%- get_services.each do |service| -%>
+      <ComponentGroupRef Id="<%= service.component_group %>" />
```
melissa: If we have an ID listed here, but not a corresponding ID in a service file, will we get an error?

McdonaldSeanp: Yup, WiX will error out almost immediately

melissa: coolio

lib/vanagon/project.rb
```diff
@@ -101,6 +101,22 @@ def get_replaces
...
+    # Grabs a specific service based on which name is passed in
+    # note that if the name is wrong or there was no
+    # @component.install_service call in the component, this
+    # will return nil
+    #
+    # @param [string] name of service to grab
+    # @return [@component.service obj] specific service
+    def get_service(name)
+      @components.each do |component|
+        if component.name == name
```
melissa: What happens if this is run on a component that doesn't have a name? I think we'll get a error for an undefined method. That also might be fine, since I think every other platform adds name to the service object.

lib/vanagon/component/dsl.rb
```diff
@@ -184,7 +184,11 @@ def install_service(service_file, default_file = nil, service_name = @component.
           # Return here because there is no file to install, just a string read in
           return
         when "windows"
-          @component.service = OpenStruct.new(:id => "#{service_name.gsub(/[^A-Za-z0-9]/, '').upcase}", :service_file => service_file)
+          @component.service = OpenStruct.new(\
+            :bindir_id => "#{service_name.gsub(/[^A-Za-z0-9]/, '').upcase}BINDIR", \
+            :service_file => service_file, \
+            :component_group => "#{service_name.gsub(/[^A-Za-z0-9]/, '')}Components"\
```
melissa: Hmm, for extra clarity, it might be nice to have this be `component_group_id`

melissa: Also, should it be plural or singular?

melissa:
:+1: just small nits, but they're not blockers

lib/vanagon/project.rb
McdonaldSeanp: I'm not sure what you're getting at, I'm fairly certain you cannot make a component without a name https://github.com/puppetlabs/vanagon/blob/master/lib/vanagon/component.rb#L44

lib/vanagon/component/dsl.rb
McdonaldSeanp: singular, there is one component group for many components

melissa: the suffix of the id is currently `Components`

lib/vanagon/project.rb
melissa: Yeah, you can definitely make services without names. But not the user, someone who's editing vanagon. If this method is called on a service object that doesn't have `name` for some reason, it's going to fail with `undefined method `name'`. I'm thinking that's on us, and this is fine as is.

melissa: Aw man, I'm goofy. Ignore me.

McdonaldSeanp:
Builds complete with the new changes

jcoconnor:
@McdonaldSeanp Sean - can we reverse this commit.
Need the componetgroups.wxs.erb file - I do have logic for why we included it.

PR #314, opened 2016-03-10
(RE-6849) UI/Localisation Vanagon side updates

jcoconnor:
Vanagon side changes to introduce UI files:
1. UI reference to ensure code is compiled into MSI
2. Fix to light.exe to all localisation reference.

This PR also includes [RE-6757](https://tickets.puppetlabs.com/browse/RE-6757) 

See Puppet Agent PR [#568](https://github.com/puppetlabs/puppet-agent/pull/568)

### DIFF
--- a/resources/windows/wix/componentgroup.wxs.erb
+++ b/resources/windows/wix/componentgroup.wxs.erb
@@ -1,0 +1,33 @@
+<?xml version="1.0" encoding="windows-1252"?>
+<Wix xmlns="http://schemas.microsoft.com/wix/2006/wi" xmlns:util="http://schemas.microsoft.com/wix/UtilExtension">
+
+  <!-- Default component groups that are to be included in this project.
+       This can be copied in the project specific area to add additional component groups.
+
+       Note that a number of the groups below are "pseudo-groups" that are necessary to
+       force the Wix Linker (light) to include fragments (e.g. FragmentProperties).
+
+       A UI reference is can also be included in the project specific area if required.
+       Otherwise, the project will default here to using the minimal UI (WixUI_Minimal). -->
+
+  <Fragment>
+    <ComponentGroup Id="MainComponentGroup">
+      <!-- We can add all components by referencing this one thing -->
+      <ComponentGroupRef Id="AppComponentGroup" />
+      <ComponentGroupRef Id="AppDataComponentGroup" />
+      <ComponentGroupRef Id="RegistryComponentGroup" />
+      <%- get_services.each do |service| -%>
+      <ComponentGroupRef Id="<%= service.component_group_id %>" />
+      <%- end -%>
+      <!-- All of these Include refs are expected to be present -->
+      <ComponentGroupRef Id="FragmentProperties" />
+      <ComponentGroupRef Id="FragmentSequences" />
+      <ComponentGroupRef Id="FragmentCustomActions" />
+    </ComponentGroup>
+
+    <UI>
+      <UIRef Id="WixUI_Minimal"/>
+    </UI>
+
+  </Fragment>
+</Wix>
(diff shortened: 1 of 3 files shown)


resources/windows/wix/directorylist.wxs.erb
```diff
@@ -2,8 +2,28 @@
...
+    </DirectoryRef>
+
+    <DirectoryRef Id="INSTALLDIR">
+      <Directory Name="puppet" Id="puppet">
+        <Directory Name="bin" Id="bin">
+          <Directory Id="PUPPETBINDIR" />
+          <Directory Id="MARIONETTECOLLECTIVEBINDIR" />
+          <Directory Id="PXPAGENTBINDIR" />
+        </Directory>
+      </Directory>
```
McdonaldSeanp: For anyone looking this over, this is temporary work that @jcoconnor did to get his own stuff to work while I refactor the directorylist vanagon work to reflect some changes. This will change at some point

resources/windows/wix/project.wxs.erb
```diff
@@ -20,6 +20,10 @@
       <!--  We are discussing a MAINT PR to move the componentfilegroup to seperate file
             This would allow project to specify additional fragments if needed -->
 
+    <UI>
+      <UIRef Id="WixUI_MainInstallDialog"/>
```
jpogran: I really think this should not be in vanagon hardcoded. WiX allows any of the provided UI dialog sets, a custom one provided by the user, or none specified (in which case it uses the Minimal set). If we hardcode this here with this name, any user **has** to create a custom WiX dialog set **and** name it this, **even** if they don't want a custom dialog set.

I think this can be better expressed by:

```
<UIRef Id="<% settings[:custom_dialog] %>"/> unless setting[:custom_dialog].nil?
```
or something else similar.

jcoconnor: @jpogran Will work at a solution for this tomorrow.
What do you think of the alternative suggestion of putting in the componentgroups.wxs in the project specific area ?

/cc @melissa @McdonaldSeanp

jcoconnor:
@McdonaldSeanp @jpogran @melissa Have updated this and drawn in [RE-6757](https://tickets.puppetlabs.com/browse/RE-6757) so that the generic/vanagon solution uses the minimal UI, whereas the Puppet-Agent uses the full PA UI.

resources/windows/wix/componentgroup.wxs.erb
```diff
@@ -0,0 +1,33 @@
...
+      <ComponentGroupRef Id="<%= service.component_group_id %>" />
+      <%- end -%>
+      <!-- All of these Include refs are expected to be present -->
+      <ComponentGroupRef Id="FragmentProperties" />
+      <ComponentGroupRef Id="FragmentSequences" />
+      <ComponentGroupRef Id="FragmentCustomActions" />
+    </ComponentGroup>
+
+    <UI>
+      <UIRef Id="WixUI_Minimal"/>
```
jpogran: This line isn't strictly necessary, as WiX will use Minimal if nothing is set. That's why I suggested the setting for the name. It's also not required to have `<UI>` element.

jcoconnor: James - I agree that its not strictly necessary, but thought it worthwhile to quote as the default.

McdonaldSeanp: I also agree, this is more for documentation for the user

McdonaldSeanp:
Has this been through a build and it works?

jcoconnor:
Yes - tested it this morning and it builds and installs correctly - or at least as correctly as we have seen up to now.

McdonaldSeanp:
Excellent, I am :+1:

resources/windows/wix/componentgroup.wxs.erb
jpogran: I disagree, but two votes out vote me.

PR #316, opened 2016-03-11
(RE-6551) Make vanagon builds retry on errors

shrug:
Currently, vanagon builds will occasionally fail on
what turn out to be transient, short-lived issues.

This wraps the main build in the retry method to try
to reduce the impact of transient infrastructure issues.
As set, will retry up to 3 times with a 1-hour timeout.

### DIFF
--- a/lib/vanagon/driver.rb
+++ b/lib/vanagon/driver.rb
@@ -25,10 +25,7 @@ class Vanagon
       @platform = Vanagon::Platform.load_platform(platform, File.join(@@configdir, "platforms"))
       @project = Vanagon::Project.load_project(project, File.join(@@configdir, "projects"), @platform, components)
       @project.settings[:skipcheck] = options[:skipcheck]
-      @@logger = Logger.new('vanagon_hosts.log')
-      @@logger.progname = 'vanagon'
-      @timeout = @project.timeout || 3600
-      @retry_count = @project.retry_count || 3
+      loginit('vanagon_hosts.log')
 
       # If a target has been given, we don't want to make any assumptions about how to tear it down.
       engine = 'base' if target
@@ -119,8 +116,16 @@ class Vanagon
     end
 
     def retry_task(&block)
+      @timeout = @project.timeout || 3600
+      @retry_count = @project.retry_count || 3
       Vanagon::Utilities.retry_with_timeout(@retry_count, @timeout) { yield }
     end
     private :retry_task
+
+    def loginit(logfile)
+      @@logger = Logger.new(logfile)
+      @@logger.progname = 'vanagon'
+    end
+    private :loginit
   end
 end
--- a/lib/vanagon/driver.rb
+++ b/lib/vanagon/driver.rb
@@ -115,6 +115,9 @@ class Vanagon
       raise e
     end
 
+    # Retry the provided block, use the retry count and timeout
+    # values from the project, if available, otherwise use some
+    # sane defaults.
     def retry_task(&block)
       @timeout = @project.timeout || 3600
       @retry_count = @project.retry_count || 3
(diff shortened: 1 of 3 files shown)


mckern:
We should figure out how to rub a little spec test on this.

shrug:
Open to suggestions on that, outside of just making sure that setting a value for retry_count and timeout in the project flow through

lib/vanagon/driver.rb
```diff
@@ -76,14 +79,13 @@ def run
...
-      install_build_dependencies
+      retry_task { install_build_dependencies }
       @project.fetch_sources(@workdir)
       @project.make_makefile(@workdir)
       @project.make_bill_of_materials(@workdir)
       @project.generate_packaging_artifacts(@workdir)
       @engine.ship_workdir(@workdir)
-      @engine.dispatch("(cd #{@engine.remote_workdir}; #{@platform.make})")
+      retry_task { @engine.dispatch("(cd #{@engine.remote_workdir}; #{@platform.make})") }
       @engine.retrieve_built_artifact
```
melissa: Should we retry this too in case something weird happens like a connection is accidentally dropped?

mckern: How do you feel about adding that when we know that this first retry is working the way we hoped it would?

melissa: Totes acceptable

mckern:
@shrug We could test this by writing a spec test to ensure that `#run` tries to build the same number of times as whatever is specified in `#timeout`, and that it times out after `#timeout` value. I can help with that.

mckern:
Stupid percieved complexity:

```
Offenses:
lib/vanagon/driver.rb:16:5: C: Metrics/PerceivedComplexity: Perceived complexity for initialize is too high. [9/7]
    def initialize(platform, project, options = { :configdir => nil, :target => nil, :engine => nil, :components => nil, :skipcheck => false })
    ^^^
41 files inspected, 1 offense detected
```

We can fix this up.

shrug:
The underlying method (Vanagon::Utilities.retry_with_timeout) has some decent tests already that take care of validating the number of retries. Testing the timeout might be finicky but we can work up something.

geoffnichols:
:+1:

lib/vanagon/driver.rb
```diff
@@ -76,14 +76,13 @@ def run
...
-
-      install_build_dependencies
+      retry_task { install_build_dependencies }
       @project.fetch_sources(@workdir)
       @project.make_makefile(@workdir)
       @project.make_bill_of_materials(@workdir)
       @project.generate_packaging_artifacts(@workdir)
       @engine.ship_workdir(@workdir)
-      @engine.dispatch("(cd #{@engine.remote_workdir}; #{@platform.make})")
+      retry_task { @engine.dispatch("(cd #{@engine.remote_workdir}; #{@platform.make})") }
```
mckern: I was thinking about this; should we assume that all engines get this part retried or should it happen per-engine? I'm open to trying it this way; I was just thinking about interface standardization vs. RE convenience earlier today.

shrug: For right now, I think it's safe to say that we want this everywhere, at least the two tasks that are currently targeted. In the future, I think adding by-engine and/or by-project configuration knobs makes sense though.

As it stands, I don't think it hurts to have (if you don't want retries set them to '1' in the project) and it's a good place to start experimenting and see where to go next.

MikaelSmith:
Looks like retries will generate a warning. Any plans to detect those and log how often and why retries are needed?

shrug:
@MikaelSmith Yep, that's going to be an important piece so that we don't ignore problems that might be masked. Right now we don't have a detailed plan in place, but that's on the agenda. For now, this will go to the jenkins console logs, so in the short term I think we'll be parsing those and sending the results to elasticsearch and graphite.

PR #318, opened 2016-03-14
Directorylist fixes

McdonaldSeanp:
This updated DirectoryList and the heat commands to create the correct directory structure in the PA MSI

### DIFF
--- a/lib/vanagon/platform/windows.rb
+++ b/lib/vanagon/platform/windows.rb
@@ -163,9 +163,25 @@ class Vanagon
       # @return [Array] list of commands required to build an msi package for the given project from a tarball
       def generate_msi_package(project)
         target_dir = project.repo ? output_dir(project.repo) : output_dir
-        cg_name = "ProductComponentGroup"
-        wix_extensions =  "-ext WiXUtilExtension -ext WixUIExtension"
-        candle_flags =  "-dPlatform=#{@architecture} -arch #{@architecture} #{wix_extensions}"
+        wix_extensions = "-ext WiXUtilExtension -ext WixUIExtension"
+        # Heat command documentation at: http://wixtoolset.org/documentation/manual/v3/overview/heat.html
+        #   dir <directory> - Traverse directory to find all sub-files and directories.
+        #   -ke             - Keep Empty directories
+        #   -cg             - Component Group Name
+        #   -gg             - Generate GUIDS now
+        #   -srd            - Suppress root element generation, we want to reference one of the default root elements
+        #                     INSTALLDIR or APPDATADIR in the directorylist.wxs file, not a newly generated one.
+        #   -sreg           - Suppress registry harvesting.
+        #   -dr             - Root DirectoryRef to point all components to
+        #   -var            - Replace "SourceDir" in the @source attributes of all components with a preprocessor variable
+        app_heat_flags = " -dr INSTALLDIR -v -ke -indent 2 -cg AppComponentGroup -gg -srd -t wix/filter.xslt -sreg -var var.AppSourcePath "
+        app_source_path = "SourceDir/#{project.settings[:base_dir]}/#{project.settings[:company_id]}/#{project.settings[:product_id]}"
+        appdata_heat_flags = "  -dr APPDATADIR -v -ke -indent 2 -cg AppDataComponentGroup -gg -srd -t wix/filter.xslt -sreg -var var.AppDataSourcePath "
+        appdata_source_path = "SourceDir/CommonAppDataFolder"
+        # Candle.exe preprocessor vars are required due to the above double run of heat.exe, both runs of heat use
+        # preprocessor variables
+        candle_preprocessor = "-dAppSourcePath=\"#{app_source_path}\" -dAppDataSourcePath=\"#{appdata_source_path}\""
+        candle_flags = "-dPlatform=#{@architecture} -arch #{@architecture} #{wix_extensions}"
         # Enable verbose mode for the moment (will be removed for production)
         # localisation flags to be added
         light_flags = "-v -cultures:en-us #{wix_extensions}"
@@ -174,21 +190,16 @@ class Vanagon
           "mkdir -p $(tempdir)/{SourceDir,wix/wixobj}",
           "#{@copy} -r wix/* $(tempdir)/wix/",
           "gunzip -c #{project.name}-#{project.version}.tar.gz | '#{@tar}' -C '$(tempdir)/SourceDir' --strip-components 1 -xf -",
-          # Run the Heat command in a single pass
-          # Heat command documentation at: http://wixtoolset.org/documentation/manual/v3/overview/heat.html
-          #   dir <directory> - Traverse directory to find all sub-files and directories.
-          #   -ke             - Keep Empty directories
-          #   -cg             - Component Group Name
-          #   -gg             - Generate GUIDS now
-          #   -srd            - Suppress root element generation, we want to reference the default root element
-          #                     TARGETDIR in the project.wxs file, not a newly generated one.
-          #   -sreg           - Suppress registry harvesting.
-          "cd $(tempdir); \"$$WIX/bin/heat.exe\" dir SourceDir -v -ke -indent 2 -cg #{cg_name} -gg -srd -t wix/filter.xslt -sreg -out wix/#{project.name}-harvest.wxs",
+          "cd $(tempdir); \"$$WIX/bin/heat.exe\" dir #{app_source_path} #{app_heat_flags} -out wix/#{project.name}-harvest-app.wxs",
+          "cd $(tempdir); \"$$WIX/bin/heat.exe\" dir #{appdata_source_path} #{appdata_heat_flags}  -out wix/#{project.name}-harvest-programdata.wxs",
           # Apply Candle command to all *.wxs files - generates .wixobj files in wix directory.
           # cygpath conversion is necessary as candle is unable to handle posix path specs
-          "cd $(tempdir)/wix/wixobj; for wix_file in `find $(tempdir)/wix -name \'*.wxs\'`; do \"$$WIX/bin/candle.exe\" #{candle_flags} $$(cygpath -aw $$wix_file) || exit 1; done",
+          # the preprocessor variables AppDataSourcePath and ApplicationSourcePath are required due to the -var input to the heat
+          # runs listed above.
+          "cd $(tempdir)/wix/wixobj; for wix_file in `find $(tempdir)/wix -name \'*.wxs\'`; do \"$$WIX/bin/candle.exe\" #{candle_flags} #{candle_preprocessor} $$(cygpath -aw $$wix_file) || exit 1; done",
           # run all wix objects through light to produce the msi
           # the -b flag simply points light to where the SourceDir location is
+          # -loc is required for the UI localization it points to the actual localization .wxl
           "cd $(tempdir)/wix/wixobj; \"$$WIX/bin/light.exe\" #{light_flags} -b $$(cygpath -aw $(tempdir)) -out $$(cygpath -aw $(workdir)/output/#{target_dir}/#{msi_package_name(project)}) *.wixobj",
           ]
       end
@@ -289,35 +300,47 @@ class Vanagon
         File.join("windows", target_repo, @architecture)
       end
 
-      # Generate correctly formatted wix elements that match the
-      # structure of the directory input
+      # Generate the underlying directory structure of
+      # any binary files referenced in services. note that
+      # this function does not generate the structure of
+      # the installation directory, only the structure above it.
       #
-      # @param services, Array of components services
-      # and optionally:
+      # @param services, list of all services in a project
+      # @param project, actual project we are creating the directory structure for
+      def generate_service_bin_dirs(services, project)
+        # All service files will need a directory reference
+        items = services.map do |svc|
+          {
+            :path => strip_and_format_path(svc.service_file, project),
+            :element_to_add => "<Directory Id=\"#{svc.bindir_id}\" />\n"
+          }
+        end
+        generate_wix_dirs(items)
+      end
+
+      # Generate correctly formatted wix elements that match the
+      # structure of the itemized input
+      #
+      # @param items
       # @return [string] correctly formatted wix element string
-      def generate_wix_dirs(services)
-        directories = []
-        services.map { |svc| directories.push({ :path => svc.service_file, :bindir_id => svc.bindir_id }) }
+      def generate_wix_dirs(items)
         # root refers to the root of an n-ary tree (which we are about to make)
         root = { :children => [] }
         # iterate over all paths specified and break each one
         # in to its specific directories. This will generate_wix_dirs
         # an n-ary tree structure matching the specs from the input
-        directories.each do |dir|
+        items.each do |item|
           # Always start at the beginning
           curr = root
-          names = strip_path(dir[:path])
-          # The last entry in this list will be the actual file,
-          # which we do not want, we only want it's base path
-          names.pop
+          names = item[:path].split(File::SEPARATOR)
           names.each do |name|
             curr = insert_child(curr, name)
           end
           # at this point, curr will be the top dir, override the id if
           # id exists
-          curr[:bindir_ids].push(dir[:bindir_id])
+          curr[:elements_to_add].push(item[:element_to_add])
         end
-        return generate_wix_from_graph(root)
+        generate_wix_from_graph(root)
       end
 
       # insert a new object with the name "name" if it doesn't already
@@ -329,42 +352,33 @@ class Vanagon
       #                 create if necessary
       def insert_child(curr, name)
         #The Id field will default to name, but be overridden later
-        new_obj = { :name => name, :id => name, :bindir_ids => [], :children => [] }
-        if (child_index = includes_child(new_obj, curr[:children]))
+        new_obj = { :name => name, :id => name, :elements_to_add => [], :children => [] }
+        if (child_index = index_of_child(new_obj, curr[:children]))
           curr = curr[:children][child_index]
         else
           curr[:children].push(new_obj)
           curr = new_obj
         end
-        return curr
+        curr
       end
 
-      # strip and split the directory path into single names
+      # strip the leading install root and the filename from the service path
+      # and replace any \ with /
+      #
       # @param [string] path string of directory
-      def strip_path(path)
-        if path.include?("/") || path.include?("\\")
-          # The regex in the last part of this if warrants some
-          # explanation. Specifically it matches any combinations
-          # of any letters, then the : char, then finally either
-          # the char / or the char \. it's meant to parse out drive
-          # roots on windows
-          if path.start_with?("/") || path.start_with?("\\") || path.start_with?("SourceDir") || path =~ (/([A-Za-z])*\:(\/|\\)/)
-            path = path.sub(/\/|\\|([A-Za-z])*\:(\/|\\)|(\/|\\)?(SourceDir)(\/|\\)?/, '')
-          end
-          names = path.split(/\/|\\/)
-        end
-        return names
+      # @param [@project] project object
+      def strip_and_format_path(path, project)
+        formatted_path = path.gsub(/\\/, "\/")
+        path_regex = /\/?SourceDir\/#{project.settings[:base_dir]}\/#{project.settings[:company_id]}\/#{project.settings[:product_id]}\//
+        File.dirname(formatted_path.sub(path_regex, ''))
       end
 
       # Find if child element is the same as one of
       # the old_children elements, return that child
-      def includes_child(new_child, old_children)
-        old_children.each_with_index do |curr_old_child, index|
-          return index if curr_old_child[:name] == new_child[:name]
-        end unless old_children.empty?
-        return nil
+      def index_of_child(new_child, old_children)
+        return nil if old_children.empty?
+        old_children.index { |child| child[:name] == new_child[:name] }
       end
-
 
       # Recursively generate wix element structure
       #
@@ -375,9 +389,9 @@ class Vanagon
         unless root[:children].empty?
           root[:children].each do |child|
             string += ("<Directory Name=\"#{child[:name]}\" Id=\"#{child[:id]}\">\n")
-            unless child[:bindir_ids].empty?
-              child[:bindir_ids].each do |bindir_id|
-                string += ("<Directory Id=\"#{bindir_id}\" />\n")
+            unless child[:elements_to_add].empty?
+              child[:elements_to_add].each do |element|
+                string += element
               end
             end
             string += generate_wix_from_graph(child)
@@ -385,7 +399,7 @@ class Vanagon
           end
           return string
         end
-        return ''
+        string
       end
 
       # Constructor. Sets up some defaults for the windows platform and calls the parent constructor
--- a/resources/windows/wix/directorylist.wxs.erb
+++ b/resources/windows/wix/directorylist.wxs.erb
@@ -1,9 +1,17 @@
 <?xml version='1.0' encoding='windows-1252'?>
 <Wix xmlns='http://schemas.microsoft.com/wix/2006/wi' xmlns:util="http://schemas.microsoft.com/wix/UtilExtension">
-
   <Fragment>
     <DirectoryRef Id='TARGETDIR'>
-      <%= @platform.generate_wix_dirs(self.get_services) %>
+      <Directory Id="<%= settings[:base_dir] %>" >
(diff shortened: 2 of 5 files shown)


jcoconnor:
Nice work
@McdonaldSeanp Pulling this down to continue work on it.
Would prefer if the branch and title referred to actual ticket (RE-6746)

lib/vanagon/platform/windows.rb
```diff
@@ -163,9 +163,22 @@ def generate_nuget_package(project)
...
+        #   dir <directory> - Traverse directory to find all sub-files and directories.
+        #   -ke             - Keep Empty directories
+        #   -cg             - Component Group Name
+        #   -gg             - Generate GUIDS now
+        #   -srd            - Suppress root element generation, we want to reference one of the default root elements
+        #                     INSTALLDIR or APPDATADIR in the directorylist.wxs file, not a newly generated one.
+        #   -sreg           - Suppress registry harvesting.
+        #   -dr             - Root DirectoryRef to point all components to
+        #   -var            - Replace "SourceDir" in the @source attributes of all components with a preprocessor variable
+        application_heat_flags = " -dr INSTALLDIR -v -ke -indent 2 -cg ApplicationComponentGroup -gg -srd -t wix/filter.xslt -sreg -var var.ApplicationSourcePath "
```
jcoconnor: Minor - but would shorten this to app_

lib/vanagon/platform/windows.rb
```diff
@@ -163,9 +163,22 @@ def generate_nuget_package(project)
...
+        application_source_path = "SourceDir/#{project.settings[:base_dir]}/#{project.settings[:company_id]}/#{project.settings[:product_id]}"
+        appdata_heat_flags = "  -dr APPDATADIR -v -ke -indent 2 -cg ApplicationDataComponentGroup -gg -srd -t wix/filter.xslt -sreg -var var.AppDataSourcePath "
```
jcoconnor: Suggest you capture the following as common heat flags ```common_heat_flags="-v -ke -indent 2 -gg -srd -t wix/filter.xslt -sreg```

lib/vanagon/platform/windows.rb
```diff
@@ -174,22 +187,17 @@ def generate_msi_package(project)
...
-          #   dir <directory> - Traverse directory to find all sub-files and directories.
-          #   -ke             - Keep Empty directories
-          #   -cg             - Component Group Name
-          #   -gg             - Generate GUIDS now
-          #   -srd            - Suppress root element generation, we want to reference the default root element
-          #                     TARGETDIR in the project.wxs file, not a newly generated one.
-          #   -sreg           - Suppress registry harvesting.
-          "cd $(tempdir); \"$$WIX/bin/heat.exe\" dir SourceDir -v -ke -indent 2 -cg #{cg_name} -gg -srd -t wix/filter.xslt -sreg -out wix/#{project.name}-harvest.wxs",
+          "cd $(tempdir); \"$$WIX/bin/heat.exe\" dir #{application_source_path} #{application_heat_flags} -out wix/#{project.name}-harvest-app.wxs",
+          "cd $(tempdir); \"$$WIX/bin/heat.exe\" dir #{appdata_source_path} #{appdata_heat_flags}  -out wix/#{project.name}-harvest-programdata.wxs",
```
jcoconnor: :+1: to the reduced verbosity here

lib/vanagon/platform/windows.rb
```diff
@@ -174,22 +187,17 @@ def generate_msi_package(project)
...
           # Apply Candle command to all *.wxs files - generates .wixobj files in wix directory.
           # cygpath conversion is necessary as candle is unable to handle posix path specs
-          "cd $(tempdir)/wix/wixobj; for wix_file in `find $(tempdir)/wix -name \'*.wxs\'`; do \"$$WIX/bin/candle.exe\" #{candle_flags} $$(cygpath -aw $$wix_file) || exit 1; done",
+          # the preprocessor variables AppDataSourcePath and ApplicationSourcePath are required due to the -var input to the heat
+          # runs listed above.
+          "cd $(tempdir)/wix/wixobj; for wix_file in `find $(tempdir)/wix -name \'*.wxs\'`; do \"$$WIX/bin/candle.exe\" #{candle_flags} -dApplicationSourcePath=\"#{application_source_path}\" -dAppDataSourcePath=\"#{appdata_source_path}\" $$(cygpath -aw $$wix_file) || exit 1; done",
```
jcoconnor: Please put the following into the flags assignments above ```-dApplicationSourcePath=\"#{application_source_path}\" -dAppDataSourcePath=\"#{appdata_source_path}\"```

lib/vanagon/platform/windows.rb
```diff
@@ -174,22 +187,17 @@ def generate_msi_package(project)
...
           # run all wix objects through light to produce the msi
           # the -b flag simply points light to where the SourceDir location is
-          "cd $(tempdir)/wix/wixobj; \"$$WIX/bin/light.exe\" #{light_flags} -b $$(cygpath -aw $(tempdir)) -out $$(cygpath -aw $(workdir)/output/#{target_dir}/#{msi_package_name(project)}) *.wixobj",
+          # -loc is required for the UI localization it points to the actual localization .wxl
+          "cd $(tempdir)/wix/wixobj; \"$$WIX/bin/light.exe\" #{light_flags} -b $$(cygpath -aw $(tempdir)) -loc $$(cygpath -aw $(tempdir)/wix/localization/puppet_en-us.wxl) -out $$(cygpath -aw $(workdir)/output/#{target_dir}/#{msi_package_name(project)}) *.wixobj",
```
jcoconnor: The ```-loc $$(cygpath -aw $(tempdir)/wix/localization/puppet_en-us.wxl)``` belongs to the UI changes - would prefer if you didn't include it in this change as I have a specific comment related to it in the UI work.

resources/windows/wix/directorylist.wxs.erb
```diff
@@ -1,9 +1,8 @@
 <?xml version='1.0' encoding='windows-1252'?>
 <Wix xmlns='http://schemas.microsoft.com/wix/2006/wi' xmlns:util="http://schemas.microsoft.com/wix/UtilExtension">
-
   <Fragment>
-    <DirectoryRef Id='TARGETDIR'>
-      <%= @platform.generate_wix_dirs(self.get_services) %>
+    <DirectoryRef Id="TARGETDIR">
+      <%= @platform.generate_wix_dir_structure(self.get_services, self) %>
```
jcoconnor: @McdonaldSeanp I think you would be better putting in the Wix directly for the top level structure and just using the ```get_services``` to list the service exclusions. My reasoning for this is:
1. Its much clearer to have the main structure written out in Wix rather than derived (the directory structure is complex enough without hiding it further :( )
2. What happens if the product has no services - will this code still run correctly.

Using the following segment would work (picked from change 47f522f57508fd6f2bf1458d89a5266afec1a0d2):
```
     <DirectoryRef Id='TARGETDIR'>
      <Directory Id="<%= settings[:base_dir] %>" >
        <Directory Id="<%= settings[:company_id] %>" Name="<%= settings[:company_name] %>">
          <Directory Id='INSTALLDIR' Name="<%= settings[:product_id] %>" />
        </Directory>
      </Directory>
      <Directory Id="CommonAppDataFolder" Name="CommonAppData">
        <Directory Id="<%= settings[:company_id] %>AppData" Name="<%= settings[:company_id] %>">
        </Directory>
      </Directory>
    </DirectoryRef>
```
This segment also corrects:
1. Correct company name for "Puppet Labs"
2. Uses Id only for the AppData and Program[64]Directory

lib/vanagon/platform/windows.rb
```diff
@@ -289,33 +297,44 @@ def output_dir(target_repo = "")
...
+      # based on that list
+      #
+      # @param services, list of all services in a project
+      # @param project, actual project we are creating the directory structure for
+      def generate_wix_dir_structure(services, project)
+        items = []
+        # All service files will need a directory reference
+        services.map { |svc| items.push({ :path => File.dirname(svc.service_file.gsub(/\\/, "/")), :element_to_add => "<Directory Id=\"#{svc.bindir_id}\" />\n" }) }
+        # INSTALLDIR is the root installation directory for the project
+        items.push({ :path => "SourceDir/#{project.settings[:base_dir]}/#{project.settings[:company_id]}/#{project.settings[:product_id]}", :element_to_add => "<Directory Id=\"INSTALLDIR\" />\n" })
```
jcoconnor: ```:company_id``` should be ```:company_name```

You also need to make sure that the ```#{project.settings[:base_dir]}``` does not use the ```"Name"``` field as this cause the ```ProgamFiles64Folder``` directory to be created.

I think it would be better to actually express these directories directly in the Wix - see https://github.com/puppetlabs/vanagon/pull/318/files#r56167505 below

lib/vanagon/platform/windows.rb
```diff
@@ -289,33 +297,44 @@ def output_dir(target_repo = "")
...
+        # APPDATADIR is the root directory for all application data (config files etc.)
+        items.push({ :path => "SourceDir/CommonAppDataFolder", :element_to_add => "<Directory Id=\"APPDATADIR\" />\n" })
```
jcoconnor: You need to make sure that ```CommonAppDataFolder``` does the ```"Name"``` field as as above - again see my suggestion in https://github.com/puppetlabs/vanagon/pull/318/files#r56167505 below

resources/windows/wix/project.wxs.erb
```diff
@@ -17,7 +17,10 @@
       Compressed="yes"
       Platform="<%= @platform.architecture %>" />
 
-      <!--  We are discussing a MAINT PR to move the componentfilegroup to seperate file
```
jcoconnor: The more I look at this file I am convinced that we do need to create a ```componentgroups.wxs.erb``` file, so either:
1. We leave this comment in or
2. Ticket a task for it (1 pointer)

Have created ticket: [RE-6757](https://tickets.puppetlabs.com/browse/RE-6757)

resources/windows/wix/project.wxs.erb
```diff
@@ -25,7 +28,8 @@
 
     <Feature Id="<%= settings[:product_id] %>Runtime" Title="<%= settings[:product_id] %> Runtime" Level="1">
       <!-- We can add all components by referencing this one thing -->
-      <ComponentGroupRef Id="ProductComponentGroup" />
+      <ComponentGroupRef Id="ApplicationComponentGroup" />
```
jcoconnor: Again wondering about shorter names e.g. ```AppCompGroup``` - not essential - just a thought.

McdonaldSeanp:
This has been tested and works with the current PA changes.

jcoconnor:
@McdonaldSeanp I have run a test on this.
Program Files looks fine, but I'm not sure the Application Data directories are quite correct:

1. C:\ProgramData\PuppetLabs is still being created - this appeared to happen after the files installed, so it might an issue with the service startup scripts.
2. Also is "C:\ProgramData\Puppet Labs\PuppetLabs" correct ?

jcoconnor:
@McdonaldSeanp I checked https://github.com/melissa/puppet-specifications/blob/maint/master/windows-path-updates/file_paths.md

The Program Data directory should be: ```C:\ProgramData\PuppetLabs```  (no space).
So apologies if I misdirected you earlier on this.

McdonaldSeanp:
The above has been remedied and tested, this is ready for merge

jcoconnor:
:+1:

lib/vanagon/platform/windows.rb
```diff
@@ -174,21 +189,16 @@ def generate_msi_package(project)
...
-          #   dir <directory> - Traverse directory to find all sub-files and directories.
-          #   -ke             - Keep Empty directories
-          #   -cg             - Component Group Name
-          #   -gg             - Generate GUIDS now
-          #   -srd            - Suppress root element generation, we want to reference the default root element
-          #                     TARGETDIR in the project.wxs file, not a newly generated one.
-          #   -sreg           - Suppress registry harvesting.
-          "cd $(tempdir); \"$$WIX/bin/heat.exe\" dir SourceDir -v -ke -indent 2 -cg #{cg_name} -gg -srd -t wix/filter.xslt -sreg -out wix/#{project.name}-harvest.wxs",
+          "cd $(tempdir); \"$$WIX/bin/heat.exe\" dir #{app_source_path} #{app_heat_flags} -out wix/#{project.name}-harvest-app.wxs",
+          "cd $(tempdir); \"$$WIX/bin/heat.exe\" dir #{appdata_source_path} #{appdata_heat_flags}  -out wix/#{project.name}-harvest-programdata.wxs",
```
melissa: Is it possible a project wont install anything into `appdata_source_path`? If so, what will this code do?

jcoconnor: That could be the case.
Would suggest we ticket this and deal with after end of March.
Puppet-Agent will be using both, so we are ok for the moment.

lib/vanagon/platform/windows.rb
```diff
@@ -163,9 +163,24 @@ def generate_nuget_package(project)
...
+        #   -srd            - Suppress root element generation, we want to reference one of the default root elements
+        #                     INSTALLDIR or APPDATADIR in the directorylist.wxs file, not a newly generated one.
+        #   -sreg           - Suppress registry harvesting.
+        #   -dr             - Root DirectoryRef to point all components to
+        #   -var            - Replace "SourceDir" in the @source attributes of all components with a preprocessor variable
+        app_heat_flags = " -dr INSTALLDIR -v -ke -indent 2 -cg AppComponentGroup -gg -srd -t wix/filter.xslt -sreg -var var.AppSourcePath "
+        app_source_path = "SourceDir/#{project.settings[:base_dir]}/#{project.settings[:company_id]}/#{project.settings[:product_id]}"
+        appdata_heat_flags = "  -dr APPDATADIR -v -ke -indent 2 -cg AppDataComponentGroup -gg -srd -t wix/filter.xslt -sreg -var var.AppDataSourcePath "
+        appdata_source_path = "SourceDir/CommonAppDataFolder"
+        # Candle.exe preprocessor vars are required due to the above double run of
```
mckern: Double run of what?

lib/vanagon/platform/windows.rb
mckern: @jcoconnor Can you take point on getting that ticket opened up? What would the side effects be if the location under `appdata_source_path` was empty?

McdonaldSeanp: @melissa It's likely it would make what is essentially an empty WiX fragment, but again, as John said, this is ticketed for after we finished

lib/vanagon/platform/windows.rb
McdonaldSeanp: Oh, ooops

lib/vanagon/platform/windows.rb
```diff
@@ -289,33 +299,46 @@ def output_dir(target_repo = "")
...
+      # @param project, actual project we are creating the directory structure for
+      def generate_service_bin_dirs(services, project)
+        items = []
+        # All service files will need a directory reference
+        services.map do |svc|
+          items.push({
+            :path => strip_and_format_path(svc.service_file, project),
+            :element_to_add => "<Directory Id=\"#{svc.bindir_id}\" />\n"
+          })
+        end
```
mckern: I think you can lose the empty array initialization here?

```ruby
items = services.map do |svc|
  {
    :path => strip_and_format_path(svc.service_file, project),
    :element_to_add => "<Directory Id=\"#{svc.bindir_id}\" />\n"
  }
end
```

lib/vanagon/platform/windows.rb
jcoconnor: Its not ticketed yet, but I'll make a point of doing it tomorrow morning.

lib/vanagon/platform/windows.rb
```diff
@@ -289,33 +299,46 @@ def output_dir(target_repo = "")
...
         # an n-ary tree structure matching the specs from the input
-        directories.each do |dir|
+        items.each do |item|
           # Always start at the beginning
           curr = root
-          names = strip_path(dir[:path])
-          # The last entry in this list will be the actual file,
-          # which we do not want, we only want it's base path
-          names.pop
+          names = item[:path].split(/\//)
```
mckern: I'd lose the regex here. Consider:

```ruby
item[:path].split(File::SEPARATOR) # (Array)
```

or 

```ruby
Pathname.new(item[:path]).each_filename.to_a # (Array)
```

lib/vanagon/platform/windows.rb
McdonaldSeanp: @mckern it's actually not possible for CommonAppDataFolder to not exist, it's a windows default.

lib/vanagon/platform/windows.rb
```diff
@@ -289,33 +299,46 @@ def output_dir(target_repo = "")
...
           names.each do |name|
             curr = insert_child(curr, name)
           end
           # at this point, curr will be the top dir, override the id if
           # id exists
-          curr[:bindir_ids].push(dir[:bindir_id])
+          curr[:elements_to_add].push(item[:element_to_add])
         end
         return generate_wix_from_graph(root)
```
mckern: I'd lose the explicit `return` on a lot of these methods; that's useful when you're exiting early because the work in a method is already done, but if it's the end of the method you should just call/state the value you want to return. In this case, that's `generate_wix_from_graph(root)` without the explicit `return`.

lib/vanagon/platform/windows.rb
```diff
@@ -339,21 +362,14 @@ def insert_child(curr, name)
...
-          # roots on windows
-          if path.start_with?("/") || path.start_with?("\\") || path.start_with?("SourceDir") || path =~ (/([A-Za-z])*\:(\/|\\)/)
-            path = path.sub(/\/|\\|([A-Za-z])*\:(\/|\\)|(\/|\\)?(SourceDir)(\/|\\)?/, '')
-          end
-          names = path.split(/\/|\\/)
-        end
-        return names
+      # @param [@project] project object
+      def strip_and_format_path(path, project)
+        path_regex = /(\\|\/)?SourceDir(\\|\/)#{project.settings[:base_dir]}(\\|\/)#{project.settings[:company_id]}(\\|\/)#{project.settings[:product_id]}(\\|\/)/
```
mckern: ... that's some regex you got there.

lib/vanagon/platform/windows.rb
melissa: Cool, can you link the ticket here so we know it's been made?

lib/vanagon/platform/windows.rb
```diff
@@ -163,9 +163,24 @@ def generate_nuget_package(project)
...
+        #   -ke             - Keep Empty directories
+        #   -cg             - Component Group Name
+        #   -gg             - Generate GUIDS now
+        #   -srd            - Suppress root element generation, we want to reference one of the default root elements
+        #                     INSTALLDIR or APPDATADIR in the directorylist.wxs file, not a newly generated one.
+        #   -sreg           - Suppress registry harvesting.
+        #   -dr             - Root DirectoryRef to point all components to
+        #   -var            - Replace "SourceDir" in the @source attributes of all components with a preprocessor variable
+        app_heat_flags = " -dr INSTALLDIR -v -ke -indent 2 -cg AppComponentGroup -gg -srd -t wix/filter.xslt -sreg -var var.AppSourcePath "
+        app_source_path = "SourceDir/#{project.settings[:base_dir]}/#{project.settings[:company_id]}/#{project.settings[:product_id]}"
```
melissa: Part of me wants to put checks in to make sure the user has at least these three variables set, so the build fails early. That's not a requirement for this PR though

lib/vanagon/platform/windows.rb
```diff
@@ -163,9 +163,24 @@ def generate_nuget_package(project)
...
+        #   dir <directory> - Traverse directory to find all sub-files and directories.
+        #   -ke             - Keep Empty directories
+        #   -cg             - Component Group Name
+        #   -gg             - Generate GUIDS now
+        #   -srd            - Suppress root element generation, we want to reference one of the default root elements
+        #                     INSTALLDIR or APPDATADIR in the directorylist.wxs file, not a newly generated one.
+        #   -sreg           - Suppress registry harvesting.
+        #   -dr             - Root DirectoryRef to point all components to
+        #   -var            - Replace "SourceDir" in the @source attributes of all components with a preprocessor variable
+        app_heat_flags = " -dr INSTALLDIR -v -ke -indent 2 -cg AppComponentGroup -gg -srd -t wix/filter.xslt -sreg -var var.AppSourcePath "
```
melissa: Where is `var.AppSourcePath` getting set? Is that something provided by wix by default?

jcoconnor: Thats a directive to heat to use that variable as a reference instead of source dir
If you look down, you will see that we actually give it a value for the candle run.

McdonaldSeanp: Yep, var.AppSourcePath is a preprocessor variable we pass to candle when it compiles these files. refer to: https://github.com/McdonaldSeanp/vanagon/blob/feature/RE-6470/allow_config_files_to_stay/lib/vanagon/platform/windows.rb#L182

lib/vanagon/platform/windows.rb
McdonaldSeanp: Yep, these deps are used outside this, so that would be its own PR

McdonaldSeanp:
Retested with all the new changes. Still builds an MSI, still installs correctly.

resources/windows/wix/filter.xslt.erb
```diff
@@ -13,9 +13,15 @@
...
+  <!-- we have to substitue the SourceDir\\etc. etc. for $(var.ProjectSourcePath) in order to facilitate heat -->
+  <!--   being run in two different places, specifically it no longer runs from SourceDir so we needed to pass -->
+  <!--   in ProjectSourcePath to the heat runs. Thus when this filter does its work the source attribute will -->
+  <!--   look like $(var.ProjectSourcePath)/puppet/bin/rubyw.exe for example, not  -->
+  <!--   SourceDir/Program64FilesFolder/Puppet/puppet/bin/ruby.exe -->
   <%- service_files = Array.new -%>
   <%- get_services.each do |service| -%>
-    <%- service_files << service.service_file -%>
+
+    <%- service_files << service.service_file.sub("SourceDir\\#{self.settings[:base_dir]}\\#{self.settings[:company_id]}\\#{self.settings[:product_id]}", "$(var.AppSourcePath)") -%>
```
melissa: Is this a change we should force in the service files rather than here?

melissa: Especially because we're manually specifying these paths anyway

McdonaldSeanp: We can't, I can explain this in way better detail offline of these comments (because it takes forever to explain) but there is a strict restriction that this stays here.

resources/windows/wix/project.wxs.erb
```diff
@@ -17,7 +17,10 @@
       Compressed="yes"
       Platform="<%= @platform.architecture %>" />
 
-      <!--  We are discussing a MAINT PR to move the componentfilegroup to seperate file 
+    <UI>
+      <UIRef Id="WixUI_PuppetInstallDir"/>
```
melissa: Should this ID be more generic?

resources/windows/wix/filter.xslt.erb
McdonaldSeanp: Suffice it to say the service files need to point to the right place, but the wix heat runs also need to point to the right place, and because of that this needs to go here

mckern:
There's some Ruby in here that I think could use a little polish but not enough to block on. :+1: for the Ruby; I can't speak to the WiX components.

resources/windows/wix/project.wxs.erb
McdonaldSeanp: This shouldn't be in this PR, oops

melissa:
I'm hesitant to merge this until after the ticket for `appdata_source_path` has been created...

resources/windows/wix/filter.xslt.erb
McdonaldSeanp: Actually I can explain this here. when this file filters out things in the heat runs, it's not actually anything but an xml filter, so it knows nothing about WiX dirs or anything. All it knows is to filter out the component if it's Source attribute looks like what we are matching. So the problem is, at the point in which the filter is looking for the file, candle has not run yet, so none of the preprocessor variables have been replaced, Thus we have to look for the string of the actual preprocessor variable, not what it will resolve to.

jcoconnor:
@melissa and @mckern Ticket [RE-6761](https://tickets.puppetlabs.com/browse/RE-6766) created for this.

PR #319, opened 2016-03-15
(RE-6743) Vanagon generic registry entries

jcoconnor:
These entries are being over-ridden by a more Puppet-Agent specific set
of registry entries for puppet-agent.

Rather than deleting this file, it is being left in place with comments
to give guidance on how to set registry entries.

This model will also be applied to the properties and other common
files.

### DIFF
--- a/resources/windows/wix/registryEntries.wxs.erb
+++ b/resources/windows/wix/registryEntries.wxs.erb
@@ -1,15 +1,44 @@
 <?xml version='1.0' encoding='windows-1252'?>
 <Wix xmlns='http://schemas.microsoft.com/wix/2006/wi' xmlns:util="http://schemas.microsoft.com/wix/UtilExtension">
+
+  <!--
+    Registry Entries
+    This is a suggested set of entries that can be used to record the Installation Directory for upgrade installs.
+    Any definitions in this will be over-ridden if the same file name is used in the Product Specific resouces directory.
+  -->
   <Fragment>
     <ComponentGroup Id="RegistryComponentGroup">
-        <Component Id="RegistryEntriesArchitectureDependent" Directory="TARGETDIR" Guid="E6D5AF4F-ACC4-4D11-AFCE-299A9ED2152C" Win64="<%= settings[:win64] %>" Permanent="yes">
-          <RegistryKey Root="HKLM" Key="SOFTWARE\<%= settings[:company_name] %>\<%= settings[:product_name] %>" ForceCreateOnInstall="yes" >
-            <RegistryValue Type="integer" Value="0"/>
+        <Component
+          Id="RegistryEntriesArchitectureDependent"
+          Directory="TARGETDIR"
+          Guid="*"
+          Win64="<%= settings[:win64] %>"
+          Permanent="yes">
+
+          <RegistryKey
+            Root="HKLM"
+            Key="SOFTWARE\<%= settings[:company_name] %>\<%= settings[:product_name] %>"
+            ForceCreateOnInstall="yes" >
+
+            <RegistryValue
+              Type="integer"
+              Value="0"/>
             <%- if @platform.architecture == "x64" -%>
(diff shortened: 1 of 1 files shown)


McdonaldSeanp:
Pretty much just verbatim the same stuff from PA, So :+1:

melissa:
Will this mess things up for projects who don't have their own version of `registryEntries.wxs`? Is this generic enough to where a project can use this by default and not have to worry?

resources/windows/wix/registryEntries.wxs.erb
```diff
@@ -1,15 +1,44 @@
...
+          Guid="E6D5AF4F-ACC4-4D11-AFCE-299A9ED2152C"
+          Win64="<%= settings[:win64] %>"
+          Permanent="yes">
+
+          <RegistryKey
+            Root="HKLM"
+            Key="SOFTWARE\<%= settings[:company_name] %>\<%= settings[:product_name] %>"
+            ForceCreateOnInstall="yes" >
+
+            <RegistryValue
```
McdonaldSeanp: What is this registry value for?

McdonaldSeanp:
@melissa yes, this will only add a few registry values that we would use on upgrade that wouldn't be referenced anywhere else on the windows machine. They should be generic enough to no not break anything.

resources/windows/wix/registryEntries.wxs.erb
jcoconnor: Would need @ferventcoder or @Iristyle to verify, but I understand this to create an integer value within the enclosing key - is this the "DefaultValue"?

resources/windows/wix/registryEntries.wxs.erb
```diff
@@ -1,15 +1,44 @@
...
+  -->
   <Fragment>
     <ComponentGroup Id="RegistryComponentGroup">
-        <Component Id="RegistryEntriesArchitectureDependent" Directory="TARGETDIR" Guid="E6D5AF4F-ACC4-4D11-AFCE-299A9ED2152C" Win64="<%= settings[:win64] %>" Permanent="yes">
-          <RegistryKey Root="HKLM" Key="SOFTWARE\<%= settings[:company_name] %>\<%= settings[:product_name] %>" ForceCreateOnInstall="yes" >
-            <RegistryValue Type="integer" Value="0"/>
+        <Component
+          Id="RegistryEntriesArchitectureDependent"
+          Directory="TARGETDIR"
+          Guid="E6D5AF4F-ACC4-4D11-AFCE-299A9ED2152C"
```
melissa: And we decided it's fine to have this GUID for every project that might use these generic files?

melissa:
:+1: pending my one question about the guid. I have a feeling that question was already answered, but it'd be nice to get it in writing

resources/windows/wix/registryEntries.wxs.erb
McdonaldSeanp: Actually, now that I think about it, for a generalized file in vanagon, the guid description at http://wixtoolset.org/documentation/manual/v3/xsd/wix/component.html makes me think maybe this should be "*"

jcoconnor:
@melissa Its a good point - we could have an issue if two products using the same generic set of files could have an issue.
I have changed this to Guid="*" - as per documentation in: http://wixtoolset.org/documentation/manual/v3/xsd/wix/component.html

This value should be a guid that uniquely identifies this component's contents, language, platform, and version. If omitted, the default value is '*' which indicates that the linker should generate a stable guid. Generatable guids are supported only for components with a single file as the component's keypath or no files and a registry value as the keypath. It's also possible to set the value to an empty string to specify an unmanaged component. Unmanaged components are a security vulnerability because the component cannot be removed or repaired by Windows Installer (it is essentially an unpatchable, permanent component). Therefore, a guid should always be specified for any component which contains resources that may need to be patched in the future.

PR #321, opened 2016-03-16
(MAINT) Fix the CommonAppData location

jcoconnor:
Files were being installed in C:\ProgramData\PuppetLabs\PuppetLabs
Move files up one level to C:\ProgramData\PuppetLabs

### DIFF
--- a/lib/vanagon/platform/windows.rb
+++ b/lib/vanagon/platform/windows.rb
@@ -177,7 +177,7 @@ class Vanagon
         app_heat_flags = " -dr INSTALLDIR -v -ke -indent 2 -cg AppComponentGroup -gg -srd -t wix/filter.xslt -sreg -var var.AppSourcePath "
         app_source_path = "SourceDir/#{project.settings[:base_dir]}/#{project.settings[:company_id]}/#{project.settings[:product_id]}"
         appdata_heat_flags = "  -dr APPDATADIR -v -ke -indent 2 -cg AppDataComponentGroup -gg -srd -t wix/filter.xslt -sreg -var var.AppDataSourcePath "
-        appdata_source_path = "SourceDir/CommonAppDataFolder"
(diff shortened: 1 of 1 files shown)


lib/vanagon/platform/windows.rb
```diff
@@ -177,7 +177,7 @@ def generate_msi_package(project)
         app_heat_flags = " -dr INSTALLDIR -v -ke -indent 2 -cg AppComponentGroup -gg -srd -t wix/filter.xslt -sreg -var var.AppSourcePath "
         app_source_path = "SourceDir/#{project.settings[:base_dir]}/#{project.settings[:company_id]}/#{project.settings[:product_id]}"
         appdata_heat_flags = "  -dr APPDATADIR -v -ke -indent 2 -cg AppDataComponentGroup -gg -srd -t wix/filter.xslt -sreg -var var.AppDataSourcePath "
-        appdata_source_path = "SourceDir/CommonAppDataFolder"
+        appdata_source_path = "SourceDir/CommonAppDataFolder/#{project.settings[:company_id]}"
```
McdonaldSeanp: Shouldn't this also require a change in https://github.com/puppetlabs/vanagon/blob/master/resources/windows/wix/directorylist.wxs.erb#L12L14 ?

jcoconnor: Its a context issue - will BJ for a few mins after the Sprint Retro if you have a moment.
If you harvest from the directory above ```PuppetLabs``` which we were doing, you create a sub-directory called ```PuppetLabs``` under it.

McdonaldSeanp: I was wrong, it does not require a change in vanagon and this is in fact the right way to go, before we were not doing things the same between app and appdata, this will make it so we are

McdonaldSeanp:
:+1:

PR #322, opened 2016-03-17
(RE-6520) Specify architecture when generating Debian packages

ScottGarman:
This allows cross-compiled packages to use correct tools (such as the
target arch strip command) when generating the package. When noarch is
set, this becomes a no-op.


McdonaldSeanp:
I like it :+1:

PR #323, opened 2016-03-18
(RE-6739) Load services unconditionally on osx

shrug:
Currently, osx will only load services if the pkgutil check
detects a previous package installed. I think this is failing
because, in the postinstall, the package is no longer installed.

This changes the postinstall to unconditionally load services
on OSX.

Also fixes a whitespace typo in the preinstall script

### DIFF
--- a/resources/osx/postinstall.erb
+++ b/resources/osx/postinstall.erb
@@ -1,7 +1,5 @@
 #!/bin/bash
 
-foundpkg=`/usr/sbin/pkgutil --volume "$3" --pkgs=<%= @identifier-%>.<%= @name -%>`
-
 # Move any new configfiles into place
 <%- get_configfiles.each do |config|
 dest_file = config.path.gsub(/\.pristine$/, '') -%>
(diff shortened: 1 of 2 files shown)


shrug:
Another option here may be to trigger based on the earlier conditions in the postinstall file where it's detecting existing config files.

But is there ever a time we don't want the service scripts to load?

ScottGarman:
I can't think of any instances where we don't want the services to load, so I'm :+1:  on this.

geoffnichols:
Do we want to unconditionally load `pxp-agent`?

resources/osx/postinstall.erb
```diff
@@ -13,12 +11,8 @@ else
 fi
 <%- end -%>
 
-# If we appear to be in an upgrade load services.
-
-if [ -n "$foundpkg" ]; then
 <%- get_services.each do |service| -%>
   /bin/launchctl load -F "<%= service.service_file %>"
```
geoffnichols: Does this need to be a fully qualified path to a plist file?

geoffnichols:
:+1:

heathseals:
That logic was added due to RE-5032 ( where it was reported that loading services, except on upgrades, was an unexpected behavior ).

PR #324, opened 2016-03-20
(RE-6739) Make service load conditional on upgrade for osx

shrug:
In a previous commit, service reloads were made unconditional,
as the test for loading them was failing. This updates the logic
to reload the services if they were loaded at the beginning of
the package operation.

### DIFF
--- a/resources/osx/preinstall.erb
+++ b/resources/osx/preinstall.erb
@@ -1,13 +1,21 @@
 #!/bin/bash
 
-# If we appear to be in an upgrade unload services.
+# Cleanup previous operations
+if [ -e "/tmp/.<%= @name %>.upgrade" ]
+  rm -rf "/tmp/.<%= @name %>.upgrade"
+fi
 
+# If we appear to be in an upgrade unload services.
 foundpkg=`/usr/sbin/pkgutil --volume "$3" --pkgs=<%= @identifier-%>.<%= @name -%>`
 oldbom="/usr/share/doc/<%= @name %>/bill-of-materials"
 
 if [ -n "$foundpkg" ]; then
+  mkdir -p "/tmp/.<%= @name %>.upgrade"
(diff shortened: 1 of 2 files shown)


resources/osx/preinstall.erb
```diff
@@ -1,13 +1,21 @@
 #!/bin/bash
 
-# If we appear to be in an upgrade unload services.
+# Cleanup previous operations
+if [ -e "/tmp/.<%= @name %>.upgrade" ]
+  rm -rf "/tmp/.<%= @name %>.upgrade"
```
mckern: could probably lose the recursive; this shouldn't be a directory, right?

resources/osx/preinstall.erb
```diff
@@ -1,13 +1,21 @@
...
+fi
 
+# If we appear to be in an upgrade unload services.
 foundpkg=`/usr/sbin/pkgutil --volume "$3" --pkgs=<%= @identifier-%>.<%= @name -%>`
 oldbom="/usr/share/doc/<%= @name %>/bill-of-materials"
 
 if [ -n "$foundpkg" ]; then
+  mkdir -p "/tmp/.<%= @name %>.upgrade"
+  chmod 0700 "/tmp/.<%= @name %>.upgrade"
```
mckern: should it be executable, or should it just be readable?

resources/osx/preinstall.erb
shrug: Yep, made it a directory (see the preinstall), then touch a file for each service found by the preinstall

resources/osx/preinstall.erb
shrug: it's a directory, then it touches a file for each service found

mckern: Yeah, I see that; my fault for skimming on the weekend and not **reading** it for reals.

underscorgan:
:+1:

PR #326, opened 2016-03-21
(maint) add windows style version function

McdonaldSeanp:
This function is necessary to facilitate the
windows vanagon work as WiX and windows require
only three digit versions with no non-digit chars.

This will also add rspec for the new function.

### DIFF
--- a/lib/vanagon/platform/windows.rb
+++ b/lib/vanagon/platform/windows.rb
@@ -402,6 +402,14 @@ class Vanagon
         string
       end
 
+      # Grab only the first three values from the version input
+      # and strip off any non-digit characters.\
+      #
+      # @param [string] version, the original version number
+      def windows_style_version(version)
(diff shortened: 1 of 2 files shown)


ScottGarman:
:+1:

lib/vanagon/platform/windows.rb
```diff
@@ -402,6 +402,14 @@ def generate_wix_from_graph(root)
         string
       end
 
+      # Grab only the first three values from the version input
+      # and strip off any non-digit characters.\
+      #
+      # @param [string] version, the original version number
+      def windows_style_version(version)
```
melissa: This should have been 'wix' specific rather than Windows

melissa:
Why do they require only 3 digits? I was under the impression the fourth was ignored, and included for human legibility, especially with testing development builds. Your commit needs more explanation

PR #329, opened 2016-03-22
(RE-6520) Specify arch when generating Debian pkgs

ScottGarman:
This allows cross-compiled packages to use correct tools (such as the
target arch strip command) when generating the package. When
project.noarch is set, this becomes a no-op.

Note: you cannot have a space between the -a and the arch, which
only works for the dpkg-buildpackage in Debian 8 and later.


ScottGarman:
I've tested this on the dynamic vanagon builder, as build #330:

http://jenkins-release.delivery.puppetlabs.net/job/vanagon_generic_job/330/

All the debian platforms passed.

shrug:
:+1:

PR #330, opened 2016-03-22
OSX pre and post scripts should actually be shell

shrug:
The previous fix for the osx install scripts had some
syntax errors.

This fixes those and should actually produce a valid
shell script


shrug:
Tested this specific configuration, installs and upgrades successfully on osx

mckern:
Whoopsie daisy. Merging under the rule of "we should not break stuff."

PR #331, opened 2016-03-22
(RE-6506) Change MSI Product Version to 3 digits.

jcoconnor:
As per the original pxp-agent build, the Version for the MSI should be 3
digits and not 4. Use the new windows_style_version method provided by
PR: https://github.com/puppetlabs/vanagon/pull/326 to use the
first 3 digits only in the ERB project template.

### DIFF
--- a/resources/windows/wix/project.wxs.erb
+++ b/resources/windows/wix/project.wxs.erb
@@ -6,7 +6,7 @@
     Name="<%= settings[:product_name] %>"
     Language="1033"
     Codepage="1252"
-    Version="<%= @version.sub(/\.g[0-9a-z]{7}$/, '') %>"
+    Version="<%= @platform.windows_style_version(@version) %>"
     Manufacturer="<%= settings[:company_name] %>" >
 
     <Package


shermdog:
:+1:

resources/windows/wix/project.wxs.erb
```diff
@@ -6,7 +6,7 @@
     Name="<%= settings[:product_name] %>"
     Language="1033"
     Codepage="1252"
-    Version="<%= @version.sub(/\.g[0-9a-z]{7}$/, '') %>"
+    Version="<%= @platform.windows_style_version(@version) %>"
```
melissa: I didn't see this method go in, but we already have a nuget version method. I'd be more comfortable if this were wix specific, since not everything windows will have the same version requirements

melissa:
Also, the commit title is a little misleading. You might be making it three digits, but the main thing that's happening is you're changing it to pull the version from a defined method

melissa:
Also, the commit title is a little misleading. You might be making it three digits, but the main thing that's happening is you're changing it to pull the version from a defined method

shermdog:
Merged with Euro Power!

melissa:
Also, wasn't the four digit version number intentional? When it's a final tag it will be three digits, but with development builds, it was decided we would add the extra digit even though it would be effectively ignored by the installer. It would be helpful to humans.

And how does pxp-agent dictate the version for puppet -agent should be 3 digits? A sub project shouldn't have that much control over the main project....

jcoconnor:
The concern at the time was that we should match all the logic in the Rake task.
When I was discussing some of the registry issues with @ferventcoder recently the point was made that we should keep to what was in the current puppet-agent build package.

melissa:
So you're confirming that the version change here has nothing to do with pxp-agent?

jcoconnor:
Will revise the initial comment to try and make it a bit clearer.

melissa:
It's already merged in, so there's not much point to that.

The rake task in pftw also does allow four digits
```
  flags['MajorVersion'] = match_data[1]
  flags['MinorVersion'] = match_data[2]
  flags['BuildVersion'] = match_data[3]
  flags['Revision'] = match_data[4] || 0
```

resources/windows/wix/project.wxs.erb
McdonaldSeanp: I am going to throw up a maint commit to change the name to make this clear it's for WiX

McdonaldSeanp:
I was going off what the comment in that rake task said: https://github.com/puppetlabs/puppet_for_the_win/blob/master/tasks/windows/windows.rake#L34

melissa:
Were there errors with the four digit version? I definitely was under the impression that wix would just ignore the fourth entry.

McdonaldSeanp:
Even if it did just ignore the last entry we were going to have to strip the letters anyway.

jcoconnor:
There weren't errors in it.
Windows installer ignores the final digit for upgrade checking, so if you get two packages with for example the following versions 1.4.0.123  and 1.4.0.453, it would treat them as the same version, i.e. (it would refuse to upgrade to 1.4.0.453 if 1.4.0.123 is already installed because it sees it at same as the installed version.

jcoconnor:
Relevant lines from puppet_for_the_win
```Version="$(var.ProductVersionNumber)"``` (puppet.wxs)
```<?define ProductVersionNumber="$(var.MajorVersion).$(var.MinorVersion).$(var.BuildVersion)" ?>``` (puppet.wxi).

This looked quite deliberate to me.
Will put it on the Trello board and ask at the Windows/Vanagon meeting today tho.

**Comment by melissa** [comment]
The code change is fine, I'm totally okay with having three digits in the version number in the wix file. My main point in all of this, is the commit message needs to have some kind of explanation in it. If we ever go back to this, we're not going to have any idea why the change was made.

I was also concerned that method was being used to give the msi a name, but that is not the case.
```
      def msi_package_name(project)
        # Decided to use native project version in hope msi versioning doesn't have same resrictions as nuget
        "#{project.name}-#{project.version}.#{project.release}-#{@architecture}.msi"
      end
```
is used, so that's fine.

jcoconnor:
Is that message in the commit clearer them, or do I still need to tidy it up a bit?

PR #332, opened 2016-03-22
(maint) update version function name

McdonaldSeanp:
windows_style_version is misleading, as
there is more than one way to create a
windows package, this updates the name
to something specific to wix


jcoconnor:
:+1: 
That makes a clearer @McdonaldSeanp

PR #333, opened 2016-03-22
(RE-6470) Remove second heat.exe run / AppData dirs

McdonaldSeanp:
In order to facilitate the static file additions of
AppData to the puppet-agent project, this commit removes
the second run of heat.exe in the CommonAppDataFolder and
removes references of that folder from directorylist.wxs


ScottGarman:
:+1:

PR #334, opened 2016-03-23
(RE-6506) Drop versions.txt into misc directory

jcoconnor:
This is a copy of the bill-of-materials file which is placed into:
C:\Program Files\Puppet Labs\Puppet\misc


jcoconnor:
This is almost working, but want to convert the versions.txt file to DOS mode so it shows up correctly in notepad.

jcoconnor:
Added an awk based Unix2dos conversion so versions.txt file is readable in Windows (notepad)

McdonaldSeanp:
:+1:

PR #335, opened 2016-03-23
(maint) add VANAGON to msi filename

McdonaldSeanp:
As a way to differentiate between pftw MSIs and
vanagon msis (and allow us to put both through pipelines
while we test the cutover) add VANAGON to the vanagon
MSI filenames


Iristyle:
:+1:

PR #337, opened 2016-03-23
(maint) adding a vanagon_devkit

McdonaldSeanp:
#As to not deprecate the current devkit
binary (that others may be using) I have
created a new dev_build that allows the
following:

* @engine.startup no longer has to run
every time, if you specify --target and -r
vanagon will just skip startup with
the assumption this target is an already running
engine. This essentially allows you to "resume"
a vanagon run

* for every step in driver the devkit will
ask if you want to execute the step "y/yes"
skip the step "s/skip" or stop running "n/no"

* The preserve flag is unecessary for the
devkit since it will ask before doing
@engine.teardown

usage for vanagon_devkit is the same as always with
the optional --target doing something different:

bundle exec vanagon_devkit <project> <platform> < --target will
resume an engine>

### DIFF
--- a/bin/dev_build
+++ b/bin/dev_build
@@ -25,4 +25,4 @@ artifact = Vanagon::Driver.new(platform, project, options.merge({ :components =>
 artifact.verbose = true if options[:verbose]
 artifact.resume = true if options[:resume]
 
-artifact.devkit_run
+artifact.dev_run
--- a/lib/vanagon/driver.rb
+++ b/lib/vanagon/driver.rb
@@ -121,7 +121,7 @@ class Vanagon
     # This function is temporary and should be used with care. It is meant to
     # facilitate vanagon development and should be deprecated
     #     Sean M.    3/28/16
-    def devkit_run
+    def dev_run
           # Simple sanity check for the project
       if @project.version.nil? or @project.version.empty?
         raise Vanagon::Error, "Project requires a version set, all is lost."
--- a/lib/vanagon/driver.rb
+++ b/lib/vanagon/driver.rb
@@ -10,16 +10,18 @@ require 'logger'
 class Vanagon
   class Driver
     include Vanagon::Utilities
-    attr_accessor :platform, :project, :target, :workdir, :verbose, :preserve
+    attr_accessor :platform, :project, :target, :workdir, :verbose, :preserve, :resume
     attr_accessor :timeout, :retry_count
 
     def initialize(platform, project, options = { :configdir => nil, :target => nil, :engine => nil, :components => nil, :skipcheck => false })
       @verbose = false
       @preserve = false
+      @resume = options[:resume]
 
       @@configdir = options[:configdir] || File.join(Dir.pwd, "configs")
       components = options[:components] || []
       target = options[:target]
+
       engine = options[:engine] || 'pooler'
 
       @platform = Vanagon::Platform.load_platform(platform, File.join(@@configdir, "platforms"))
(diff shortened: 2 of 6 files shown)


jcoconnor:
Interesting - Have we a documented set of vanagon options?

McdonaldSeanp:
Yeah, the README specifies the arguments available

underscorgan:
I feel like it might make more sense for this to be an extension of the existing devkit instead of an additional tool. I think having two separate and very similarly named things will just cause a lot of confusion down the line, and it is unclear to me why this shouldn't be a flag.

McdonaldSeanp:
I left this as it's own tool simply to keep this as a purely additive PR. I have no idea who uses devkit and if anyone uses it at all, so I wanted to make an only additive PR so things would go faster. if no one uses it I am cool to just change devkit but again, IDK if anyone does use it. As far as it being  a flag I'm only hesitant to do that because I fundamentally changed what happens when you pass devkit the --target flag and I think it's clearer to use exec devkit instead of exec build --devkit.

underscorgan:
If it fundamentally changed what happened with `--target` does it make sense for that to be a different flag? 

[caveat: apologies if these are silly questions. I have not used vanagon devkit much in the past, it's just these appear vaguely similar and I worry about extra overhead of trying to decipher what I should be using.]

If it makes sense for them to be two different tools can the names better differentiate them? `devkit` vs `vanagon_devkit` really is unclear as someone who doesn't use these much.

underscorgan:
There is also some existing devkit documentation in the README, I'd want to see that updated before anything is merged here.

McdonaldSeanp:
Yeah I can change the name, and yes I should likely change the flag too.

lib/vanagon/driver.rb
```diff
@@ -115,6 +117,49 @@ def prepare(workdir = nil)
...
+      if @project.version.nil? or @project.version.empty?
+        raise Vanagon::Error, "Project requires a version set, all is lost."
+      end
+      @workdir = Dir.mktmpdir
+
+      if @devkit_target_specified.nil?
+        @engine.startup(@workdir)
+      end
+
+      puts "Target is #{@engine.target}"
```
mckern: Honest question: wouldn't these prompts make more sense inside of each of these methods themselves, and all `devkit_run` does is enable them? That's a real good use of a mixin.

McdonaldSeanp:
This has been updated. It is meant as only a temporary solution to the issue, and will likely need revision, but as "scaffolding" this will work nicely to allow us to resume runs of vanagon

lib/vanagon/driver.rb
```diff
@@ -50,6 +53,7 @@ def self.logger
       @@logger
     end
 
+
```
shrug: so much whitespace

shrug: room-for-activities

shrug:
:+1:

bin/dev_build
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env ruby
+
+# *****NOTE:
+# This is functionally almost identical to the devkit binary. and should be deprecated
+# in the near future by just updating devkit
+#
+
+load File.expand_path(File.join(File.dirname(__FILE__), "..", "lib", "vanagon.rb"))
```
mckern: This is not a reliable way to require a library. There's no guarantee that this path will actually resolve. What's wrong with `require 'vanagon'`?

McdonaldSeanp: I did not write that, it's a cut over from devkit, I see no problem with that.

lib/vanagon/driver.rb
```diff
@@ -10,16 +10,18 @@
 class Vanagon
   class Driver
     include Vanagon::Utilities
-    attr_accessor :platform, :project, :target, :workdir, :verbose, :preserve
+    attr_accessor :platform, :project, :target, :workdir, :verbose, :preserve, :resume
```
mckern: Not in this PR maybe but we should doc all of these attributes and their accessors. We got some big fat classes nowadays.

McdonaldSeanp: I'll add that to https://tickets.puppetlabs.com/browse/RE-6854

lib/vanagon/driver.rb
```diff
@@ -125,6 +173,28 @@ def retry_task(&block)
...
+    # facilitate vanagon development and should be deprecated
+    #     Sean M.    3/28/16
+    # Ask (and wait) for user input, note that this
+    # is mostly for devkit purposes
+    #
+    # @param [string] prompt, what to display before ask
+    def continue(&block)
+      answer = $stdin.gets.chomp!
+      if answer =~ /^y$|^yes$/
+        puts "y"
```
mckern: Nitpick: is `puts` necessary` here?

lib/vanagon/driver.rb
```diff
@@ -125,6 +173,28 @@ def retry_task(&block)
...
+    # This function is temporary and should be used with care. It is meant to
+    # facilitate vanagon development and should be deprecated
+    #     Sean M.    3/28/16
+    # Ask (and wait) for user input, note that this
+    # is mostly for devkit purposes
+    #
+    # @param [string] prompt, what to display before ask
+    def continue(&block)
+      answer = $stdin.gets.chomp!
+      if answer =~ /^y$|^yes$/
```
mckern: You matches here are not case-insensitive.

lib/vanagon/driver.rb
McdonaldSeanp: no that was debug I'll take it out

lib/vanagon/driver.rb
```diff
@@ -125,6 +173,28 @@ def retry_task(&block)
...
+        puts "y"
+
+        yield
+      elsif answer =~ /^s$|^skip$/
+        return
```
mckern: `return` what? `Nil`? The value of the last evaluation?

McdonaldSeanp: doesn't matter, nil I guess, the function above this expects no return value

mckern: I'd probably move this around slightly to make it a little more useful for end users who might try to use this (even in its "this is gonna die soon" state):

```ruby
def continue(&block)
  # `gets` reads from $stdin by default so we can omit that
  answer = gets.downcase.chomp

  # 100% less regex
  case answer
  when "y", "yes"
    yield
  when "n", "no"
    puts %(Building on "#{@engine.target}" stopped by user)
    exit
  else
    # I didn't see an easy way to get the name of the stage being run, but this
    # has no explicit `return` and provides a clean fall-through in case something
    # goes really, really weird.
    puts "skipping this stage"
  end
end
```

stahnma:
Closing as this is not the path forward.

PR #339, opened 2016-03-29
(RE-6100) OS X Service Postinstall fixup

mckern:
Vanagon should not try to generate postinstall service stanzas
for OS X packages when there are no services to worry about.
Vanagon::Project#get_services will return an empty array if there
are no services, which evaluates to falsey in an `if` statement
so this should be a safe way to prevent this script from accidentally
generating invalid syntax in the Bash if/then block.


shrug:
:+1:

PR #341, opened 2016-03-30
(RE-6100) Rub some booleans on the erb

mckern:
My previous OS X post-install fix was incomplete. This expands the use of the `#has_<values>?` boolean methods to help ensure that we don't accidentally construct empty `if`/`else` stanzas in our pre & post install scripts.

This is needed to help the PE Client Tools .pkg install successfully on OS X.


McdonaldSeanp:
:+1:

PR #343, opened 2016-03-30
(RE-6882) do not copy package Replaces into Conflicts

ScottGarman:
There was no need to follow previous behavior of making Conflicts
include packages specified in Replaces, so we are now keeping each
definition separate.


ScottGarman:
Currently running through the ad-hoc CI pipeline build #105: https://jenkins.puppetlabs.com/view/puppet-agent/view/ad%20hoc/job/platform_puppet-agent_init-van-int_suite-manual-ad-hoc/105/

mckern:
:+1:

PR #345, opened 2016-04-04
Reject empty strings so we dont get bad versions

cwood:
We have tags that use double dashes, so this will make a version of .. so this fails when trying to make the RPM.


mckern:
Hi @cwood,
Thanks for the PR! Can you add one or more spec tests to validate that this is both doing what you expect and doesn't break the existing behavior?

cwood:
Can do.

cwood:
Should be good now.

PR #346, opened 2016-04-05
(maint) remove the checksum for local sources

McdonaldSeanp:
If a source file is local to the vanagon project and
not fetched from http, there is no reason to check the
sum for the file since it's already in a repo and local.

This commit removes the validate step for local files
by making the "verify" function just do nothing for
local files.

### DIFF
--- a/lib/vanagon/component/source/http.rb
+++ b/lib/vanagon/component/source/http.rb
@@ -35,13 +35,18 @@ class Vanagon
         end
 
         # Verify the downloaded file matches the provided sum
+        # Note: This function will not verify the checksum of
+        # a local file, only those files fetched from a
+        # remote source.
         #
         # @raise [RuntimeError] an exception is raised if the sum does not match the sum of the file
         def verify
-          puts "Verifying file: #{@file} against sum: '#{@sum}'"
-          actual = get_md5sum(File.join(@workdir, @file))
-          unless @sum == actual
-            fail "Unable to verify '#{@file}'. Expected: '#{@sum}', got: '#{actual}'"
+          unless @scheme == 'file'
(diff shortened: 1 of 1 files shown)


mckern:
Not sure I agree -- "checksum everything" is not the worst idea, even with local sources. I'm guessing the use-case for this was quickly iterating on local sources (patches, init scripts, etc) and getting caught up by checksum mismatches?

McdonaldSeanp:
Yup, and I disagree with checksum everything b/c we don't do this anywhere else, and it's not a common practice in software engineering in general and this workflow is why. There is no way every time I want to iterate over these sources I want to change a checksum, we don't checksum anything in the vanagon config directory so why would we checksum other local sources?

McdonaldSeanp:
Yes, and I'll update this with a better comment. I don't believe we see any value in checksumming a local file. The file already exists in the repo, so when you pull down the repo you are guaranteed to get the latest, uncorrupted version of the file, if that's not true then we have bigger problems, specifically with git.

shermdog:
Although wouldn't the checksum insure that you haven't accidentally modified the thing locally post pull or also verify that the thing you are putting back up is correct if modified?

mckern:
elephant

We checksum everything because our `http://` handler is also our `file://` handler.
I won't block this PR for this, but we should absolutely document that decision in an almost gory level of detail in the code and the commit message. I'm opening a ticket to address the elephant, and I'll reference this and #262 from it.

McdonaldSeanp:
I can update the comments, that's pretty easy.

McdonaldSeanp:
@shermdog git already does all those things for us.

PR #347, opened 2016-04-05
(maint)(RE-6995) Add a new LocalSource Class

McdonaldSeanp:
This commit adds a new Component::Source child called
LocalSource to handle local files. This functionality
exists in the http child class of source and this commit
removes the functionality from that class and moves it to a
new one. With the new class we can now avoid checksumming
local files by making "verify" a noop function.

The new class comes with spec tests and updated spec
tests for source


McdonaldSeanp:
This has been both spec tested and tested with a windows build of P-A, everything works

lib/vanagon/component/source/http.rb
```diff
@@ -53,6 +53,7 @@ def download
           target_file = File.basename(uri.path)
           puts "Downloading file '#{target_file}' from url '#{@url}'"
 
+
```
stahnma: Why the new line here?

stahnma:
I'd probably reword the commit message. Basically the commit purpose is to add the class. The reason you're making the class is largely due to the sum stuff.

shrug:
:+1:

PR #348, opened 2016-04-06
(ORCH-1293)add remote_workdir flag and support workdir with build

adreyer:
Previously the build command would ignore the workdir this is now fixed.
This commit also adds a remote_workdir flag to allow users to specify
the workdir on the target. This workdir will not be cleaned up since the
main reason to specify a workdir is to preserve it.

### DIFF
--- a/lib/vanagon/driver.rb
+++ b/lib/vanagon/driver.rb
@@ -16,6 +16,9 @@ class Vanagon
     def initialize(platform, project, options = { :configdir => nil, :target => nil, :engine => nil, :components => nil, :skipcheck => false })
       @verbose = false
       @preserve = false
+      @workdir = options[:workdir]
+      remote_workdir = options[:remote_workdir]
+
 
       @@configdir = options[:configdir] || File.join(Dir.pwd, "configs")
       components = options[:components] || []
@@ -32,7 +35,7 @@ class Vanagon
       # Hardware has explicit teardown to unlock the node
       engine = 'hardware' if @platform.build_hosts
       require "vanagon/engine/#{engine}"
-      @engine = Object.const_get("Vanagon::Engine::#{engine.capitalize}").new(@platform, target)
+      @engine = Object.const_get("Vanagon::Engine::#{engine.capitalize}").new(@platform, target, :remote_workdir => remote_workdir)
 
     rescue LoadError => e
       raise Vanagon::Error.wrap(e, "Could not load the desired engine '#{engine}'")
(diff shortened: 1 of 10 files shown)


underscorgan:
@briancain I hear you're planning to test this ... can you let us know how those tests go?

briancain:
@underscorgan - Sounds good, I'll let you know how it goes!

briancain:
It works like I expected and placed all of the projects in the remote workdir I specified!

underscorgan:
Hey @mckern, you had thoughts on this. Can you review/comment?

mckern:
I had it up in a tab I never submitted and I lost it. Let me rewrite it.

lib/vanagon/driver.rb
```diff
@@ -16,6 +16,9 @@ class Driver
     def initialize(platform, project, options = { :configdir => nil, :target => nil, :engine => nil, :components => nil, :skipcheck => false })
       @verbose = false
       @preserve = false
+      @workdir = options[:workdir]
+      remote_workdir = options[:remote_workdir]
```
mckern: Nitpick: I'd continue to assign to the instance variable rather than the accessor (for consistency). It doesn't have to be fixed up before this is merged, but we'll probably fix it up immediately after we merge it.

lib/vanagon/engine/base.rb
```diff
@@ -6,12 +6,14 @@ class Engine
     class Base
       attr_accessor :target, :remote_workdir, :name
 
-      def initialize(platform, target = nil)
+      def initialize(platform, target = nil, opts = {})
```
mckern: Nitpick: Since Vanagon requires Ruby 2.1 or better, we chose not to use the option hash pattern. Instead we used [keyword arguments](https://robots.thoughtbot.com/ruby-2-keyword-arguments). Again, I won't block on it but we're probably going to fix it up immediately after merge.

mckern:
I added a couple general comments about code we'll likely modify after merge, but generally I like this. @adreyer how do you feel about renaming the `:workdir` parameter to something like `:tmpdir` (`:stagingdir`, `:stashdir`, whatever. Something besides `:workdir`) and then changing `:remote_workdir` to `:workdir`? The current `:workdir` param is basically just a Bag of Holding where artifacts are staged before they're shipped to the builder for assembly, and it's not configurable. I'm wondering if that keeps the configuration namespace a little simpler (and introduces less potential namespace collision when digging into the code).

adreyer:
@mckern I'm ok with that. I dont' use --workdir but I expected it to do what I've made --remote_workdir do. Do we know who uses --workdir?

stahnma:
Can we have a link to a public-facing ticket here instead of orch? (I don't think orch is public)

mckern:
@adreyer I think `--workdir` was intended to be more "complete" than it is and I don't know if anyone uses it. But Vanagon is public now and people outside PL are submitting PRs so now I'm a little more gun-shy about the change that I proposed than I was a week and a half ago.

stahnma:
@mckern has recommend closing -- and we'll do this in another fix that's inbound soonish.

PR #349, opened 2016-04-08
(RE-6739) Add service type for Solaris services

shrug:
Currently, the solaris smf service uses a service type,
but the type is not being referenced in the calls to
the smf service handler.

This adds the service type to the service data structure
and uses it in the solaris smf service calls.

### DIFF
--- a/lib/vanagon/component/dsl.rb
+++ b/lib/vanagon/component/dsl.rb
@@ -214,7 +214,7 @@ class Vanagon
         end
 
         # Register the service for use in packaging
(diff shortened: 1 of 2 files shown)


shrug:
```
Changed packages:
puppetlabs.com
  puppet-agent
    None -> 1.4.1.55.490,5.11-1:20160408T182238Z
Services:
  restart_fmri:
    svc:/network/mcollective:*
    svc:/network/puppet:*
    svc:/network/pxp-agent:*
    svc:/system/manifest-import:default
```

ScottGarman:
:+1:

PR #352, opened 2016-04-08
(maint) update the LocalSource name

McdonaldSeanp:
Source::LocalSource is inconsistent with the rest of
the source children names, update this to just be
Source:Local as that's consistent with Source:http and the
like


melissa:
👍 you did a check to see if you missed any invocations of this?

McdonaldSeanp:
Yep

PR #353, opened 2016-04-12
(RE-6525) remove platform option from WiX

McdonaldSeanp:
We use vanagon's erb engine to populate fields
and remove sections based on platform. Thus we
do not need the -dplatform option in the WIX
compiler options.


McdonaldSeanp:
This was tested for both x86 and x64 and seems to not cause any problems.

jcoconnor:
I'm :+1: on this, but please check with @ferventcoder on any other comparisons needed.

McdonaldSeanp:
I've also looked through the MSIs with instedit and couldn't find differences

ferventcoder:
I don't recall what this sets anymore, but I believe it was some variable in puppet.wxi that we used as a determination for certain things being available or not.

ferventcoder:
If that is all done elsewhere then :+1:

PR #355, opened 2016-04-14
(maint) add retry_count to dsl

McdonaldSeanp:
Because of the way we are using dsl objects in vanagon,
if you want an attribute to be accessible you need to
add a corresponding function to the dsl.rb in order to
set the attribute.

project.retry_count had no such function and so it was
not accessible


shrug:
:+1: 

@stahnma and I were discussing changing the default to '1' and overriding it in the project, but that's for another PR

underscorgan:
:+1: pending travis

PR #359, opened 2016-04-15
(maint) add load_engine helper methods

McdonaldSeanp:
This commit adds engine loading helper
methods in order to facilitate a cleaner
and less complex driver init

### DIFF
--- a/lib/vanagon/driver.rb
+++ b/lib/vanagon/driver.rb
@@ -27,15 +27,25 @@ class Vanagon
       @project.settings[:skipcheck] = options[:skipcheck]
       loginit('vanagon_hosts.log')
 
-      # If a target has been given, we don't want to make any assumptions about how to tear it down.
-      engine = 'base' if target
-      # Hardware has explicit teardown to unlock the node
-      engine = 'hardware' if @platform.build_hosts
-      require "vanagon/engine/#{engine}"
-      @engine = Object.const_get("Vanagon::Engine::#{engine.capitalize}").new(@platform, target)
-
+      load_engine(engine, @platform, target)
     rescue LoadError => e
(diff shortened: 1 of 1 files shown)


shrug:
Should be load_engine?

stahnma:
@shrug it doesn't appear to be returning a bool, so the ? wouldn't be idiomatic.

shrug:
Sorry, that was an actual question, not a suggestion for the method name. :) Referring to the PR comment and commit message, which references engine_load, vs the code which is load_engine

McdonaldSeanp:
Oh yeah sure, I can change the name. It was a mindless message

lib/vanagon/driver.rb
```diff
@@ -27,17 +28,35 @@ def initialize(platform, project, options = { :configdir => nil, :target => nil,
...
+        @engine = Vanagon::Engine::Hardware.new(platform, target)
+      when 'docker'
+        @engine = Vanagon::Engine::Docker.new(platform, target)
+      when 'local'
+        @engine = Vanagon::Engine::Local.new(platform, target)
+      when 'pooler'
+        @engine = Vanagon::Engine::Pooler.new(platform, target)
+      else
+        @engine = Vanagon::Engine::Base.new(platform, target)
+      end
```
mckern: This could probably all be collapsed into 

```ruby
def load_engine_object(engine_type, platform, target)
  Object::const_get("Vanagon::Engine::#{engine_type.capitalize}").new(platform, target)
rescue
  fail "No such engine '#{engine_type.capitalize}'"
end
```

mckern: Oh, that's basically what it already did. What's wrong with what it basically already did? Don't say "variable interpolation".

shrug:
:+1:

PR #362, opened 2016-04-18
(maint) fix broken load_engine_object method

McdonaldSeanp:
As pointed out by @cwood , the load_engine_object method is missing an
assignment to engine.


melissa:
👍 pending travis

PR #358, opened 2016-04-15
(PA-341) Add AWS/EC2 engine to Vanagon

cwood:
This is some initial work to use aws with vanagon. I am expecting some review of what else needs be done and some feedback.

PR #363, opened 2016-04-19
It will be nice to set arguments when cloneing like --depth

cwood:
I would like the ability to clone and not get the full history but just the 1st depth.

### DIFF
--- a/lib/vanagon/engine/ec2.rb
+++ b/lib/vanagon/engine/ec2.rb
@@ -7,12 +7,11 @@ class Vanagon
   class Engine
     class Ec2 < Base
       attr_accessor :ami, :key_name, :userdata, :key, :key_name, :shutdown_behavior
-      attr_accessor :subnet_id, :instance_type,
+      attr_accessor :subnet_id, :instance_type
 
       def initialize(platform, target = nil)
         super
-        @name = 'ec2'
-        @platform = platform
+
         @ami = @platform.aws_ami
         @target_user = @platform.target_user
         @subnet_id = @platform.aws_subnet_id
@@ -28,7 +27,11 @@ class Vanagon
         @resource = ::Aws::EC2::Resource.new(client: @ec2)
       end
 
-      def get_userdata()
+      def name
+        'ec2'
+      end
+
+      def get_userdata
         unless @userdata.nil?
           Base64.encode64(ERB.new(@userdata).result(binding))
         end
--- a/spec/lib/vanagon/engine/ec2_spec.rb
+++ b/spec/lib/vanagon/engine/ec2_spec.rb
@@ -1,0 +1,25 @@
+require 'vanagon/engine/ec2'
+require 'vanagon/platform'
+
+describe 'Vanagon::Engine::Ec2' do
+  let(:platform_ec2) do
+    plat = Vanagon::Platform::DSL.new('el-7-x86_64')
+    plat.instance_eval(<<-END)
+      platform 'el-7-x86_64' do |plat|
+        plat.aws_ami 'ami'
+        plat.target_user 'root'
+        plat.aws_subnet_id 'subnet_id'
+        plat.aws_user_data 'user_data'
+        plat.aws_region 'us-west-1'
+        plat.aws_key_name 'vanagon'
+        plat.aws_instance_type 't1.micro'
+        plat.ssh_port '22'
+      end
+    END
+    plat._platform
+  end
+
+  it 'returns "ec2" name' do
+    expect(Vanagon::Engine::Ec2.new(platform_ec2).name).to eq('ec2')
+  end
+end


cwood:
Looks like maybe my spec test never got added in?

lib/vanagon/component/source/git.rb
```diff
@@ -5,20 +5,30 @@ class Component
...
+          @clone_args = clone_args
+        end
+
+
+        def get_clone_args()
+          unless @clone_args
+            ""
+          else
+            @clone_args
+          end
```
underscorgan: Rubocop doesn't like unless/else. That will need to be reworked to be an if/else.

lib/vanagon/component/source/git.rb
```diff
@@ -5,20 +5,30 @@ class Component
...
-        attr_accessor :url, :ref, :workdir, :version, :cleanup
+        attr_accessor :url, :ref, :workdir, :version, :cleanup, :clone_args
 
         # Constructor for the Git source type
         #
         # @param url [String] url of git repo to use as source
         # @param ref [String] ref to checkout from git repo
         # @param workdir [String] working directory to clone into
-        def initialize(url, ref, workdir)
+        def initialize(url, ref, workdir, clone_args)
```
underscorgan: `clone_args` should have a default so it can maintain backwards compatibility.

lib/vanagon/component/source/git.rb
cwood: Sounds good. I didnt test this much

lib/vanagon/component/source/git.rb
cwood: I wanted it to just be nil if it wasnt set. But I can just default it to a empty string

underscorgan: You could also default to `nil`. Mostly other tests were failing with 3 args instead of 4.

cwood: Im not much of a rubiest. I was trying to kwarg clone_args but then I ran into other problems/

cwood:
This actully saved like 10 minutes of my build. So this is really handy to set. Would it be okay to set to --depth to 10 by default?

mckern:
@cwood I'm taking a look at this today, but I can't see a reason why you'd need a depth of more than 1 for building. I'm going to have some comments on this for sure, but I'm focusing on #358 today.

cwood:
yea that was my thought as well.

lib/vanagon/component/source/git.rb
```diff
@@ -5,20 +5,30 @@ class Component
...
             fail "ref parameter is required for the git source"
           end
           @url = url
           @ref = ref
           @workdir = workdir
+          @clone_args = clone_args
+        end
+
+
+        def get_clone_args()
```
mckern: There's already an accessor defined for `clone_args`, so a separate accessor isn't needed.When `get_clone_args` is called in `fetch`, it'll cast to an empty string if it's `nil`.

cwood: Thats fine; Im just from python land where a None will get cast to a string None.

mckern: I could tell you came from python just from the `if`/`else` usage :satisfied:

mckern:
I've been thinking about this some and I'm into the idea of shallow cloning when possible. But I think it needs some more safety bumpers around it -- it's pretty easy to specify a commit, tell Vanagon to make a shallow clone, and then watch git have a mini-meltdown trying to checkout a SHA it doesn't have a reference for. I'm looking at using `git ls-remote origin HEAD` to check whether or not the expected reference to build is also HEAD (which means we could just do a shallow clone by default), along with some other stupid git tricks. I think I'm going to pull in your PR to a branch on my fork and make that work. Any objections?

cwood:
Thats fine with me. Right now what I am doing to deal with the reference not being there is having my CI system give me the branch, and the reference. Then tell git to do --branch branch and then it should be able to checkout that revision

stahnma:
@mckern what needs to happen here to make this real? I really like the idea of depth on the clone operation.

mckern:
@stahnma @cwood can you think of a scenario where depth needs to be greater than 1? I've got an alternate approach but it doesn't store a configurable depth; it just assumes `--depth=1` or full clone (depending on circumstances).

cwood:
I have ran into the issue of a race condition where the branch you are checking out has been updated between. So I have been making it 50 for safety.

mckern:
@stahnma and @cwood what I've been working on is in https://github.com/puppetlabs/vanagon/compare/master...mckern:RE-6200/you-are-i?expand=1

It's overcomplicated for *just* controlling clone depth, but I'm trying to add that functionality into shaving the "how to clone from `http://` or `git://` URLs" problem.

cwood:
Oh wow that seems a lot nicer than my hax. :thumbsup: I am kind of using this in my production build right now; the clone_args anyway.

PR #361, opened 2016-04-18
RFC: (RE-7015) Add cross_compiled utility matcher methods

ScottGarman:
With all the cross-compiling we're now doing, these methods will help us
to refactor the puppet-agent component configs to improve readability.

PR #365, opened 2016-04-22
RFC (RE-7015) Add cross_compiled and cross_compiled_linux platform flags

ScottGarman:
With all the cross-compiling we're now doing, these flags will help us
to refactor the puppet-agent component configs to improve readability.

### DIFF
--- a/lib/vanagon/platform/dsl.rb
+++ b/lib/vanagon/platform/dsl.rb
@@ -198,6 +198,20 @@ class Vanagon
       # Because single vs plural is annoying to remember
       alias_method :build_host, :build_hosts
 
+      # Set a flag whether this platform is cross-compiled
+      #
+      # @param type [Boolean] true if this platform is cross-compiled
+      def cross_compiled(flag)
+        @platform.cross_compiled = true unless flag.nil?
+      end
+
+      # Set a flag whether this is a cross-compiled Linux platform
+      #
+      # @param type [Boolean] true if this is a cross-compiled Linux platform
+      def cross_compiled_linux(flag)
+        @platform.cross_compiled_linux = true unless flag.nil?
+      end
+
       # Set the name of this platform as the vm pooler expects it
       #
       # @param name [String] name of the target template to use from the vmpooler
--- a/lib/vanagon/platform.rb
+++ b/lib/vanagon/platform.rb
@@ -3,7 +3,8 @@ require 'vanagon/platform/dsl'
 class Vanagon
   class Platform
     attr_accessor :make, :servicedir, :defaultdir, :provisioning, :num_cores, :tar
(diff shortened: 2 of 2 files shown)


stahnma:
Could you put on a more detailed commit message? I don't know if we should reference puppet-agent in the commit message. Also, why are we doing this at this level as opposed to another way?

McdonaldSeanp:
IMO I think this approach makes more sense for cross-compiled platforms, specifically because with https://github.com/puppetlabs/vanagon/pull/361/files the user has less control over which platforms are cross-compiled. For me, the more user control the better, and adding the extra platform.cross_compiled line to the platform config is trivial.

lib/vanagon/platform/dsl.rb
```diff
@@ -198,6 +198,20 @@ def build_hosts(*args)
       # Because single vs plural is annoying to remember
       alias_method :build_host, :build_hosts
 
+      # Set a flag whether this platform is cross-compiled
+      #
+      # @param type [Boolean] true if this platform is cross-compiled
+      def cross_compiled(flag)
```
McdonaldSeanp: What is the reason behind the parameter passed to this function? if cross_compiled is just meant to set the cross_compiled flag it could have no parameter.

mckern: In theory, project files should be treated like a DSL and not like raw Ruby. In practice, they're straight up raw Ruby. I'm not opposed to making this a simple thing that gets called in a project like this:

```ruby
project "puppet-agent" do |proj|
  cross_compiled
  platform = proj.get_platform
```

lib/vanagon/platform/dsl.rb
```diff
@@ -198,6 +198,20 @@ def build_hosts(*args)
...
+        @platform.cross_compiled = true unless flag.nil?
+      end
+
+      # Set a flag whether this is a cross-compiled Linux platform
+      #
+      # @param type [Boolean] true if this is a cross-compiled Linux platform
+      def cross_compiled_linux(flag)
+        @platform.cross_compiled_linux = true unless flag.nil?
+      end
```
mckern: This feels like something we could derive in `platform.rb` with an `&&` operation on `is_linux?` and a new helper method like `cross_compiled?` (which would itself probably just return the value of `@platform[:cross_compiled]`

McdonaldSeanp: @ScottGarman do you see scenarios where the platform is linux and you do _not_ want to set the cross_compiled_linux flag, but instead just cross_compiled?

lib/vanagon/platform/dsl.rb
McdonaldSeanp: These are platform files. But the point stands, and I assume that means you agree with me? we seem to be talking about the same thing

ScottGarman:
Closing this PR in favor of PR-368

PR #367, opened 2016-04-26
(maint) timeout not set in project dsl

McdonaldSeanp:
for any attribute of the project object we require a subsequent method in the
dsl for project. This was missing for the timeout attribute

### DIFF
--- a/lib/vanagon/project/dsl.rb
+++ b/lib/vanagon/project/dsl.rb
@@ -85,6 +85,13 @@ class Vanagon
         @project.homepage = page
       end
 
+      # Sets the timeout for the project retry logic
+      #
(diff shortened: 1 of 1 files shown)


underscorgan:
👍

lib/vanagon/project/dsl.rb
```diff
@@ -85,6 +85,13 @@ def homepage(page)
         @project.homepage = page
       end
 
+      # Sets the timeout for the project.
+      #
+      # @param page [String] url of homepage of the project
```
heathseals: This comment should probably be updated with the relevant info.

McdonaldSeanp: Oh oops =)

PR #368, opened 2016-04-26
(RE-7015) Add is_cross_compiled utility matcher methods

ScottGarman:
With all the cross-compiling we're now doing, these methods will help us
to refactor vanagon-based component configs to improve readability.

Will remove the don't merge label once I've thoroughly tested this.


ScottGarman:
It blends!

PR #370, opened 2016-05-04
(CPR-336) Retry was not printing out error messages

underscorgan:
Also, it seems like retry should raise the same error type as the
underlying command when available to avoid rogue rescues.


underscorgan:
@stahnma can you test this out?

Current theory is that something was rescuing the Vanagon::Error and that that was causing the ensure not to fire, but I'm not certain.

stahnma:
👍 this works for my use cases.

PR #372, opened 2016-05-05
(QENG-3556) Add CLI for retrieving the type of build host

joshcooper:
Adds a `build_host` CLI for retrieving the type of build host for a vanagon project and platform to enable the following workflow:

```
$ bundle exec build_host puppet-agent el-7-x86_64
centos-7-x86_64
$ ~/bin/vmpooler.py -g centos-7-x86_64
vwoafc8cxlc8u03
$ bundle exec build puppet-agent el-7-x86_64 vwoafc8cxlc8u03
Executing 'rpm -q curl > /dev/null || yum -y install curl && ...
```

### DIFF
--- a/lib/vanagon/engine/pooler.rb
+++ b/lib/vanagon/engine/pooler.rb
@@ -19,6 +19,16 @@ class Vanagon
         'pooler'
       end
 
+      # Return the vmpooler template name to build on
+      def build_host_name
+        if @build_host_template_name.nil?
+          validate_platform
+          @build_host_template_name = @platform.vmpooler_template
+        end
+
+        @build_host_template_name
+      end
+
       # This method loads the pooler token from one of two locations
       # @return [String, nil] token for use with the vmpooler
       def load_token
@@ -40,12 +50,12 @@ class Vanagon
         response = Vanagon::Utilities.http_request(
           "#{@pooler}/vm",
           'POST',
-          '{"' + @platform.vmpooler_template + '":"1"}',
+          '{"' + build_host_name + '":"1"}',
           { 'X-AUTH-TOKEN' => @token }
         )
         if response["ok"]
-          @target = response[@platform.vmpooler_template]['hostname'] + '.' + response['domain']
-          Vanagon::Driver.logger.info "Reserving #{@target} (#{@platform.vmpooler_template}) [#{@token ? 'token used' : 'no token used'}]"
+          @target = response[build_host_name]['hostname'] + '.' + response['domain']
+          Vanagon::Driver.logger.info "Reserving #{@target} (#{build_host_name}) [#{@token ? 'token used' : 'no token used'}]"
 
           tags = {
             'tags' => {
@@ -56,13 +66,13 @@ class Vanagon
           }
 
           Vanagon::Utilities.http_request(
-            "#{@pooler}/vm/#{response[@platform.vmpooler_template]['hostname']}",
+            "#{@pooler}/vm/#{response[build_host_name]['hostname']}",
             'PUT',
             tags.to_json,
             { 'X-AUTH-TOKEN' => @token }
           )
         else
-          raise Vanagon::Error, "Something went wrong getting a target vm to build on, maybe the pool for #{@platform.vmpooler_template} is empty?"
+          raise Vanagon::Error, "Something went wrong getting a target vm to build on, maybe the pool for #{build_host_name} is empty?"
         end
       end
 
--- a/lib/vanagon/engine/pooler.rb
+++ b/lib/vanagon/engine/pooler.rb
@@ -7,11 +7,16 @@ class Vanagon
 
       # The vmpooler_template is required to use the pooler engine
       def initialize(platform, target = nil)
+        super
+
         @pooler = "http://vmpooler.delivery.puppetlabs.net"
         @token = load_token
-        super
         @required_attributes << "vmpooler_template"
-        @name = 'pooler'
+      end
+
+      # Get the engine name
+      def name
+        'pooler'
       end
 
       # This method loads the pooler token from one of two locations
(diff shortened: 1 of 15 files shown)


joshcooper:
I need to add tests for this (hence the do not merge), but wanted to get feedback before doing that. /cc @rick @stahnma @waynr

stahnma:
I was wondering about the UX and thinking maybe it should return what type of answer it's giving. Like in the example you provide in the commit message, it's returning a poolname, not a hostname. Whereas if you did the same thing with aix-7.1-ppc it would return a hostname. 

Perhaps something like

pool: centos-7-x86_64

hostname: pe-aix-71-builder

container: a-stupid-container-name

Just an idea though.

waynr:
Does this require vanagon to be operating within the context of the specified project (`puppet-agent` in your example)?

I ask because in a CI model where we hold off on consuming job execution resources until we have the necessary dynamic infrastructure compute resources it's not clear how that will be possible if each project still has its own alias for dynamic infrastructure compute resource templates and can only acquire its dynamic infrastructure compute resources in a vmpooler-specific way.

joshcooper:
> I was wondering about the UX and thinking maybe it should return what type of answer it's giving. Like in the example you provide in the commit message, it's returning a poolname, not a hostname. Whereas if you did the same thing with aix-7.1-ppc it would return a hostname.

Currently puppet-agent CI packaging jobs know the [build hosts for aix](https://github.com/puppetlabs/ci-job-configs/blob/33b17e6f819f2143b8f6d3b277771e624919f751/resources/scripts/vanagon-exec-ship.sh#L59-L67):

```bash
elif [[ "$BUILD_TARGET" = "aix-5.3-ppc" ]]; then
    bundle exec build puppet-agent $BUILD_TARGET pe-aix-53-builder
elif [[ "$BUILD_TARGET" = "aix-6.1-ppc" ]]; then
    bundle exec build puppet-agent $BUILD_TARGET pe-aix-61-builder
elif [[ "$BUILD_TARGET" = "aix-7.1-ppc" ]]; then
    bundle exec build puppet-agent $BUILD_TARGET pe-aix-71-builder
else
    bundle exec build puppet-agent $BUILD_TARGET
fi
```

And that mapping from build target to build host name is repeated in the puppet-agent `configs/platform/aix-7.1-ppc.rb` :

```ruby
platform "aix-7.1-ppc" do |plat|
 ...
 plat.build_host ["pe-aix-71-builder.delivery.puppetlabs.net"]
end
```

Given the current state of CI jobs, we would never need to call `bundle exec build_host aix-7.1-ppc` to resolve the `build_host`.

It might make sense to remove that logic from the CI job, at which point the response would need to include more information. But I'm not 100% clear about how we expect hardware locking to work in the mesos world. I imagine the resource name, e.g. `pe-aix-71-builder`, `centos-7-x86_64`, and the engine name, e.g. `hardware`, `pooler` would both be part of the resource offer/accept logic. That said, I think we should consider hardware locking as a special case now and come back to that later.

stahnma:
1. We should remove that from the CI job, as that means it's not locking currently, which is bad. (I actually thought that was fixed /cc @geoffnichols )
2. Can we just output the engine with the value? 
3. Totally agree hardware locking may change quite a bit in a future world. No idea how though.

joshcooper:
> Does this require vanagon to be operating within the context of the specified project (puppet-agent in your example)?

Yes, but the `bundle exec build_host puppet-agent <target>` call can happen in an earlier job before the real packaging job is started. For example, an upstream job could clone `puppet-agent`, bundle install, execute `bundle exec build_host puppet-agent el-7-x86_64`, and pass the template name & puppet-agent tarball artifact through to a downstream job. That way all of the vanagon packaging dependencies are known before we execute the packaging job.

>  it's not clear how that will be possible if each project still has its own alias for dynamic infrastructure compute resource templates

It's important to note that it's more than aliases. Puppet-agent supports a number of different compilation modes (native and cross-compilation) and doesn't always rely on vmpooler template names. That logic lives in puppet-agent as it should, since the project needs to define how to build itself on a given platform. The actual build dependency may be a vmpooler template or static aix/netdev build host:

> and can only acquire its dynamic infrastructure compute resources in a vmpooler-specific way [EDIT: ...during job execution].

The `build_host` command doesn't acquire any resources. It only resolves what resources are necessary. So that an upstream job and get the appropriate resource from mesos. Once the resource becomes available, a downstream job can execute `bundle exec build puppet-agent el-7-x86_64 <fqdn>` to start the build.

joshcooper:
> We should remove that from the CI job, as that means it's not locking currently, which is bad. (I actually thought that was fixed /cc @geoffnichols )

Vanagon knows that if a project [defines a `build_host` then it should use the `hardware` engine and lock the node](https://github.com/puppetlabs/vanagon/blob/master/lib/vanagon/driver.rb#L36-L38).  So I believe we are locking during aix builds correctly

> Can we just output the engine with the value?

Yeah we could, though it feels like the interface is not well defined. I need to think about this more.

mckern:
> Vanagon knows that if a project defines a build_host then it should use the hardware engine and lock the node. So I believe we are locking during aix builds correctly

That's kind of true -- it's more like "if  a specific node is defined, use the least invasive engine that supports tearing down and locking" than a conscious choice to use Hardware as the type of engine in question.

joshcooper:
> 22:15:54 Lock acquired on pe-aix-61-builder.delivery.puppetlabs.net for jenkins.

Yep, we're locking

lib/vanagon/engine/pooler.rb
```diff
@@ -35,12 +48,12 @@ def select_target
         response = Vanagon::Utilities.http_request(
           "#{@pooler}/vm",
           'POST',
-          '{"' + @platform.vmpooler_template + '":"1"}',
+          '{"' + @build_host_type + '":"1"}',
```
rick: Random question -- why not use `build_host_type` here instead of `@build_host_type`?

joshcooper: This method shouldn't be called before the instance variable has been set, but just to be sure, I modified it to call the getter method instead.

rick:
I'm :+1: on this.

geoffnichols:
This looks good to me. I'm 👍 .

PR #373, opened 2016-05-11
(maint) Add unit test for ec2 engine

joshcooper:
Commit bab55a46 had a trailing comma after `:instance_type`, which seems
to cause the following `initialize` method to be ignored. So trying to
create an EC2 engine and passing in a platform didn't work:

    irb(main):033:0> Vanagon::Engine::Ec2.new(plat._platform)
    ArgumentError: wrong number of arguments (1 for 0)

Commit b8f537a74 changed how engines return their name, so it's no
longer an instance variable. That combined with bab55a46, caused the
Ec2 engine to report its name as 'base':

    irb(main):035:0> Vanagon::Engine::Ec2.new.name
    => "base"

This commit removes the trailing comma, fixes the engine name, and adds
a unit test to make sure we can create an EC2 engine and get the correct
name.

### DIFF
--- a/lib/vanagon/engine/ec2.rb
+++ b/lib/vanagon/engine/ec2.rb
@@ -7,12 +7,11 @@ class Vanagon
   class Engine
     class Ec2 < Base
       attr_accessor :ami, :key_name, :userdata, :key, :key_name, :shutdown_behavior
-      attr_accessor :subnet_id, :instance_type,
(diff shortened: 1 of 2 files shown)


joshcooper:
ping @cwood @McdonaldSeanp

McdonaldSeanp:
👍

cwood:
Looks fine to me.

PR #375, opened 2016-05-16
(maint) Add vanagon user docs

McdonaldSeanp:
These will eventually turn in to user docs for all of vanagon, but for now they
contain the user docs for creating a windows project with vanagon.

PR #376, opened 2016-05-16
(maint) Add the vanagon design doc to vanagon

McdonaldSeanp:
This is an incomplete doc only containing the docs for the windows platform,
more docs are to come.

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -1,0 +1,110 @@
+# Vanagon
+- Note that this is an incomplete doc, documentation on all aspects of vanagon to come
+
+## Vanagon for the Windows platform
+
+### Overview
+Vanagon was adapted for the windows platform using [WiX](http://wixtoolset.org/). At a very high level each component of the project gets built on a build system, then gets moved to a staging directory where we use the WiX toolset to build an MSI. This workflow is consistent with the rest of vanagon, however the WiX toolset creates some caveats, specifically:
+
+* Vanagon does not contain functionality to handle registry searching
+* Vanagon does not contain functionality to handle the MSI UI
+* If the project requires MSI custom actions (such as those to search the registry, set properties used in the MSI etc.) Vanagon has no functionality to handle those
+* For directories and files that require special attention (Such as a requirement for an app data directory to persist after uninstall) vanagon cannot handle them
+* Vanagon cannot handle the service files required by WiX for an MSI
+* Vanagon cannot handle the localization engine that MSIs use
+* For localisation purposes, key Windows directories (e.g. Program files) must be specified using the install time dependancy localisation structure for windows
+
+We deal with these special cases by allowing the user to provide their own WiX files inside the vanagon project. This way the user can use the WiX toolset to solve problems like registry searching and service files.
+
+It’s possible many of these edge cases can be remedied in future releases of vanagon, such as the special casing of directories and files. But the benefit of “consistency” may not be enough to drive change for other cases, specifically because for things like custom actions the changes to vanagon would be so drastic and edge cased for windows, vanagon would start to look inconsistent with itself.
+
+### Vanagon classes, functions, and WiX
+We made no major changes to the overall class structure in vanagon, most of the work has been done in the class functions to incorporate WiX. Following along the Vanagon::Driver class function [run](https://github.com/puppetlabs/vanagon/blob/master/lib/vanagon/driver.rb#L89L116) :
+
+1. [Install_build_dependancies](https://github.com/puppetlabs/vanagon/blob/master/lib/vanagon/driver.rb#L98) remains unchanged. The underlying changes required to get dependencies for windows lies in the actual project, not vanagon itself.
+2. [Fetch_sources](https://github.com/puppetlabs/vanagon/blob/master/lib/vanagon/driver.rb#L99) also remains unchanged, though take note that this is NOT where we fetch any of the WiX source files. However this is where we will grab any .bat files to be put on the user’s PATH on install.
+3. [Make_makefile](https://github.com/puppetlabs/vanagon/blob/master/lib/vanagon/driver.rb#L100) is where some of the WiX magic begins to happen. Specifically the WiX compiler invocation exists under the generate_package function of Vanagon::Platform::Windows
+4. [Make_bill_of_materials](https://github.com/puppetlabs/vanagon/blob/master/lib/vanagon/driver.rb#L101) remains unchanged
+6. [Generate_pacakging_artifacts](https://github.com/puppetlabs/vanagon/blob/master/lib/vanagon/driver.rb#L102) is where we are actually adding / generating WiX files for the compiler to use. If you follow the objects down eventually you’ll find the [generate_msi_packaging_artifacts](https://github.com/puppetlabs/vanagon/blob/master/lib/vanagon/platform/windows.rb#L34L48) function inside Vanagon::Platform::Windows, where you’ll notice we do a mix of taking default files in vanagon and files from the project source and putting them in the workdir.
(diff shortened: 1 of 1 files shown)


DESIGN.md
```diff
@@ -0,0 +1,108 @@
+# Vanagon
+- Note that this is an incomplete doc, documentation on all aspects of vanagon to come
+
+## Vanagon for the Windows platform
+
+### Overview
+Vanagon was adapted for the windows platform using WiX. At a very high level each component of the project gets built on a build system, then gets moved to a staging directory where we use the WiX toolset to build an MSI. This workflow is consistent with the rest of vanagon, however the WiX toolset is not so simple that there are no caveats, specifically:
```
jcoconnor: Put in URL links for Wix like you have in #375 
Also shouldn't that be "there are caveats"

DESIGN.md
```diff
@@ -0,0 +1,108 @@
...
+
+* Vanagon does not contain functionality to handle registry searching
+* Vanagon does not contain functionality to handle the MSI UI
+* If the project requires MSI custom actions (such as those to search the registry, set properties used in the MSI etc.) Vanagon has no functionality to handle those
+* For directories and files that require special attention (Such as a requirement for an app data directory to persist after uninstall) vanagon cannot handle them
+* Vanagon cannot handle the service files required by WiX for an MSI
+* Vanagon cannot handle the localization engine that MSIs use
```
jcoconnor: You also need to put in the very specific requirement that key Windows directories (e.g. Program files) must be specified using the install time dependancy localisation structure for windows. This is a key difference from the Unix style where we can specify in advance exactly where the files go, so we can actually build a replica of where the files go at build time, whereas with Windows we need to rely instead on install time relocatable root points.

DESIGN.md
```diff
@@ -0,0 +1,108 @@
...
+
+We deal with these special cases by allowing the user to provide their own WiX files inside the vanagon project. This way the user can use the WiX toolset to solve problems like registry searching and service files.
+
+It’s possible many of these edge cases can be remedied in future releases of vanagon, such as the special casing of directories and files. But the benefit of “consistency” may not be enough to drive change for other cases, specifically because for things like custom actions the changes to vanagon would be so drastic and edge cased for windows, vanagon would start to look inconsistent with itself.
+
+### Vanagon classes, functions, and WiX
+We made no major changes to the overall class structure in vanagon, most of the work has been done in the class functions to incorporate WiX. Following along the Vanagon::Driver class function “run” :
+
+1. Install_build_dependancies remains unchanged. The underlying changes required to get dependencies for windows lies in the actual project, not vanagon itself.
```
jcoconnor: Should you put in links to the code examples you are referring to here and generally in this list?

DESIGN.md
```diff
@@ -0,0 +1,108 @@
...
+2. Fetch_sources also remains unchanged, though take note that this is NOT where we fetch any of the WiX source files. However this is where we will grab any .bat files to be put on the user’s PATH on install.
+3. Make_makefile is where some of the WiX magic begins to happen. Specifically the WiX compiler invocation exists under the generate_package function of Vanagon::Platform::Windows
+4. Make_bill_of_materials remains unchanged
+6. Generate_pacakging_artifacts is where we are actually adding / generating WiX files for the compiler to use. If you follow the objects down eventually you’ll find the generate_msi_packaging_artifacts function inside Vanagon::Platform::Windows, where you’ll notice we do a mix of taking default files in vanagon and files from the project source and putting them in the workdir.
+7. - 10. The remainder of the driver functionality remains unchanged
```
jcoconnor: This is where a diagram would really help - way too much tribal knowledge in getting to the stage where you and I are now ;)

DESIGN.md
```diff
@@ -0,0 +1,108 @@
...
+
+Other than SourceDir the other related naming convention is ProgramFilesFolder / ProgramFiles64Folder. These two names are [MSI defaults](https://msdn.microsoft.com/en-us/library/windows/desktop/aa370905(v=vs.85).aspx) that allow us to utilize the localization engine for MSIs. On install these folders will resolve to the actual program files standard for windows, i.e. Program Files or Program Files (x86).
+
+### Heat.exe
+The heat process generates appropriate WiX elements corresponding to each file and directory under the stagedir. Vanagon uses heat’s output alongside the other static WiX files to include all files under the stagedir in the final MSI.
```
jcoconnor: This process is referred to as "harvesting" the files

DESIGN.md
```diff
@@ -0,0 +1,108 @@
...
+Heat uses the SourceDir naming convention via input parameters to the command line. The process by which vanagon uses heat differs from the normal usage. This is because we cannot run heat anything in the CommonAppData folder. It’s likely the files/directories that go there need to be permanent and heat cannot handle permanent files/directories. In order to work around this we do not run heat.exe from SourceDir, as we only want to collect those files NOT under CommonAppData. Instead we run heat.exe from the path:
```
jcoconnor: Maybe re-word to say that some of the directories and files are added separately to the heat/harvest operation, so we only do the harvest under the program directory.
This also allows us to do additional special handling for the commonAppData (e.g. make sure configuration files are not over-written).
The harvest operation is like a bulk collection of multiple files which are handled in the default way (if that makes sense) - all of these files are treated as simple single file components with no additional handling.
Also mention that service files are filtered out from the harvest operation as they need additional handling as service component files.

DESIGN.md
```diff
@@ -0,0 +1,108 @@
...
+
+In cases where there is more than one bindir_id corresponding to the same place there is no problem, there will just be two nameless directory elements under that directory. This is allowed in WiX because they have no names.
+
+### INSTALLDIR
+Inside the directorylist file the last root directory defined has the unique id INSTALLDIR. Vanagon uses this name on purpose. Used correctly, INSTALLDIR provides the user the means to change the location of the final install of the MSI. Using WiX properties and the INSTALLDIR variable the user can override all root directories underneath INSTALLDIR with what the end user wants.
+
+Note that this is in no way the correct way for the vanagon user to override directorylist root install defaults. If the user wishes the default installation location to be something other than DRIVE:/ProgramFilesFolder/CompanyName/ProductName they should override the entire directorylist file with a file of the same name in their own project. Vanagon will default to using the one in their project.
+
+INSTALLDIR itself is not a WiX property or requirement, the name was a convention that vanagon uses.
```
jcoconnor: Put in a link/reference to this.
Its not a standard MSI property, but has become a kind of de-facto standard after Installshield brought it into common usage.

DESIGN.md
```diff
@@ -0,0 +1,108 @@
...
+
+The invocation of candle is very straightforward, with one caveat: as discussed earlier, we need to pass the preprocessor variable “AppSourcePath” for heat to use.
+
+More information on the Candle compiler:
+
+* [docs](http://wixtoolset.org/documentation/manual/v3/)
+* [tools](http://wixtoolset.org/documentation/manual/v3/overview/alltools.html)
+
+### Light.exe
+Light.exe is akin to the linking step in compilers. In the invocation of light we don’t use anything out of the ordinary, we simply pass it what the location the files exist in, and where the output file should go. The output of Light.exe is an actual MSI, so after this step vanagon has completed compilation.
```
jcoconnor: Mention that localisation is also fed in at this stage from localised string tables.

jcoconnor:
@McdonaldSeanp Overall this looks very good - have added comments on individual sections.

DESIGN.md
McdonaldSeanp: As the more complete vanagon docs come out I will add some diagrams =)

DESIGN.md
McdonaldSeanp: done

PR #378, opened 2016-05-20
(MAINT) Confine aws-sdk gem to bundler group

mckern:
It's currently an optinal engine, and it should specify an optional dependency. This also makes the spec test for the EC2 engine option depending on whether or not the aws-sdk is available.

When this engine is declared a prime-time engine, the dependency on AWS-SDK should be moved directly into the Gemspec and this confinement should be removed.

### DIFF
--- a/spec/lib/vanagon/engine/ec2_spec.rb
+++ b/spec/lib/vanagon/engine/ec2_spec.rb
@@ -1,25 +1,33 @@
-require 'vanagon/engine/ec2'
-require 'vanagon/platform'
+begin
+  require 'aws-sdk'
+rescue LoadError
+  $stderr.puts "Unable to load AWS SDK; skipping optional EC2 engine spec tests"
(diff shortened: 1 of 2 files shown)


McdonaldSeanp:
👍

spec/lib/vanagon/engine/ec2_spec.rb
```diff
@@ -1,25 +1,33 @@
-require 'vanagon/engine/ec2'
-require 'vanagon/platform'
+begin
+  require 'aws-sdk'
+rescue LoadError
```
stahnma: Will these run for sure on the Travis?

mckern: Bundler groups are inclusionary, not exclusionary. With no flags passed, Bundler will install all  gems and groups:

```

PR #380, opened 2016-05-25
(maint) Disable brp-strip-static-archive from RPM builds

ScottGarman:
brp-strip-static-archive causes cross-compiled static libs to end up
with "no machine" as their architecture, breaking builds.

With this fix, I'm now able to build all of the pl-build-tools targets for el-6-s390x. Going to run this through jenkins to ensure there are no regressions.

### DIFF
--- a/resources/rpm/project.spec.erb
+++ b/resources/rpm/project.spec.erb
@@ -2,6 +2,10 @@
 
 # Turn off the brp-python-bytecompile script
 %global __os_install_post %(echo '%{__os_install_post}' | sed -e 's!/usr/lib[^[:space:]]*/brp-python-bytecompile[[:space:]].*$!!g')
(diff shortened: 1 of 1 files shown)


shrug:
👍

shrug:
we may want to come up with arch-specific strip (in pl-binutils) if we ever need to produce debug packages? Out of scope for this work though

ScottGarman:
I checked with @stahnma on how important it would be to limit this to cross-compiled targets; he seemed to think it was okay to just disable that brp for all builds.

ScottGarman:
An extensive suite of RPM-based platforms has passing builds on jenkins: http://jenkins-release.delivery.puppetlabs.net/job/vanagon_generic_job/430/

Should be good to go!

PR #381, opened 2016-06-07
(maint) remove some wix defaults

McdonaldSeanp:
After realizing several "default" wxs files were not actually generic files but
project dependent, we need to move them out of vanagon/resources.

Instead of keeping these files as generic "defaults" they have all already
moved under resources in the example directory in the case that the user wants
to follow our suggested outlines in the docs. Those files should be pretty plug
and play with any project (meaning you should be able to specify some
variables in the project configs and then just copy these files and things
should work)


geoffnichols:
👍

PR #382, opened 2016-06-09
(MAINT) Finesse retries a little bit

mckern:
Retries and timeouts can now be defined with the environment variables `RETRY_COUNT` and `TIMEOUT`. Useful for turning them up or down without having to commit a new value to a project every time you change it.

Also, the default retry count has been changed from 3 to 1. For _**reasons**_.

### DIFF
--- a/lib/vanagon/driver.rb
+++ b/lib/vanagon/driver.rb
@@ -138,8 +138,8 @@ class Vanagon
     # values from the project, if available, otherwise use some
     # sane defaults.
     def retry_task(&block)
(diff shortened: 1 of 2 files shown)


shrug:
👍 

I was also going to remove the retry from the actual make call, it doesn't seem to buy us anything. We can do that in another pr.

stahnma:
😂 😂 😄 😀 💃 💯

stahnma:
@mckern could you update the README to document those variables?

mckern:
@stahnma I **can** but **will** I?

PR #383, opened 2016-06-09
(maint) Increase default timeout

stahnma:
When building toolchain components such as GCC, it takes more than 1
hours. This ups the default timeout to 2 hours since our jenkins configs
don't pass timeout as a parameter.


mckern:
👍

mckern:
Still 👍

PR #384, opened 2016-06-20
(PA-323) Update osx packaging to include an uninstaller

McdonaldSeanp:
For packages requiring extra steps in uninstallation, such as unloading
services, add an uninstaller to vanagon packages for OSX.

### DIFF
--- a/lib/vanagon/platform/osx.rb
+++ b/lib/vanagon/platform/osx.rb
@@ -25,8 +25,11 @@ class Vanagon
          # Setup build directories
         ["bash -c 'mkdir -p $(tempdir)/osx/build/{dmg,pkg,scripts,resources,root,payload,plugins}'",
          "mkdir -p $(tempdir)/osx/build/root/#{project.name}-#{project.version}",
+         "mkdir -p $(tempdir)/osx/build/pkg",
          # Grab distribution xml, scripts and other external resources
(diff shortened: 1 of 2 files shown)


heathseals:
have we considered shipping this as an uninstaller script that ships/installs with the installer pkg?  seems like a hassle to keep the dmg around to run an uninstaller pkg.

McdonaldSeanp:
Yes, @geoffnichols and I discussed this, and at this point there is no really great way to package an uninstaller for OSX, so whatever we do is really not canonical with OSX. What we had decided on was to stick with legacy behavior and keep things consistent on our end. At this point this .pkg will be consistent with what docs say and users would have read. 

As a side note I would like to point out that you actually don't have to keep the whole dmg around, you can just cp the unstaller.pkg out of it and only keep that (if you so wish)

McdonaldSeanp:
So what I have changed this to is just a script, but not part of the installer payload. At this point it looks like others (such as vagrant) will do exactly what's done here, which is ship the uninstaller script next to the installer.pkg

McdonaldSeanp:
I have manually tested this PR with puppet-agent and everything works great.

stahnma:
👍

mckern:
I don't know if there's any work or logic to handle positioning this thing in the resulting DMG but I'm 👍  on the work itself

PR #386, opened 2016-06-22
(MAINT) Namespace VANAGON_TIMEOUT and VANAGON_RETRY_COUNT env. vars

mckern:
These are pretty common names, so we should try to minimize the
possibility of collision.

Also, added documentation for VANAGON_TIMEOUT environment variable,
which was previously undocumented.


stahnma:
👍

PR #387, opened 2016-06-27
(PA-390) On AIX, stop services also on upgrade

geoffnichols:
Prior to this, services on AIX are not restarted on upgrade, so
the old puppet and mcollective services may be left running.

This updates the RPM `%preun` to stop AIX services also on upgrade.

In addition, prior to this, vanagon creates empty "install" and "upgrade"
files in the RPM localstatedir on all platforms. These files are created
for use by `%posttrans` scriptlets.

AIX and EL4 use an older version of RPM that does not support use of
`%posttrans`, so the state files are not needed on those platforms.

This also updates vanagon's rpm spec template so the empty "install"
and "upgrade" files are not created on AIX and EL4.

### DIFF
--- a/resources/rpm/project.spec.erb
+++ b/resources/rpm/project.spec.erb
@@ -241,9 +241,7 @@ fi
       chkconfig --del <%= service.name %> || :
     fi
   <%- elsif @platform.servicetype == "aix" -%>
-    if [ "$1" -eq 0 ]; then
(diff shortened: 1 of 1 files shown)


geoffnichols:
Depends on https://github.com/puppetlabs/puppet-agent/pull/736.

geoffnichols:
Closing this for now.

PR #388, opened 2016-06-27
(CPR-369) Don't create "install", "upgrade" files on AIX, EL4.

geoffnichols:
Prior to this commit, vanagon creates empty "install" and "upgrade"
files in the RPM localstatedir on all platforms. These files are created
for use by `%posttrans` scriptlets.

AIX and EL4 use an older version of RPM that does not support use of
`%posttrans`, so the state files are not needed on those platforms.

This commit updates vanagon's rpm spec template so the empty "install"
and "upgrade" files are not created on AIX and EL4.

### DIFF
--- a/resources/rpm/project.spec.erb
+++ b/resources/rpm/project.spec.erb
@@ -123,6 +123,7 @@ done
 
 
 %pre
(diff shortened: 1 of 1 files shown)


McdonaldSeanp:
👍

PR #390, opened 2016-06-28
(PA-390) On AIX, stop services also on upgrade.

geoffnichols:
Prior to this commit, on upgrade, services on AIX are not restarted, so
the old puppet and mcollective services may be left running.

This commit updates the `%preun` to stop AIX services also on upgrade.


shrug:
👍

PR #392, opened 2016-07-06
(maint) Update the directorylist creation for windows to use uniq id

McdonaldSeanp:
It's posible to cause a failure in WiX by registering two services on paths
that are different but have the same named dir somewhere, for example,
registering two services under:

    /dir/bin

    /other_dir/bin

will cause a failure because when we create the directorylist.wxs file the
"bin" directories will both have the same "Id" field because we just used
name there. This commit appends values to the Id field that will ensure
uniqueness.

This commit upates both the windows platform object and rspec

### DIFF
--- a/lib/vanagon/platform/windows.rb
+++ b/lib/vanagon/platform/windows.rb
@@ -335,12 +335,14 @@ class Vanagon
         # iterate over all paths specified and break each one
         # in to its specific directories. This will generate_wix_dirs
         # an n-ary tree structure matching the specs from the input
-        items.each do |item|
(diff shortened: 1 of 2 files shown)


underscorgan:
assuming  travis is 🍏 , 👍

melissa:
Wil this require that we specify an index when declaring the directory? Where does the index get set?

McdonaldSeanp:
No, this is all under the vanagon hood, the index is set here in the code that's changing.

melissa:
fancy

melissa:
👍 will merge when travis is 💚

PR #394, opened 2016-07-12
(RE-6200) URI parsing, Source handling, Git cloning

mckern:
I started refactoring how `Vanagon::Component::Source` works... and wound up making **sweeping** revisions to source handling in general. There's probably more to come, but I wanted to get people looking at this before it gets even larger and more difficult to grok. Spec tests pass, but coverage is incomplete so I'd like to start testing this out before anyone even considers merging it.


lib/vanagon/component/dsl.rb
```diff
@@ -317,11 +317,11 @@ def build_dir(path)
 
       # This will add a source to the project and put it in the workdir alongside the other sources
       #
-      # @param url [String] url of the source
+      # @param uri [String] uri of the source
```
stahnma: Update docs here

lib/vanagon/component/source.rb
```diff
@@ -1,78 +1,114 @@
...
+      @rewrite_rules = {}
 
-      def self.register_rewrite_rule(protocol, rule)
-        if rule.is_a?(String) or rule.is_a?(Proc)
-          if SUPPORTED_PROTOCOLS.include?(protocol)
-            @@rewrite_rule[protocol] = rule
+      class << self
+        attr_reader :rewrite_rules
+
+        # Deprecate this
```
stahnma: Should we add a deprecation warning to it now? Like, please don't build upon this?

lib/vanagon/component/source.rb
```diff
@@ -1,78 +1,114 @@
...
-                        when /^http/
-                          Vanagon::Component::Source::Http.new(self.rewrite(url, 'http'), options[:sum], workdir)
-                        when /^file/
-                          Vanagon::Component::Source::Local.new(self.rewrite(url, 'file'), workdir)
-                        when /^git/
-                          Vanagon::Component::Source::Git.new(self.rewrite(url, 'git'), options[:ref], workdir)
-                        else
-                          fail "Don't know how to handle source of type '#{uri_scheme}' from url: '#{url}'"
-                        end
+        # Needs a better name
```
stahnma: Name and yard docs.

lib/vanagon/component/source.rb
```diff
@@ -1,78 +1,114 @@
...
+          # Then we try local
+          if Vanagon::Component::Source::Local.valid_file?(parse_and_rewrite(uri))
+            return Vanagon::Component::Source::Local.new parse_and_rewrite(uri),
+              workdir: options[:workdir]
+          end
 
-        return local_source
+          # Failing all of that, we give up
+          raise Vanagon::Error,
+            "Don't know how to handle unknown type: '#{uri}'"
```
stahnma: Either have a subject or get rid of the verb :)

lib/vanagon/component/source.rb
```diff
@@ -1,78 +1,114 @@
...
+        end
+
+        # Needs a better name
+        def absolute_certainty(uri, type)
+          # try to parse the URI as whatever type was passed
```
stahnma: Is this needed?

lib/vanagon/component/source/git.rb
```diff
@@ -1,37 +1,88 @@
 require 'vanagon/utilities'
+require 'vanagon/errors'
+# This stupid library requires a capital 'E' in its name
```
stahnma: You say it needs a capital E but then I don't see english with a capital E

lib/vanagon/component/source/git.rb
```diff
@@ -1,37 +1,88 @@
...
+            output
+          end
+          private :git
+
+          # Attempt to connect to whatever URL is provided and
+          # return True or False depending on whether or not
+          # `git` thinks it's a valid Git repo.
+          #
+          # @return [Boolean] whether #url is a valid Git repo or not
+          def valid_remote?(url)
```
stahnma: Should these type of generic git operations that aren't manipulating vanagon objects be in another library?

lib/vanagon/component/source/local.rb
```diff
@@ -12,11 +12,17 @@ class Local
...
+            File.exist?(target_file.to_s)
+          end
+        end
+
         # Constructor for the File source type
         #
         # @param url [String] url of the http source to fetch
         # @param workdir [String] working directory to download into
-        def initialize(url, workdir)
+        def initialize(url, workdir:, **options)
```
stahnma: update yard docs

lib/vanagon/errors.rb
```diff
@@ -26,3 +26,4 @@ def initialize(mesg)
     end
   end
 end
+
```
stahnma: remove trailing whitespace

lib/vanagon/utilities.rb
```diff
@@ -5,10 +5,14 @@
 require 'digest'
 require 'erb'
 require 'timeout'
+# This stupid library requires a capital 'E' in its name
+# but it provides a wealth of useful constants
+require 'English'
```
stahnma: So it has a big E here....

lib/vanagon/component.rb
```diff
@@ -140,10 +141,13 @@ def get_build_dir
     #
     # @param workdir [String] working directory to put the source into
     def get_sources(workdir)
-      @sources.each do |source|
-        cur_source = Vanagon::Component::Source.source(source.url, { :ref => source.ref, :sum => source.sum }, workdir)
-        cur_source.fetch
-        cur_source.verify
+      sources.each do |source|
```
McdonaldSeanp: Am I missing something that ruby does to resolve the source variable here? From what it looks like the sources var is only in scope to the get_sources function and it's not defined yet in this function.

mckern: https://github.com/mckern/vanagon/blob/97bd1c4a8b697d9f3f0fd1f236ab9fae5589d96f/lib/vanagon/component.rb#L14

`attr_accessor` is the magic you're looking for.

McdonaldSeanp: I thought even with attr_accessor ruby still requires you use @{var}

mckern: `attr_accessor` defines both the getter and setter methods, and directly translates those methods to the instance variable if no overriding method is defined. So calling `sources` here is actually calling the method, not the instance variable (`@sources`). You would only call the instance variable directly if you were actually doing something that would change the state of the value (beyond calling it or assigning it).

McdonaldSeanp: Alrighty

lib/vanagon/component/source/git.rb
mckern: Ostensibly, and @shrug and I evaluated using something like rugged (Ruby bindings to libgit2) for these operations. The benefit was negligible for our use case, since git cloning is basically analogous to an HTTP `GET` or a local file `cp` for how we're using Git.

We could probably research other options down the road.

lib/vanagon/component/source/git.rb
mckern: Thanks. I fixed it.

lib/vanagon/component/source.rb
mckern: ... `"Don't know how to unknown type: '#{uri}'"`?

lib/vanagon/component/source.rb
mckern: Probably.

lib/vanagon/component/source.rb
mckern: I'm going to add `:type` as a value for source definition, and this will be used there. I didn't realize I'd left the stub in when I committed this.

stahnma:
<pre>
time bundle exec build puppet-agent el-7-x86_64
Error loading project 'puppet-agent' using '/home/stahnma/puppet-agent/configs/projects/puppet-agent.rb':
undefined method `git_version' for Vanagon::Utilities:Module
/home/stahnma/vanagon/lib/vanagon/project/dsl.rb:147:in `version_from_git'
/home/stahnma/vanagon/lib/vanagon/project.rb:169:in `block in load_project'
/home/stahnma/vanagon/lib/vanagon/project/dsl.rb:26:in `call'
/home/stahnma/vanagon/lib/vanagon/project/dsl.rb:26:in `project'
/home/stahnma/vanagon/lib/vanagon/project.rb:28:in `load_project'
/home/stahnma/vanagon/lib/vanagon/project.rb:28:in `instance_eval'
/home/stahnma/vanagon/lib/vanagon/project.rb:28:in `load_project'
/home/stahnma/vanagon/lib/vanagon/driver.rb:26:in `initialize'
/home/stahnma/vanagon/bin/build:27:in `new'
/home/stahnma/vanagon/bin/build:27:in `block in <top (required)>'
/home/stahnma/vanagon/bin/build:25:in `each'
/home/stahnma/vanagon/bin/build:25:in `<top (required)>'
/home/stahnma/.gem/ruby/bin/build:23:in `load'
/home/stahnma/.gem/ruby/bin/build:23:in `<main>'
/home/stahnma/vanagon/lib/vanagon/project/dsl.rb:147:in `version_from_git': undefined method `git_version' for Vanagon::Utilities:Module (NoMethodError)
	from /home/stahnma/vanagon/lib/vanagon/project.rb:169:in `block in load_project'
	from /home/stahnma/vanagon/lib/vanagon/project/dsl.rb:26:in `call'
	from /home/stahnma/vanagon/lib/vanagon/project/dsl.rb:26:in `project'
	from /home/stahnma/vanagon/lib/vanagon/project.rb:28:in `load_project'
	from /home/stahnma/vanagon/lib/vanagon/project.rb:28:in `instance_eval'
	from /home/stahnma/vanagon/lib/vanagon/project.rb:28:in `load_project'
	from /home/stahnma/vanagon/lib/vanagon/driver.rb:26:in `initialize'
	from /home/stahnma/vanagon/bin/build:27:in `new'
	from /home/stahnma/vanagon/bin/build:27:in `block in <top (required)>'
	from /home/stahnma/vanagon/bin/build:25:in `each'
	from /home/stahnma/vanagon/bin/build:25:in `<top (required)>'
	from /home/stahnma/.gem/ruby/bin/build:23:in `load'
	from /home/stahnma/.gem/ruby/bin/build:23:in `<main>'

real	0m0.581s
user	0m0.528s
sys	0m0.050s
</pre>

I also had to update the vanagon.gemspec to include fuistigit

lib/vanagon/component/source.rb
stahnma: Jerk.

mckern: > Jerk.

Truthier words have never been written.

lib/vanagon/component/source.rb
mckern: I'm going to add deprecation notices to all of the rewrite methods in the next big PR, that will add the concept of mirror URLs. That will give me someplace else to point users for their needs.

lib/vanagon/component/dsl.rb
```diff
@@ -310,11 +310,11 @@ def build_dir(path)
...
       # @param ref [String] Used for git sources, must be a git ref of some sort
       # @param sum [String] sum used to validate http and file sources
-      def add_source(url, ref: nil, sum: nil)
-        @component.sources << OpenStruct.new(:url => url, :ref => ref, :sum => sum)
+      def add_source(uri, **options)
+        @component.sources << OpenStruct.new(options)
```
mckern: I hate that we're using OpenStruct here, because it means that there's really no documentation about what kind of object (or structure) we SHOULD be pushing into `@component.sources` here. I had to fish around for `#add_source` to figure out what is expected of each `source` we push into `@component.sources`.

mckern:
I have produced a probably-working build of `puppet-agent` (~1.5.2ish, with Ruby 2.3.0) from this codebase. I'm addressed a couple of the documentation fixups Mike suggested, but I think this is :raised_hands:.

mckern:
I don't know why this failed in Travis because:

```
mckern@flexo vanagon (git:RE-6200/you-are-i) $ bundle exec rspec spec --color --require spec_helper
.......................................................................................................................................................................................vcloud_name is a deprecated platform DSL method, and will be removed in a future vanagon release. Please use vmpooler_template instead.
.....vcloud_name is a deprecated platform DSL method, and will be removed in a future vanagon release. Please use vmpooler_template instead.
......................Please use the add_build_repository DSL method instead. apt_repo will be removed in a future vanagon release.
.Please use the add_build_repository DSL method instead. apt_repo will be removed in a future vanagon release.
.Please use the add_build_repository DSL method instead. apt_repo will be removed in a future vanagon release.
.Please use the add_build_repository DSL method instead. yum_repo will be removed in a future vanagon release.
.Please use the add_build_repository DSL method instead. yum_repo will be removed in a future vanagon release.
.Please use the add_build_repository DSL method instead. yum_repo will be removed in a future vanagon release.
.Please use the add_build_repository DSL method instead. zypper_repo will be removed in a future vanagon release.
.Please use the add_build_repository DSL method instead. zypper_repo will be removed in a future vanagon release.
....vcloud_name is a deprecated platform DSL method, and will be removed in a future vanagon release. Please use vmpooler_template instead.
........................................................................An error was encountered evaluating block. Retrying..
An error was encountered evaluating block. Retrying..
An error was encountered evaluating block. Retrying..
An error was encountered evaluating block. Retrying..
An error was encountered evaluating block. Retrying..
An error was encountered evaluating block. Retrying..
An error was encountered evaluating block. Retrying..
.An error was encountered evaluating block. Retrying..
An error was encountered evaluating block. Retrying..
An error was encountered evaluating block. Retrying..
An error was encountered evaluating block. Retrying..
An error was encountered evaluating block. Retrying..
An error was encountered evaluating block. Retrying..
..

Finished in 13.41 seconds (files took 0.2615 seconds to load)
296 examples, 0 failures

mckern@flexo vanagon (git:RE-6200/you-are-i) $
```

lib/vanagon/component/source/git.rb
mckern: These generic primitives have been moved into a 3rd party library, with wrappers around that library for our needs.

lib/vanagon/utilities.rb
mckern: Fixed.

mckern:
### Still needs

- [ ] Documentation
- [ ] `:type` keywords for component initialization
- [ ] `Source#absolute_certainty` needs to be written

lib/vanagon/component/source/git.rb
```diff
@@ -1,39 +1,69 @@
...
+        class << self
+          # Attempt to connect to whatever URL is provided and
+          # return True or False depending on whether or not
+          # `git` thinks it's a valid Git repo.
+          #
+          # @return [Boolean] whether #url is a valid Git repo or not
+          def valid_remote?(url)
+            !!::Git.ls_remote(url)
+          rescue ::Git::GitExecuteError
+            nil
```
haus: Since you're making it a boolean (in the yard, and explicitly above), should this return `false` instead?

mckern: Yeah, that makes some sense.

haus:
did you consider using refinements instead of monkey patching?

mckern:
Ruby 2.2/2.3 docs state:
> Refinements only modify classes, not modules so the argument must be a class.

They might work for this particular use case but it's a crapshoot.

stahnma:
@mckern are you planning to squash/clean commits more than this or is this what you want for final review?

mckern:
@stahnma I want to add a doc commit today but I wasn't planning to squash/rebase any more than this. These commits have a fair amount of churn in them so the best I could probably do would be to flatten it into one commit. I think we lose some important history here but I am open to being persuaded.

McdonaldSeanp:
FWIW I really don't think we should squash most of these. In the near future we may need to fix/update major parts to fit our needs and having the commit history will allow us to understand the original iterations and thought process so we can avoid the process of making a decision Ryan already realized was not going to work.

mckern:
`s/realized/assumed/`

stahnma:
Trying to build el-7-x86_64 off of PA stable. 

https://gist.github.com/stahnma/9df32ad6a8a510abaa03eb98a9469ab2

haus:
@mckern yea, monkey patching the gem is pretty safe. if it were touching anything from ruby core i'd be much more interested in limiting the scope of the effect.

mckern:
@stahnma thanks for helping test things off the happy path. You caught a thing I'd missed. Fixed and testing now.

lib/git/basic_submodules.rb
```diff
@@ -0,0 +1,36 @@
+require 'git'
+
+module BasicSubmodulePrimitives
+  # Extend Git::Lib to support shotgunning submodules. This command
+  # is not smart, and it has very poor support for submodule options.
+  # For example, you can't pass any arguments to the handful of submodule
+  # options that accept them (like --depth). We may extend it later,
+  # but for rev. 0001, simply initializing them will suffice
```
melissa: It might be useful to have an example of how this method will be used

mckern: Done.

mckern:
## High Level Breakdown

### Git
- Our custom `%x{git do-a-thing}` commands have been offloaded to an external library. This is an under-the-hood swap, and the existing API (such as it is) has been preserved while hopefully making it easier to replace the Git library when/if needed. 
- Git functionality is expanded, supporting virtually all variants of acceptable Git URIs (including the all-too-common Triplet format of `<user>@<host>:<repo>.git`).
- I tried (and failed) to make [Shallow Cloning](https://git-scm.com/docs/git-clone) a viable thing, but it turns out that shallow cloning has a TON of edge cases. [libgit2](https://github.com/libgit2/libgit2/issues/3058) doesn't even support them, so that should give you a reasonable idea of why I abandoned it. The commits and code related to shallow cloning have been left here to preserve the notes about why it was abandoned.

### HTTP source handling 
- HTTP sources now support redirects. Remote sources are now more likely to come down correctly the first time as a result of the new redirection support.
- Additionally, duplicate code between `Local` sources and `HTTP` sources has been cleaned up so that the `Local` class does the heavy lifting.
- Reduced code in the `HTTP` class will hopefully make it easier to replace/extend in the future if we have to move to a more performant HTTP backend than the native Net::HTTP implementation in Ruby.

### Local source handling
- Archive types have been expanded greatly, with support for bzip2, xz, rar, etc.

### `Component::Source` definition
- `Component::Source.source` now uses a cleaner heuristic to determine what sort of object to return: it asks each subclass if a given URI is valid and returns the first one that says "yes". In order, it tries Git, then HTTP/HTTPS, and then the Local filesystem before giving up.
- Refactoring use of class variables to class instance variables, which should make reasoning about scope a little easier as we move towards a better defined object model for Vanagon's guts.

### Misc
- Use of Ruby's `English` library, allowing us to use more descriptive values for common operations, e.g. checking `$CHILD_STATUS.success?` instead of `$?.success?`
- More status messages and error messages. Vanagon still exits with a stack dump if operations don't succeed but at least now the output is slightly more meaningful.
- Updated tests for code paths that were touched during this refactoring.

McdonaldSeanp:
👍 on everything, I believe with spec tests and all, this is ready for merge

PR #396, opened 2016-07-15
(CPR-237) Update rubocop to 0.41.2

melissa:
I had to make a few compromises in order to get both tests and rubocop
to pass. Unfortunately, a lot of the checks are set to fairly high
thresholds. Higher than I would prefer. This lets me get away with the
minimum amount of refactoring, but leaves us vulnerably to overly
complex methods, etc. in future code.


melissa:
Currently two offenses I'm trying to figure out the best way to fix:
```
Offenses:

lib/vanagon/component/dsl.rb:341:7: W: Lint/DuplicateMethods: Method Vanagon::Component::DSL#environment is defined at both /home/melissa/vanagon/lib/vanagon/component/dsl.rb:84 and /home/melissa/vanagon/lib/vanagon/component/dsl.rb:341.
      def environment(env)
      ^^^
lib/vanagon/platform/windows.rb:377:31: C: Performance/StringReplacement: Use tr instead of gsub.
        formatted_path = path.gsub(/\\/, "\/")
                              ^^^^^^^^^^^^^^^^

48 files inspected, 2 offenses detected
```
The first is going to be challenging. The second shouldn't be triggering the rubocop failure.

From http://www.rubydoc.info/gems/rubocop/0.27.0/RuboCop/Cop/Performance/StringReplacement
```
@bad
'abc'.gsub('b', 'd')
'abc'.gsub('a', '')
'abc'.gsub(/a/, 'd')
'abc'.gsub!('a', 'd')

@good
'abc'.gsub(/.*/, 'a')
'abc'.gsub(/a+/, 'd')
'abc'.tr('b', 'd')
'a b c'.delete(' ')
```

mckern:
I think I have an "update rubocop" branch floating around somewhere that I can compare this to. What's the priority on this?

melissa:
Not super high. I'd like to finish it up before the end of the sprint though

lib/vanagon/platform/rpm.rb
```diff
@@ -62,11 +62,11 @@ def add_repository(definition)
           else
             reponame = "#{SecureRandom.hex}-#{File.basename(definition.path)}"
             reponame = "#{reponame}.repo" if File.extname(reponame) != '.repo'
-            if is_cisco_wrlinux?
-              commands << "curl -o '/etc/yum/repos.d/#{reponame}' '#{definition}'"
-            else
-              commands << "curl -o '/etc/yum.repos.d/#{reponame}' '#{definition}'"
-            end
+            commands << if is_cisco_wrlinux?
```
stahnma: I find this way less readable.

lib/vanagon/platform/rpm/sles.rb
```diff
@@ -10,21 +10,21 @@ class SLES < Vanagon::Platform::RPM
         # @return [Array] A list of commands to add a zypper repo for the build system
         def add_repository(definition)
           definition = URI.parse(definition)
-          if @os_version == '10'
-            flag = 'sa'
-          else
-            flag = 'ar'
-          end
+          flag = if @os_version == '10'
```
stahnma: I also dislike this.

lib/vanagon/platform/solaris_10.rb
```diff
@@ -27,13 +27,15 @@ def generate_package(project)
...
         # in favor of just letting the makefile deliver the bill-of-materials
         # to the correct directory. This shouldn't be required at all then.
-        if project.bill_of_materials.nil?
-          bom_install = [# Move bill-of-materials into a docdir
-          "mkdir -p $(tempdir)/#{name_and_version}/usr/share/doc/#{project.name}",
-          "mv $(tempdir)/#{name_and_version}/bill-of-materials $(tempdir)/#{name_and_version}/usr/share/doc/#{project.name}/bill-of-materials",]
-        else
-          bom_install = []
-        end
+        bom_install = if project.bill_of_materials.nil?
```
stahnma: 🙅

lib/vanagon/platform/windows.rb
```diff
@@ -281,11 +282,11 @@ def add_repository(definition)
         commands = []
 
         if definition.scheme =~ /^(http|ftp|file)/
-          if File.extname(definition.path) == '.ps1'
-            commands << %(powershell.exe -NoProfile -ExecutionPolicy Bypass -Command 'iex ((new-object net.webclient).DownloadString(\"#{definition}\"))')
-          else
-            commands << %(C:/ProgramData/chocolatey/bin/choco.exe source add -n #{definition.host}-#{definition.path.gsub('/', '-')} -s "#{definition}" --debug || echo "Oops, it seems that you don't have chocolatey installed on this system. Please ensure it's there by adding something like 'plat.add_repository 'https://chocolatey.org/install.ps1'' to your platform definition.")
-          end
+          commands << if File.extname(definition.path) == '.ps1'
```
stahnma: no me gusta

stahnma:
Left my comments. I assume they're more for the cop than for @melissa's opinions. Nothing blocking, just some things look more difficult. I also fear change.

lib/vanagon/platform/rpm/sles.rb
melissa:
```
lib/vanagon/platform/rpm/sles.rb:13:11: C: Style/ConditionalAssignment: Use the return of the conditional for variable assignment and comparison.
          if @os_version == '10' ...
          ^^^^^^^^^^^^^^^^^^^^^^
```
http://www.rubydoc.info/gems/rubocop/0.41.2/RuboCop/Cop/Style/ConditionalAssignment describes the cop


It looks like I might be able to set 
```
Style/ConditionalAssignment:
  EnforcedStyle: assign_inside_condition
```
but I'm unsure

melissa:
@stahnma I agree with your assessment. I'm trying to figure out how to customize the cop to allow us to do that. It should, but it's not cooperating :(

.rubocop.yml
```diff
@@ -512,3 +512,21 @@ Style/Encoding:
 # so once RE-5565 is resolved, this should be enabled.
 Metrics/PerceivedComplexity:
   Enabled: true
+
+# This defaults to a much smaller number, but correcting the offending methods
+# will take a bit of work, so let's just not write anything more complex than
+# what we already have!
+Metrics/AbcSize:
+  Max: 45
```
mckern: Consider instead leaving this set to the default, letting it flag the offending methods, and then using inline overrides to note them in comments:

```
def someinsanemethod # rubocop:disable Metrics/AbcSize
```

(I think this will work; untested but I'll test it today)

.rubocop.yml
```diff
@@ -512,3 +512,21 @@ Style/Encoding:
...
+
+# We have a few modules that are much larger than the default limit even when
+# we don't count the comments
+Metrics/ModuleLength:
+    CountComments: false
+    Max: 200
```
mckern: I'm not sure this check is even valuable for us.

.rubocop.yml
```diff
@@ -512,3 +512,21 @@ Style/Encoding:
...
+
+# I have no idea why, but it's complaining that we aren't using octal notation
+# correctly when specifying file permissions. I have no idea how to resolve the
+# complaint without disabling this cop
+Style/NumericLiteralPrefix:
+  Enabled: false
```
mckern: [Here's the doc](https://github.com/bbatsov/ruby-style-guide#numeric-literal-prefixes)

I don't love this.

lib/vanagon/engine/base.rb
```diff
@@ -76,7 +76,7 @@ def retrieve_built_artifact
       def validate_platform
         missing_attrs = []
         @required_attributes.each do |attr|
-          if (!@platform.instance_variables.include?("@#{attr}".to_sym)) or @platform.instance_variable_get("@#{attr}".to_sym).nil?
+          if !@platform.instance_variables.include?("@#{attr}".to_sym) or @platform.instance_variable_get("@#{attr}".to_sym).nil?
```
mckern: If we're in here anyway, changing `or` to `||` is nevera bad idea.

lib/vanagon/platform/rpm.rb
```diff
@@ -48,7 +48,7 @@ def rpm_defines
         defines =  %(--define '_topdir $(tempdir)/rpmbuild' )
         # RPM doesn't allow dashes in the os_name. This was added to
         # convert cisco-wrlinux to cisco_wrlinux
-        defines << %(--define 'dist .#{@os_name.gsub('-', '_')}#{@os_version}' )
+        defines << %(--define 'dist .#{@os_name.tr('-', '_')}#{@os_version}' )
```
mckern: I had to double check, but `String#tr` was added in 1.9.x, so that's more than usable for our needs here.

lib/vanagon/utilities.rb
```diff
@@ -281,7 +280,7 @@ def local_command(command, return_command_output: false)
       clean_environment do
         puts "Executing '#{command}' locally"
         if return_command_output
-          ret = %x(#{command}).chomp
+          ret = `#{command}`.chomp
```
mckern: I kind of prefer `%x()`; [we can tune that](https://github.com/bbatsov/rubocop/blob/master/config/default.yml) using this in rubocop.yaml:

```yaml
Style/CommandLiteral:
  EnforcedStyle: percent_x
```

lib/vanagon/platform/windows.rb
mckern: We have a lot of these. We can rewrite them by splitting things into discrete methods or using ternary operators, or disable that particular Cop. I'm leaning towards "disable that particular Cop", but maybe doing so inline so we've got a searchable record of where the bodies are buried.

shrug: 👍 , I also find these less readable, I think disabling this cop is probably the best way forward for now, if it's not giving us any real value

mckern: Here's how we'd disable them/tune them in rubocop.yaml:

```yaml
Style/ConditionalAssignment:
  EnforcedStyle: assign_to_condition
  SupportedStyles:
    - assign_to_condition
    - assign_inside_condition
  # When configured to `assign_to_condition`, `SingleLineConditionsOnly`
  # will only register an offense when all branches of a condition are
  # a single line.
  # When configured to `assign_inside_condition`, `SingleLineConditionsOnly`
  # will only register an offense for assignment to a condition that has
  # at least one multiline branch.
  SingleLineConditionsOnly: true
```

Inline would be:

```ruby
commands << %(powershell.exe -NoProfile -ExecutionPolicy Bypass -Command 'iex ((new-object net.webclient).DownloadString(\"#{definition}\"))') # rubocop:disable Style/ConditionalAssignment
```

lib/vanagon/platform/rpm/sles.rb
mckern: I left some notes about this but I am strongly in favor of inline disabled Cops.

mckern:
I left some notes. I am so glad you're moving this forward. Thank you!

melissa:
@stahnma @mckern updated

lib/vanagon/platform/windows.rb
melissa: I ended up just disabling this cop, as it was throwing too many errors to be useful and enforce the style we want

lib/vanagon/component/dsl.rb
```diff
@@ -54,35 +54,28 @@ def method_missing(method, *args)
...
       #
       # @param block [Proc] the command(s) required to install the component
       def install(&block)
-        @component.install << block.call
-      end
-
-      # Setup any specific environment required to configure, build or install the component
-      #
-      # @param block [Proc] the environment required to configure, build or install the component
-      def environment(&block)
```
McdonaldSeanp: Why is this function going away?

melissa: It's a duplicate of line 334

McdonaldSeanp: 👍

McdonaldSeanp:
This all seems reasonable to me 👍

mckern:
I am generally 👍 but I have an important question: have we built anything with this yet? It's not a blocker, but that's always the best test of functionality.

melissa:
Good question! I just built puppet-agent for el-5-i386 and I am currently building puppet-agent for ubuntu-16.04-amd64. Are there any other platforms or projects you would like me to test?

mckern:
WE'LL DO IT LIVE.

PR #397, opened 2016-07-20
(maint) Update platform output directory

McdonaldSeanp:
In order to facilitate better output directory control, add the ability to
specify platform.output_dir in the platform config file and override any
defaults for output.

As part of this, to make vanagon more reusable, remove all platform specific
"defaults" that were there and create one single "default" for each platform
following the following pattern:

@platform/@platform_version/(optional)@target_repo/@plaform_arch

### DIFF
--- a/lib/vanagon/platform.rb
+++ b/lib/vanagon/platform.rb
@@ -8,7 +8,7 @@ class Vanagon
     attr_accessor :docker_image, :ssh_port, :rpmbuild, :install, :platform_triple
     attr_accessor :target_user, :package_type, :find, :sort, :build_hosts, :copy, :cross_compiled
     attr_accessor :aws_ami, :aws_user_data, :aws_shutdown_behavior, :aws_key_name, :aws_region, :aws_key
-    attr_accessor :aws_instance_type, :aws_vpc_id, :aws_subnet_id
+    attr_accessor :aws_instance_type, :aws_vpc_id, :aws_subnet_id, :output_dir
 
     # Platform names currently contain some information about the platform. Fields
     # within the name are delimited by the '-' character, and this regex can be used to
@@ -115,6 +115,16 @@ class Vanagon
       if instance_variable_get("@#{key}")
         instance_variable_get("@#{key}")
       end
+    end
+
+    # Get the output dir for packages. If the output_dir was defined already (by
+    # the platform config) then don't change it.
+    #
+    # @param target_repo [String] optional repo target for built packages defined
+    #   at the project level
+    # @return [String] relative path to where packages should be output to
+    def output_dir(target_repo = "")
+      @output_dir ||= File.join(@os_name, @os_version, target_repo, @architecture)
     end
 
     # Sets and gets the name of the operating system for the platform.
--- a/spec/lib/vanagon/platform_spec.rb
+++ b/spec/lib/vanagon/platform_spec.rb
@@ -4,22 +4,31 @@ describe "Vanagon::Platform" do
   let(:platforms) do
     [
       {
-        :name         => "debian-6-i386",
-        :os_name      => "debian",
-        :os_version   => "6",
-        :architecture => "i386",
+        :name                   => "debian-6-i386",
+        :os_name                => "debian",
+        :os_version             => "6",
+        :architecture           => "i386",
+        :output_dir             => "debian/6/i386",
+        :output_dir_with_target => "debian/6/thing/i386",
+        :block                  => %Q[ platform "debian-6-i386" do |plat| end ],
       },
       {
-        :name         => "el-5-i386",
-        :os_name      => "el",
-        :os_version   => "5",
-        :architecture => "i386",
+        :name                   => "el-5-i386",
+        :os_name                => "el",
+        :os_version             => "5",
+        :architecture           => "i386",
+        :output_dir             => "el/5/i386",
+        :output_dir_with_target => "el/5/thing/i386",
+        :block                  => %Q[ platform "el-5-i386" do |plat| end ],
       },
       {
-        :name         => "CumulusLinux-2.2-amd64",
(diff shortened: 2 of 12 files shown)


ScottGarman:
:+1:

melissa:
After talking with Sean about this, I think it makes sense to move forward with. I'd like to get a few more opinions on it though. It is technically a breaking change, and we'll have to coordinate all the vanagon projects to update the default pathing to maintain the current settings once the projects switch to the version of vanagon that contains this change.

@stahnma I'd definitely like to get your opinion on this. We'll have to make sure to communicate this change so anyone using vanagon to build a project knows to expect it and any future vanagon projects know to customize these paths to match with our current convention

McdonaldSeanp:
For context for everyone: My thought process on the new default is that the functionality to fit output directories with our own infrastructure does not belong in vanagon. So yes, there is going to need to be updates to things like debian platforms and solaris etc. etc. brought on by this change. And these updates will need to happen in multiple projects. However, if the only argument against this change is "it's going to be a pain to change all those platform configs in the other projects" IMO the idea of a more consistent user experience trumps that argument. Especially since the change to those projects is not actually that large.

stahnma:
I don't hate this, but I have a few questions. 
1. @mckern can you look it over?
2. When can we build out a set of platforms that is re-usable across projects? Seems like we could have a gem or something that has basic platform definitions and then we'd get updates like this automatically.

McdonaldSeanp:
As for question #2, it feels like realistically we could ticket that out quickly, make some decisions and do it in a sprint or two. Also +1000 to that idea

lib/vanagon/platform.rb
```diff
@@ -8,7 +8,7 @@ class Platform
     attr_accessor :docker_image, :ssh_port, :rpmbuild, :install, :platform_triple
     attr_accessor :target_user, :package_type, :find, :sort, :build_hosts, :copy, :cross_compiled
     attr_accessor :aws_ami, :aws_user_data, :aws_shutdown_behavior, :aws_key_name, :aws_region, :aws_key
-    attr_accessor :aws_instance_type, :aws_vpc_id, :aws_subnet_id
+    attr_accessor :aws_instance_type, :aws_vpc_id, :aws_subnet_id, :output_dir
```
mckern: I'd probably set `:output_dir` as an `attr_writer` because the corresponding accessor method is being explicitly defined, but that's the pedantry talking.

lib/vanagon/platform.rb
```diff
@@ -117,6 +117,20 @@ def [](key)
...
+    #
+    # @param target_repo [String] optional repo target for built packages defined
+    #   at the project level
+    # @return [String] relative path to where packages should be output to
+    def output_dir(target_repo = "")
+      if @output_dir
+        @output_dir
+      else
+        File.join(@os_name, @os_version, target_repo, @architecture)
+      end
```
mckern: This could be more concisely represented like this:

```
@output_dir ||= File.join(@os_name, @os_version, target_repo, @architecture)
```

That will both use the existing value if it's defined, and populate the instance variable with the computed value if it's not -- then future calls to `#output_dir` will return the populated instance variable instead of recomputing every run-through. It's not functionally identical to the existing code (which will **always** return the computed value if `@output_dir` isn't truthy) but it's probably a preferred way to handle this value since we're pretty lazy about using the accessor method instead directly calling the instance variable.

mckern: Also, I did validate that an empty string in `target_repo` will implicitly do the right thing when run through File.join:

```
>> target_repo = ""
=> ""
>> @os_name = 'debian'
=> "debian"
>> @os_version = 'jessie'
=> "jessie"
>> @architecture = 'amd64'
=> "amd64"
>> File.join(@os_name, @os_version, target_repo, @architecture)
=> "debian/jessie/amd64"
>>
```

spec/lib/vanagon/platform_spec.rb
```diff
@@ -4,22 +4,31 @@
...
-        :name         => "debian-6-i386",
-        :os_name      => "debian",
-        :os_version   => "6",
-        :architecture => "i386",
+        :name                   => "debian-6-i386",
+        :os_name                => "debian",
+        :os_version             => "6",
+        :architecture           => "i386",
+        :output_dir             => "debian/6/i386",
+        :output_dir_with_target => "debian/6/thing/i386",
```
mckern: This was originally in the form `deb/lucid/thing` but this test indicates that we should expect thesecomputed values to transition to a form like `debian/6/thing/i386`; will this require updates to the packaging repo to keep things from blowing up when we try to ship Vanagon packages built from this point on? It's late and I'm fuzzy on the interactions here.

McdonaldSeanp: Yeah, so the deal here is actually that we are going to have to add an override to each platform config for debian platforms in the projects themselves to conform to stuff like `deb/lucid/thing` . The intent of this PR is not to change any of the output paths, but instead move the functionality of specialized output pathing out of vanagon and in to the projects.

McdonaldSeanp: So to actually answer the question, no we will not change anything in packaging, we will just be adding to the platform configs in the project.

lib/vanagon/platform.rb
McdonaldSeanp: This seems reasonable. Update to come.

mckern:
Tentatively 👍 ; have you produced a build with the new code?

McdonaldSeanp:
Yep, test build for Xenial worked both with the new default and the override to conform to old standards.

shrug:
This broke RPM ships where they don't override yum_repo - they used to default to 'products' , now there is no default.

Opened https://tickets.puppetlabs.com/browse/RE-7865

PR #398, opened 2016-07-25
(PA-432) Constrain removing brp-strip-static to cross-compiled platforms

shrug:


mckern:
This looks reasonable. 👍

resources/rpm/project.spec.erb
```diff
@@ -5,8 +5,9 @@
 # Disable brp-strip-static-archive, which on EL platforms causes
 # cross-compiled static libs to end up with "no machine" as their
 # architure, breaking builds:
+<% if @platform.is_cross_compiled_linux? -%>
```
stahnma: Put the comment in this block?

shrug: Done. Now we wait another 6 hours for travis...

shrug:
Building with this, at least for cisco-wrlinux-7 I still wind up with an unstripped static libruby

mckern:
We might have to test out stripping as part of the component's build process; not in love with that idea but also not strictly opposed.

stahnma:
AFAICT, this size change isn't from strippping libs or not stripping them.

shrug:
No, it is. libruby-static.a in 1.5.2 had no debug symbols, in 1.5.3 it does (on el-7).

```
121MB	./1.5.2
105MB	./1.5.2/opt/puppetlabs/puppet/lib
164MB	./1.5.3
148MB	./1.5.3/opt/puppetlabs/puppet/lib
```

shrug:
I was wrong, this PR does seem to fix the issue, my bundle hadn't pulled in the updated vanagon. Built with this PR, libruby-static.a has no debug symbols and the size is back to what we expect:

```
121MB	./test
105MB	./test/opt/puppetlabs/puppet/lib
```

mckern:
Then my 👍 stands

PR #406, opened 2016-08-15
(RE-5403) Modify Vanagon to support more than md5 sum

sarameisburger:
Before, we were only verifying http sources use md5. Modified vanagon to support multiple sum types by adding a sum_type parameter. We now support sha256, sha512, and md5.

### DIFF
--- a/lib/vanagon/component/dsl.rb
+++ b/lib/vanagon/component/dsl.rb
@@ -278,6 +278,23 @@ class Vanagon
       # @param md5 [String] md5 sum of the source for verification
       def md5sum(md5)
         @component.options[:sum] = md5
+        @component.options[:sum_type] = "md5"
+      end
+
+      # Sets the sha256 sum to verify the sum of the source
+      #
+      # @param sha256 [String] sha256 sum of the source for verification
+      def sha256sum(sha256)
+        @component.options[:sum] = sha256
+        @component.options[:sum_type] = "sha256"
+      end
+
+      # Sets the sha512 sum to verify the sum of the source
+      #
+      # @param sha512 [String] sha512 sum of the source for verification
+      def sha512sum(sha512)
+        @component.options[:sum] = sha512
+        @component.options[:sum_type] = "sha512"
       end
 
       # Sets the ref of the source for use in a git source
--- a/lib/vanagon/component/source/http.rb
+++ b/lib/vanagon/component/source/http.rb
@@ -10,7 +10,7 @@ class Vanagon
         include Vanagon::Utilities
 
         # Accessors :url, :file, :extension, :workdir, :cleanup are inherited from Local
-        attr_accessor :sum
+        attr_accessor :sum, :sum_type
 
         class << self
           def valid_url?(target_url) # rubocop:disable Metrics/AbcSize
@@ -42,14 +42,19 @@ class Vanagon
         # @param url [String] url of the http source to fetch
         # @param sum [String] sum to verify the download against
         # @param workdir [String] working directory to download into
+        # @param sum_type [String] type of sum we are verifying
         # @raise [RuntimeError] an exception is raised is sum is nil
-        def initialize(url, sum:, workdir:, **options)
+        def initialize(url, sum:, workdir:, sum_type:, **options)
           unless sum
             fail "sum is required to validate the http source"
+          end
+          unless sum_type
+            fail "sum_type is required to validate the http source"
           end
           @url = url
           @sum = sum
           @workdir = workdir
+          @sum_type = sum_type
         end
 
         # Download the source from the url specified. Sets the full path to the
@@ -65,12 +70,12 @@ class Vanagon
         # Verify the downloaded file matches the provided sum
         #
         # @raise [RuntimeError] an exception is raised if the sum does not match the sum of the file
-        def verify
+        def verify # rubocop:disable Metrics/AbcSize
           puts "Verifying file: #{file} against sum: '#{sum}'"
-          actual = get_md5sum(File.join(workdir, file))
+          actual = get_sum(File.join(workdir, file), sum_type)
           return true if sum == actual
 
-          fail "Unable to verify '#{File.join(workdir, file)}': md5sum mismatch (expected '#{sum}', got '#{actual}'')"
+          fail "Unable to verify '#{File.join(workdir, file)}': #{sum_type} mismatch (expected '#{sum}', got '#{actual}')"
         end
 
         # Downloads the file from @url into the @workdir
--- a/lib/vanagon/utilities.rb
+++ b/lib/vanagon/utilities.rb
@@ -19,7 +19,7 @@ class Vanagon
     # @param file [String] file to md5sum
     # @return [String] md5sum of the given file
     def get_md5sum(file)
-      Digest::MD5.file(file).hexdigest.to_s
+      get_sum(file, 'md5')
     end
 
     # Generic file summing utility
@@ -32,6 +32,8 @@ class Vanagon
       case type.downcase
       when 'md5'
         Digest::MD5.file(file).hexdigest.to_s
+      when 'sha256'
+        Digest::SHA256.file(file).hexdigest.to_s
       when 'sha512'
         Digest::SHA512.file(file).hexdigest.to_s
       else
--- a/spec/lib/vanagon/component/source/http_spec.rb
+++ b/spec/lib/vanagon/component/source/http_spec.rb
@@ -10,21 +10,59 @@ describe "Vanagon::Component::Source::Http" do
   let (:plaintext_url) { "#{base_url}/#{plaintext_filename}" }
   let (:plaintext_dirname) { './' }
   let (:md5sum) { 'abcdssasasa' }
+  let (:sha256sum) { 'foobarbaz' }
+  let (:sha512sum) { 'teststring' }
   let (:workdir) { "/tmp" }
 
   describe "#dirname" do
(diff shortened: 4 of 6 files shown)


underscorgan:
heeeey @mckern ... wanna help with rubocoppery?

mckern:
Yes. Yes I do.

underscorgan:
other than that, 👍 though

lib/vanagon/component/source/http.rb
```diff
@@ -67,10 +72,10 @@ def file
         # @raise [RuntimeError] an exception is raised if the sum does not match the sum of the file
         def verify
```
mckern: Go ahead and add `# rubocop:disable Metrics/AbcSize` to the end of this line:

```ruby
 def verify # rubocop:disable Metrics/AbcSize
```

underscorgan:
🎉  👍

spec/lib/vanagon/component/source_spec.rb
```diff
@@ -108,7 +108,7 @@
           'http://buildsources.delivery.puppetlabs.net'
       end
       it "applies rewrite rules to HTTP URLs" do
-        expect(klass.source(original_http_url, sum: sum, workdir: workdir).url)
+        expect(klass.source(original_http_url, sum: sum, workdir: workdir, sum_type: "md5").url)
           .to eq(rewritten_http_url)
       end
     end
```
mckern: Can we see some tests around the new SHA sum methods? Even just validating that things work as expected would make me 😄 .

lib/vanagon/component/dsl.rb
```diff
@@ -276,8 +276,25 @@ def url(the_url)
       # Sets the md5 sum to verify the sum of the source
       #
       # @param md5 [String] md5 sum of the source for verification
+      # md5 is our default
```
mckern: For the sake of YARDoc, I'd move this comment to line 277 or at the end of line 276. It's a nitpick, but the rendered YARDoc will change where it's displayed based on that.

lib/vanagon/component/dsl.rb
```diff
@@ -276,8 +276,25 @@ def url(the_url)
...
       def md5sum(md5)
         @component.options[:sum] = md5
+        @component.options[:sum_type] = "md5"
+      end
```
mckern: Should we expose corresponding `#sha256sum` and `#sha512sum` methods? And if we should, should we not expose the `#sum` and `#sum_type` methods?

lib/vanagon/component/source/http.rb
```diff
@@ -65,12 +70,12 @@ def file
...
         # @raise [RuntimeError] an exception is raised if the sum does not match the sum of the file
-        def verify
+        def verify # rubocop:disable Metrics/AbcSize
           puts "Verifying file: #{file} against sum: '#{sum}'"
-          actual = get_md5sum(File.join(workdir, file))
+          actual = get_sum(File.join(workdir, file), sum_type)
           return true if sum == actual
 
-          fail "Unable to verify '#{File.join(workdir, file)}': md5sum mismatch (expected '#{sum}', got '#{actual}'')"
+          fail "Unable to verify '#{File.join(workdir, file)}': #{sum_type} mismatch (expected '#{sum}', got '#{actual}'')"
```
mckern: It's a prexisting typo, that I probably made, but can you clean up the extra single quote in `'#{actual}''` while you're in here? As a favor to me, maybe? Please & thank you.

lib/vanagon/utilities.rb
```diff
@@ -34,6 +34,8 @@ def get_sum(file, type)
         Digest::MD5.file(file).hexdigest.to_s
       when 'sha512'
         Digest::SHA512.file(file).hexdigest.to_s
+      when 'sha256'
+        Digest::SHA256.file(file).hexdigest.to_s
```
mckern: I love that we already had SHA512 and just never exposed it. We're great.

spec/lib/vanagon/component/source_spec.rb
mckern: Disregard my previous comment about tests; until we move checksumming back into the `Local` class, testing it in the `HTTP` will require a bunch of mocks and stubs.

underscorgan: aww, we just finished testing :(

mckern: Then let's merge some tests! I am OK being wrong here.

underscorgan: I mean, there's definitely stubs. but it didn't end up being the worst.

underscorgan:
@mckern updated! travis is still running, but want to take a look? I think we addressed all your comments!

lib/vanagon/component/dsl.rb
```diff
@@ -278,6 +278,23 @@ def url(the_url)
...
+        @component.options[:sum_type] = "sha256"
+      end
+
+      # Sets the sha512 sum to verify the sum of the source
+      #
+      # @param sha512 [String] sha512 sum of the source for verification
+      def sha512sum(sha512)
+        @component.options[:sum] = sha512
+        @component.options[:sum_type] = "sha512"
       end
```
melissa: What if we need to provide multiple sums for something? Like, how old windows versions don't even recognize sha256, so we have to provide both sha1 and sha256? Given that this is vanagon, that's a little bit of a leap, but I've definitely seen providers provide multiple checksums. Just a thought. Open to discussion.

mckern: @melissa isn't that just for the certificate encryption at signing time? This is strictly an internal function to Vanagon, and will only be used to validate the contents of retrieved components before compilation.

melissa: yeah, it generally is. I'm just trying to think about ways we might shift in the future. This query is definitely not a blocker

lib/vanagon/utilities.rb
mckern: This is a non-blocking nitpick, but can `when 'sha256'` above `when 'sha512'`?

melissa:
👍  from me

spec/lib/vanagon/component/source/http_spec.rb
```diff
@@ -10,21 +10,57 @@
...
+    end
+
+    it "calls md5 digest when it's supposed to" do
+      Digest::MD5.any_instance.stub(:file).and_return(Digest::MD5.new)
+      Digest::MD5.any_instance.stub(:hexdigest).and_return(md5sum)
+      http_source = Vanagon::Component::Source::Http.new(plaintext_url, sum: md5sum, workdir: workdir, sum_type: "md5")
+      expect(http_source).to receive(:download).and_return(plaintext_filename)
+      http_source.fetch
+      http_source.verify
+    end
```
mckern: (nitpick) Add a newline

spec/lib/vanagon/component/source/http_spec.rb
```diff
@@ -10,21 +10,57 @@
...
       expect(http_source.dirname).to eq(plaintext_dirname)
     end
   end
+
+  describe "verify" do
+    it "fails with a bad sum_type" do
+      http_source = Vanagon::Component::Source::Http.new(plaintext_url, sum: md5sum, workdir: workdir, sum_type: "md4")
+      expect(http_source).to receive(:download).and_return(plaintext_filename)
+      http_source.fetch
+      expect{http_source.verify}.to raise_error(RuntimeError, /Don't know how to produce a sum/)
```
mckern: nitpick: Curly brackets means we're passing a block, so we're testing behavior. Can we add some space around the curly brackets to highlight that we're passing a block intentionally, à la `expect { http_source.verify } .to`?

mckern: Also, do we care about the contents of the `RuntimeError` message, or just that a `RuntimeError` was raised?

spec/lib/vanagon/component/source/http_spec.rb
```diff
@@ -10,21 +10,57 @@
...
+      http_source.verify
+    end
+    it "calls sha256 digest when it's supposed to" do
+      Digest::SHA256.any_instance.stub(:file).and_return(Digest::SHA256.new)
+      Digest::SHA256.any_instance.stub(:hexdigest).and_return(sha256sum)
+      http_source = Vanagon::Component::Source::Http.new(plaintext_url, sum: sha256sum, workdir: workdir, sum_type: "sha256")
+      expect(http_source).to receive(:download).and_return(plaintext_filename)
+      http_source.fetch
+      http_source.verify
+    end
```
mckern: (nitpick) Moar newlines

nic cage moar

mckern:
I left some nitpick comments/questions, but I am so into this.
its happening

spec/lib/vanagon/component/source/http_spec.rb
sarameisburger: @mckern I'm always a fan of more information, but if you think it's unnecessary I can remove it!

mckern: More information **is** often good, but for a test it means you've tightly coupled the expected error value when all you really care about is that an error (or at least the correct error) is raised. The message for the user isn't what's being tested; correct behavior is.

sarameisburger: Ah yeah, that makes sense. I was just thinking from an intern standpoint where I have little idea of what's going on, maybe more information would be helpful. But I see your point now about just wanting to make sure an error is being produced. Changing and committing now!

PR #408, opened 2016-09-15
[DO NOT MERGE] (RE-8057) Use .PHONY on component rules

McdonaldSeanp:
We uncovered an issue when installing a file with the same name as one of the
make targets. This commit updates the make invocation to include component
names in the list of .PHONY targets in the makefile which fixes the issue.

### DIFF
--- a/resources/Makefile.erb
+++ b/resources/Makefile.erb
@@ -62,4 +62,4 @@ clean: <%= @components.map {|comp| "#{comp.name}-clean" }.join(" ") %>
 
 clobber: <%= @components.map {|comp| "#{comp.name}-clobber" }.join(" ") %>
 
-.PHONY: clean clobber  <%= @components.map {|comp| "#{comp.name}-clean #{comp.name}-clobber" }.join(" ") %>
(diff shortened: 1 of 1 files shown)


branan:
Making the component targets `PHONY` is probably a separate hygiene fix (as -r is sufficient to resolve our issue), but I am +1 on both changes getting in

McdonaldSeanp:
Yeah, we should do this in two PR, going to update

branan:
flagging this as do not merge until we get some issues with openssl build sorted out

shrug:
@branan @McdonaldSeanp is this still blocked?

stahnma:
Does this actually fix the issue? One of the things I've noticed is that when a build fails and you restart it, you have to rebuild anything where the component target is the same as the directory name. In my experiments, adding that as a .PHONY didn't solve that problem. In fact, if you make all components a .PHONY you will have to rebuild the world every time, if I am understanding how this all works properly.

shrug:
ping @branan @McdonaldSeanp

shrug:
@McdonaldSeanp @branan Is this still desirable?

stahnma:
The maintainers of this project will close this PR if no activity in 10 days.

underscorgan:
Closing due to inactivity. If this is still needed please reopen or resubmit.

PR #409, opened 2016-09-20
(VANAGON-8) Add the ability to only build single components in vanagon

McdonaldSeanp:
The use case for devs is they need to only build a single component and it's
dependencies. This PR solves this by adding a filter_components function to
the project class that recursively compiles all the components build
dependencies (and the build dependencies' dependencies and so on) together in
to one array. The driver class will use that function and replace the project
objects' components array with the output of filter_components.

Vanagon is built well enough to handle the rest, all build steps and makefile
rules can remain the same.

This should be a strictly additive change, and should not affect any current
build processes.

Usage: bundle exec build <project> <platform> [--only-build <components>]
where <components> is a comma seperated list

### DIFF
--- a/lib/vanagon/driver.rb
+++ b/lib/vanagon/driver.rb
@@ -13,12 +13,13 @@ class Vanagon
     attr_accessor :platform, :project, :target, :workdir, :verbose, :preserve
     attr_accessor :timeout, :retry_count
 
-    def initialize(platform, project, options = { :configdir => nil, :target => nil, :engine => nil, :components => nil, :skipcheck => false, :verbose => false, :preserve => false }) # rubocop:disable Metrics/AbcSize
+    def initialize(platform, project, options = { :configdir => nil, :target => nil, :engine => nil, :components => nil, :skipcheck => false, :verbose => false, :preserve => false, :only_build => nil }) # rubocop:disable Metrics/AbcSize
       @verbose = options[:verbose]
       @preserve = options[:preserve]
 
       @@configdir = options[:configdir] || File.join(Dir.pwd, "configs")
       components = options[:components] || []
+      only_build = options[:only_build]
       target = options[:target]
       engine = options[:engine] || 'pooler'
 
@@ -26,11 +27,22 @@ class Vanagon
       @project = Vanagon::Project.load_project(project, File.join(@@configdir, "projects"), @platform, components)
       @project.settings[:verbose] = options[:verbose]
       @project.settings[:skipcheck] = options[:skipcheck]
+      filter_out_components(only_build) if only_build
       loginit('vanagon_hosts.log')
 
       load_engine(engine, @platform, target)
     rescue LoadError => e
       raise Vanagon::Error.wrap(e, "Could not load the desired engine '#{engine}'")
+    end
+
+    def filter_out_components(only_build)
+      # map each element in the only_build array to it's set of filtered components, then
+      # flatten all the results in to one array and set project.components to that.
+      @project.components = only_build.flat_map { |comp| @project.filter_component(comp) }.uniq
+      if @verbose
+        puts "Only building:"
+        @project.components.each { |comp| puts comp.name }
+      end
     end
 
     def load_engine(engine_type, platform, target)
--- a/lib/vanagon/optparse.rb
+++ b/lib/vanagon/optparse.rb
@@ -10,6 +10,7 @@ class Vanagon
         :skipcheck => ['--skipcheck', 'Skip the `check` stage when building components'],
         :preserve => ['-p', '--preserve', 'Whether to tear down the VM on success or not (defaults to false)'],
         :verbose => ['-v', '--verbose', 'Verbose output (does nothing)'],
+        :only_build => ["--only-build COMPONENT,COMPONENT,...", Array, 'Only build this array of components']
       }.freeze
 
     def initialize(banner, options = [])
--- a/lib/vanagon/project.rb
+++ b/lib/vanagon/project.rb
@@ -82,6 +82,39 @@ class Vanagon
       files.flatten.uniq
     end
 
+    # Returns a filtered out set of components only including those
+    # components necessary to build a specific component. This is a
+    # recursive function that will call itself until it gets to a
+    # component with no build requirements
+    #
+    # @param name [String] name of component to add. must be present in configdir/components and named $name.rb currently
+    # @return [Array] array of Vanagon::Component including only those required to build "name", or [] if there are none
+    def filter_component(name)
+      filtered_component = get_component(name)
+      return [] if filtered_component.nil?
+      included_components = [filtered_component]
+
+      unless filtered_component.build_requires.empty?
+        filtered_component.build_requires.each do |build_requirement|
+          unless get_component(build_requirement).nil?
+            included_components += filter_component(build_requirement)
+          end
+        end
+      end
+      included_components.uniq
+    end
+
+    # Gets the component with component.name = "name" from the list
+    # of project.components
+    #
+    # @param [String] component name
+    # @return [Vanagon::Component] the component with component.name = "name"
+    def get_component(name)
+      comps = @components.select { |comp| comp.name.to_s == name.to_s }
+      raise "ERROR: two or more components with the same name: #{comps.first.name}" if comps.size > 1
+      comps.first
+    end
+
     # Collects all of the requires for both the project and its components
     #
     # @return [Array] array of runtime requirements for the project
--- a/bin/build
+++ b/bin/build
@@ -2,7 +2,7 @@
 load File.expand_path(File.join(File.dirname(__FILE__), "..", "lib", "vanagon.rb"))
 
 optparse = Vanagon::OptParse.new("#{File.basename(__FILE__)} <project-name> <platform-name> [<target>] [options]",
-                                 [:workdir, :configdir, :engine, :preserve, :verbose, :skipcheck])
+                                 [:workdir, :configdir, :engine, :preserve, :verbose, :skipcheck, :only_build])
 options = optparse.parse! ARGV
 
 project = ARGV[0]
--- a/spec/lib/vanagon/project_spec.rb
+++ b/spec/lib/vanagon/project_spec.rb
@@ -11,13 +11,15 @@ describe 'Vanagon::Project' do
     end"
   }
 
-  before do
-    allow_any_instance_of(Vanagon::Project::DSL).to receive(:puts)
-    allow(Vanagon::Driver).to receive(:configdir).and_return(configdir)
-    allow(Vanagon::Component).to receive(:load_component).with('some-component', any_args).and_return(component)
-  end
 
   describe '#get_root_directories' do
+
+    before do
+      allow_any_instance_of(Vanagon::Project::DSL).to receive(:puts)
+      allow(Vanagon::Driver).to receive(:configdir).and_return(configdir)
+      allow(Vanagon::Component).to receive(:load_component).with('some-component', any_args).and_return(component)
+    end
+
     let(:test_sets) do
       [
         {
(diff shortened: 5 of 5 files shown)


stahnma:
Can you reformat the usage to look like man page <required argument> [optional argument] and the - means literal dash.

McdonaldSeanp:
Is that better @stahnma ?

stahnma:
Yes thanks

lib/vanagon/driver.rb
```diff
@@ -13,17 +13,19 @@ class Driver
     attr_accessor :platform, :project, :target, :workdir, :verbose, :preserve
     attr_accessor :timeout, :retry_count
 
-    def initialize(platform, project, options = { :configdir => nil, :target => nil, :engine => nil, :components => nil, :skipcheck => false, :verbose => false, :preserve => false }) # rubocop:disable Metrics/AbcSize
+    def initialize(platform, project, options = { :configdir => nil, :target => nil, :engine => nil, :components => nil, :skipcheck => false, :verbose => false, :preserve => false, :only_build => nil }) # rubocop:disable Metrics/AbcSize
```
mckern: This is a good candidate to refactor -- that options hash is
bananas-smaller

probably not important in this PR but SOON because this is starting to get out of control. For now, consider moving the `options` hash itself into an external value, and then loading it in:

```ruby

# This is a naive implementation since some values are
# booleans and some are nil. But it'll demonstrate a
# viable direction we can head towards in the shorter term.
# Rather than leave the options that are nil undefined for
# brevity, by specifying them here we are at least communicating
# what options we would expect #new to receive.
DEFAULT_OPTIONS = {
  configdir: nil,
  target: nil,
  engine: nil,
  components: nil,
  skipcheck: false,
  verbose: false,
  preserve: false,
  only_build: nil
}
def initialize(platform, project, options = DEFAULT_OPTIONS) # rubocop:disable Metrics/AbcSize
```

This still has the same shortcomings, like default values not being initialized for each key in the `options` hash when any values are passed (options hashes are all-or-nothing) but it at least keeps the options hash readable.

lib/vanagon/driver.rb
```diff
@@ -13,17 +13,19 @@ class Driver
...
       @verbose = options[:verbose]
       @preserve = options[:preserve]
 
       @@configdir = options[:configdir] || File.join(Dir.pwd, "configs")
       components = options[:components] || []
+      only_build = options[:only_build]
```
mckern: No default value?

lib/vanagon/driver.rb
```diff
@@ -13,17 +13,19 @@ class Driver
...
       target = options[:target]
       engine = options[:engine] || 'pooler'
 
       @platform = Vanagon::Platform.load_platform(platform, File.join(@@configdir, "platforms"))
       @project = Vanagon::Project.load_project(project, File.join(@@configdir, "projects"), @platform, components)
+      @project = filter_out_components(@project, only_build) unless only_build.nil?
```
mckern: isn't `unless only_build.nil?` the same as `if only_build`?

lib/vanagon/driver.rb
```diff
@@ -33,6 +35,19 @@ def initialize(platform, project, options = { :configdir => nil, :target => nil,
       raise Vanagon::Error.wrap(e, "Could not load the desired engine '#{engine}'")
     end
 
+    def filter_out_components(project, only_build)
+      filtered_components = []
+      only_build.each do |comp|
+        filtered_components += @project.filter_component(comp)
```
mckern: Should this be `@project` (instance variable) or `project` (passed in argument to this method)? Which one are you intending to call? Consider a more descriptive variable name to avoid conceptual (if not literal) shadowing?

lib/vanagon/driver.rb
```diff
@@ -33,6 +35,19 @@ def initialize(platform, project, options = { :configdir => nil, :target => nil,
...
+      end
+      project.components = filtered_components.uniq
```
mckern: Let's talk about `Enumerable#map`:

```ruby
only_build.map { |comp| @project.index(comp) }.uniq
```

lib/vanagon/driver.rb
```diff
@@ -33,6 +35,19 @@ def initialize(platform, project, options = { :configdir => nil, :target => nil,
...
+      puts "Only building:" if @verbose
+      project.components.each do |comp|
+        puts comp.name if @verbose
+      end
```
mckern: You will branch a lot less if you do don't keep calling `if` here:

```ruby
 if @verbose
  puts "Only building:"
  project.components.each { |comp| puts comp.name }
end
```

mckern: Tangent: we have GOT to decouple notification/logging from actual implementation. Vanagon needs a better `Logger`, stat.

lib/vanagon/driver.rb
McdonaldSeanp: Oh yeah

lib/vanagon/driver.rb
McdonaldSeanp: the default value should come from the default at https://github.com/McdonaldSeanp/vanagon/blob/7aec62908986544032b6ef83a82fb0edd943f71a/lib/vanagon/driver.rb#L16 I think?

lib/vanagon/driver.rb
McdonaldSeanp: Yeah, It should be @project

lib/vanagon/project.rb
```diff
@@ -82,6 +82,32 @@ def get_files
...
+      filter_component = get_component(name)
+      return nil if filter_component.nil?
+      components_to_include << filter_component
+      filter_component.build_requires.each do |build_requirement|
+        unless get_component(build_requirement).nil?
+          components_to_include += filter_component(build_requirement)
+        end
+      end unless filter_component.build_requires.nil?
+      components_to_include
+    end
```
mckern: There is a LOT going on in this method name, and one of your variable names is also the same as your method name. This is taking me a little longer to unpack than it should, so that's probably a code smell.

lib/vanagon/driver.rb
McdonaldSeanp: This doesn't make sense to me, is the map meant to replace the last 4 lines here?

mckern: Sorry, I gave you the final product of some of my tinkering, but I didn't give you the steps. The real chunk of code I should have thrown at you is:

```ruby
filtered_components = only_build.map { |comp| @project.filter_component(comp) }.uniq
```

lib/vanagon/driver.rb
```diff
@@ -33,6 +37,18 @@ def initialize(platform, project, options = { :configdir => nil, :target => nil,
       raise Vanagon::Error.wrap(e, "Could not load the desired engine '#{engine}'")
     end
 
+    def filter_out_components(only_build)
+      filtered_components = []
+      only_build.each do |comp|
+        filtered_components += @project.filter_component(comp)
+      end
```
mckern: No idea where the last comment I left went, but the gist of it is that we can boil this down some using `Enumerable#map` here:

```ruby
filtered_components = only_build.map { |comp| @project.filter_component(comp) }.uniq
```

lib/vanagon/project.rb
```diff
@@ -82,6 +82,32 @@ def get_files
...
+
+    def get_component(name)
+      unless @components.select { |comp| comp.name.to_s == name.to_s }.nil?
+        return @components.select { |comp| comp.name.to_s == name.to_s }.pop
+      end
+      nil
+    end
```
mckern: `#select` returns a new array with the value of whatever matched in the block, or empty otherwise. Checking for `empty?` would be more appropriate than `nil?` since an empty array is still truthy.

```ruby
>> components = [ { name: 'shazam' }, { name: 'mary marvel' }, { name: 'black adam' } ]
=> [
    [0] {
        :name => "shazam"
    },
    [1] {
        :name => "mary marvel"
    },
    [2] {
        :name => "black adam"
    }
]
>> components.select { |comp| comp[:name].to_s == 'shazam' }
=> [
    [0] {
        :name => "shazam"
    }
]
>> components.select { |comp| comp[:name].to_s == 'spider-man' }
=> []
>> components.select { |comp| comp[:name].to_s == 'spider-man' }.empty?
=> true
>> components.select { |comp| comp[:name].to_s == 'spider-man' }.nil?
=> false
```

lib/vanagon/project.rb
mckern: OK, I dug into `filter_component` a little bit and rewrote it to understand it. The important takeaways are: don't shadow variable names by reusing method names, and consider streamlining some of your operations a little bit by standardizing on the expected values are expected from outside your new code. Comments inline:


```ruby
def filter_component(name)
  filtered_component = get_component(name)
  return nil unless filtered_component

  included_components = [filtered_component]

  # is the default value of #build_requires `nil` or `false`?
  # if it's iteratable, shouldn't it just be []? So we'd test
  # for whether or not it's empty, not if it's nil. If it's
  # not [], then we should probably fix and formalize that.
  # here, I'm testing to see if it's #empty? but that's kind of
  # a sign that it's a wonky interface more than anything else.
  unless filtered_component.build_requires.empty?
    filtered_component.build_requires.each do |build_requirement|
      unless get_component(build_requirement).empty?
        included_components << filtered_component(build_requirement)
      end
    end
  end
  included_components
end

# I would allow this method to return an empty array rather than coercing it
# to nil and then I would standardize on that as an expected interface since
# iterating over an empty array is functionally a NOOP.
# 
# >> [].each { |v| puts v }
# => []
# >>
# 
def get_component(name)
  @components.select { |comp| comp.name.to_s == name.to_s }
end
```

This is probably broken or missing pieces since I don't have good data to feed to it here to test this out. But the idea is that if something is expected to be iterable, let it be returned as an empty iterable (basically anything inheriting from `Enumerable`). If it's imperative that you not attempt to operate on an empty `Enumerable`, then use `empty?` instead of `nil?`.

lib/vanagon/project.rb
McdonaldSeanp: Sounds good

lib/vanagon/optparse.rb
```diff
@@ -10,6 +10,7 @@ class OptParse
         :skipcheck => ['--skipcheck', 'Skip the `check` stage when building components'],
         :preserve => ['-p', '--preserve', 'Whether to tear down the VM on success or not (defaults to false)'],
         :verbose => ['-v', '--verbose', 'Verbose output (does nothing)'],
+        :only_build => ["--only-build [ARRAY]", Array, 'Only build this array of components']
```
McdonaldSeanp: It's a comma seperated list, I'm not sure the best way to display that though

lib/vanagon/project.rb
McdonaldSeanp: I did write spec tests, you could likely use those to test out yo' stuff

lib/vanagon/project.rb
McdonaldSeanp: So what I'm actually looking to do is return just the component, not an array consisting of only the component. I looked and couldn't find anything in ruby that allows me to select a specific element out of an array based on some criteria. (other than this)

Ideally the return value of this function should not be an array at all, so I believe we go back to using nil if the component is not in the array.

McdonaldSeanp:
@mckern This has been updated, please do another pass for review

lib/vanagon/project.rb
mckern: I thought that the `@components` instance variable in `Project` was initialized as an `Array` (or a `Set`) always; BRB, cross checking.

mckern: Cross-checking confirmed that `@project.components` is always an Array. So `#select` is always safe, and yes, it is the way to pluck a single item out of an array. But `#select` is going to result in an empty array if your item isn't in the array and it will always return an Array.

Here's your method as written:

```ruby
def get_component(name)
  unless @components.select { |comp| comp.name.to_s == name.to_s }.nil?
    return @components.select { |comp| comp.name.to_s == name.to_s }.pop
  end
  nil
end
```

`@components.select { |comp| comp.name.to_s == name.to_s }` is **always** an Array. So testing `#nil?`will always fail because an Array is never `nil`. If you wanted to test the value of the first (only?) component, you'd use `.first`:

```ruby
def get_component(name)
  @components.select { |comp| comp.name.to_s == name.to_s }.first
end
```

Would this do the thing you intended it to do? What about:

```ruby
def get_component(name)
  comps = @components.select { |comp| comp.name.to_s == name.to_s }
  raise "WHY ARE THERE SO MANY COMPONENTS WITH THE SAME NAME" if comps.size > 1
  comps.first
end

Comment (mckern) [inline lib/vanagon/project.rb]: (That error message in my example shouldn't be used, BTW)

**Comment by McdonaldSeanp** [comment]
Updated with your proposed updates.

**Comment by stahnma** [comment]
ping @mckern

File: lib/vanagon/driver.rb (hunk 1 above)
Comment (mckern) [inline lib/vanagon/driver.rb]: In the interest of closing this up, I'm willing to overlook this bananas hash but we should probably finally figure out how to build a proper configuration class for Vanagon.

File: lib/vanagon/driver.rb [hunk 11]
```diff
@@ -13,17 +13,21 @@ class Driver
...
     attr_accessor :timeout, :retry_count
 
-    def initialize(platform, project, options = { :configdir => nil, :target => nil, :engine => nil, :components => nil, :skipcheck => false, :verbose => false, :preserve => false }) # rubocop:disable Metrics/AbcSize
+    def initialize(platform, project, options = { :configdir => nil, :target => nil, :engine => nil, :components => nil, :skipcheck => false, :verbose => false, :preserve => false, :only_build => nil }) # rubocop:disable Metrics/AbcSize
       @verbose = options[:verbose]
       @preserve = options[:preserve]
 
       @@configdir = options[:configdir] || File.join(Dir.pwd, "configs")
       components = options[:components] || []
+      only_build = options[:only_build] || nil
```
mckern: This is functionally identical to `only_build = options[:only_build]`, isn't it? A non-existent key will still return `nil`.

lib/vanagon/driver.rb
```diff
@@ -13,17 +13,21 @@ class Driver
...
       target = options[:target]
       engine = options[:engine] || 'pooler'
 
       @platform = Vanagon::Platform.load_platform(platform, File.join(@@configdir, "platforms"))
       @project = Vanagon::Project.load_project(project, File.join(@@configdir, "projects"), @platform, components)
+      if only_build
+        filter_out_components(only_build)
+      end
```
mckern: It's a nitpick, but lines 28-30 could probably be moved under the current line 32 (`@project.settings[:skipcheck] = options[:skipcheck]`), so that filtering happens after all of the settings are initialized (filtering shouldn't depend on settings one way or the other). You could also collapse the `if` into a single line for the sake of conciseness:

```ruby
filter_out_components(only_build) if only_build
```

(I like using this approach when possible because it makes poor variable name choices immediately obvious when your brain tries to parse the instructions as plain English and balks at bad naming choices.)

lib/vanagon/project.rb
```diff
@@ -82,6 +82,39 @@ def get_files
       files.flatten.uniq
     end
 
+    # Returns a filtered out set of components only including those
+    # components necessary to build a specific component. This is a
+    # recursive function that will call itself until it gets to a
+    # component with no build requirements
+    #
+    # @param name [String] name of component to add. must be present in configdir/components and named $name.rb currently
+    # @return [Array] array of Vanagon::Component including only those required to build "name"
```
mckern: This isn't complete; it can also return`nil` (line 94), which is not iterable like an empty array would be. I'd actually return an empty array instead of `nil` to minimize the amount of type checking needed (instead of looking for `#nil?` **and** `#empty?`, you can just look for `#empty?` safely).

lib/vanagon/project.rb
```diff
@@ -82,6 +82,39 @@ def get_files
...
+    def filter_component(name)
+      filtered_component = get_component(name)
+      return nil if filtered_component.nil?
+      included_components = [filtered_component]
+
+      unless filtered_component.build_requires.nil?
```
mckern: Will `filtered_component.build_requires` be an iterable object like an array even if there's nothing to build? Or will it be another neither-fish-nor-fowl method that returns `nil` **or** an array?

mckern:
@McdonaldSeanp I finished up some review, as individual comments. I'm tired of fighting GitHub's new review features. I am making some time for this today, so we can talk about these suggestions IRL if you want to.

McdonaldSeanp:
Great, I'll push to get this through with you

stahnma:
@McdonaldSeanp any chance of this getting fixed up soon?

McdonaldSeanp:
@mckern updated

shrug:
ping @mckern

lib/vanagon/driver.rb
```diff
@@ -13,26 +13,38 @@ class Driver
...
 
       load_engine(engine, @platform, target)
     rescue LoadError => e
       raise Vanagon::Error.wrap(e, "Could not load the desired engine '#{engine}'")
     end
 
+    def filter_out_components(only_build)
+      # map each element in the only_build array to it's set of filtered components, then
+      # flatten all the results in to one array and set project.components to that.
+      @project.components = only_build.flat_map { |comp| @project.filter_component(comp) }.uniq
```
mckern: bad-smell

(not a blocker; just... whoof, any time you're running `#uniq` after `#map`, you know you should probably fix something further up in the code path.

lib/vanagon/project.rb
```diff
@@ -82,6 +82,39 @@ def get_files
       files.flatten.uniq
     end
 
+    # Returns a filtered out set of components only including those
+    # components necessary to build a specific component. This is a
+    # recursive function that will call itself until it gets to a
+    # component with no build requirements
```
mckern: resize-620_465

McdonaldSeanp:
@mckern updated

shrug:
Ping @mckern

mckern:
84e8e45c227eba202d4170a662887b1c

traylenator:
This will be very useful indeed.

PR #412, opened 2016-09-28
(PA-569) Enable brp strip scripts for Cisco platforms

ScottGarman:
Our cisco-5/7 build platforms are missing the definition of
__os_install_post, so we have to manually set it here. This
then enables the brp strip scripts, which are needed to reduce
the size of our agent packages for these platforms. This is
necessary as some targets will fail to even install the package
without stripped binaries because yum's package cache is kept
on a partition with ~50MB of free space.

### DIFF
--- a/resources/rpm/project.spec.erb
+++ b/resources/rpm/project.spec.erb
@@ -1,5 +1,21 @@
 %global debug_package %{nil}
 
+<%- if @platform.is_cisco_wrlinux? -%>
+# Our cisco-5/7 build platforms are missing the definition of
+# __os_install_post, so we have to manually set it here. This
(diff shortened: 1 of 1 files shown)


shrug:
This matches the el-7 macro, does it work (ie, is the build environment in wrlinux calling this correctly)?

ScottGarman:
@shrug yes, I have verified that the brp scripts are now called with this fix. I'll also note that the size of the p-a package for cisco-wrlinux-5 is reduced to 45 MB (from 52 MB) with this change.

McdonaldSeanp:
Rob has approved this, merging.

PR #414, opened 2016-09-29
(VANAGON-14) Use md5 as the default checksum type if not specified

ScottGarman:
Now that we support multiple checksum types for validating http sources,
assume md5 if sum_type isn't explicitly set.


shrug:
I'm not opposed to having a default, but would it be better to explicitly set the sum type to md5 and have the default be something more robust?

shrug:
ah, I actually read the ticket. If we're using md5 as the default in other methods then it makes sense to be consistent.

PR #417, opened 2016-10-13
(QENG-4456) sneak in support for scheduled hardware under a feature flag

rick:
We have an internal need for supporting the management of hardware build hosts through a common centralized scheduler, specifically [always-be-scheduling](https://github.com/puppetlabs/always-be-scheduling) (ABS).

Prior to this work, vanagon would manage hardware itself: configs would list the specific hardware resources which were known for building on a given platform/target. We would allocate the hardware instance, use the `lock_manager` to control access, etc.

In a world of scheduled hardware, allocation, lock management, setup, teardown, etc., are all Someone Else's Responsibility™.  This change allows authors to opt-in to an "always_be_scheduling" engine, which leaves the host management up to a third-party scheduler for certain platforms.

Additionally, as we wish [vmpooler](https://github.com/puppetlabs/vmpooler)-based vanagon builds in the always-be-scheduling world to allow ABS to manage assignment of vmpooler resources, here we allow the `always_be_scheduling` vanagon engine to manage those resources as well.

This is feature-flagged behind a command-line switch and can be enabled like so:

``` shell
$ build_host_info puppet-agent aix-5.3-ppc
{"name":"pe-aix-53-builder.example.com","engine":"hardware"}

$ build_host_info puppet-agent aix-5.3-ppc --engine always_be_scheduling
{"name":"aix-5.3-ppc","engine":"always_be_scheduling"}
```

In project platform configuration files we introduce the `abs_resource_name` attribute. When specified, the `always_be_scheduling` engine will use that `abs_resource_name` value as the returned platform name. If this is not available, the `vmpooler_template` value will be used.  If neither is available, the platform name string will be returned unchanged.

**tasks**
- [x] Create a `scheduled` engine
- [x] Add support for an `engine` verb in the platform DSL
- [x] Add an `engine` attribute on the platform class
- [x] add a feature flag hook method to allow enabling scheduled hardware hosts
- [x] update the driver class to use a scheduled engine when the feature flag is on and the 'engine' verb is used in a build config
- [x] fix order-dependent test failures from times of yore
- [x] fix deprecations in rspec from times of yore
- [x] remove extraneous test output from times of yore
- [x] get :eyes: on this, to make sure it's remotely sane
- [x] drop `ENV` feature-flagging in favor of specifying the vanagon engine to trigger the new behavior
- [x] review naming, select appropriate names for the domain
- [x] use `always-be-scheduling` engine name instead of `scheduled`
- [x] add more documentation on the intent of the new scheduler support
- [x] translate outbound platform names to be vmpooler-style (for abs consumption)
  - 🚦 **but** have more discussion on this, as I think there are still a number of things to sort out
- [x] use `always-be-scheduling` engine for both hardware and vmpooler nodes
- [x] now we can drop the inline `.engine` support
- [x] document the `abs_resource_name` option better
- [ ] get another round of :eyes: on this 🛥 
- [ ] squash commits before merging

cc @puppetlabs/cinext @mckern @stahnma @shrug @sbeaulie @jameswagner2015

### DIFF
--- a/lib/vanagon/driver.rb
+++ b/lib/vanagon/driver.rb
@@ -34,23 +34,31 @@ class Vanagon
     end
 
     def load_engine(engine_type, platform, target)
-      if platform.build_hosts
-        engine_type = 'hardware'
-      elsif platform.aws_ami
-        engine_type = 'ec2'
-      elsif platform.docker_image
-        engine_type = 'docker'
-      elsif target
-        engine_type = 'base'
+      if engine_type != 'always_be_scheduling'
+        if platform.build_hosts
+          engine_type = 'hardware'
+        elsif platform.aws_ami
+          engine_type = 'ec2'
+        elsif platform.docker_image
+          engine_type = 'docker'
+        elsif target
+          engine_type = 'base'
+        end
       end
       load_engine_object(engine_type, platform, target)
     end
 
     def load_engine_object(engine_type, platform, target)
       require "vanagon/engine/#{engine_type}"
-      @engine = Object::const_get("Vanagon::Engine::#{engine_type.capitalize}").new(platform, target)
+      @engine = Object::const_get("Vanagon::Engine::#{camelize(engine_type)}").new(platform, target)
     rescue
-      fail "No such engine '#{engine_type.capitalize}'"
+      fail "No such engine '#{camelize(engine_type)}'"
+    end
+
+    def camelize(string)
+      string.gsub(/(?:^|_)([a-z])?/) do |match|
+        (Regexp.last_match[1] || '').capitalize
+      end
     end
 
     def cleanup_workdir
--- a/lib/vanagon/engine/always_be_scheduling.rb
+++ b/lib/vanagon/engine/always_be_scheduling.rb
@@ -1,0 +1,92 @@
+require 'vanagon/engine/base'
+require 'json'
+
+class Vanagon
+  class Engine
+    # This engine allows build resources to be managed by the "Always be
+    # Scheduling" (ABS) scheduler (https://github.com/puppetlabs/always-be-scheduling)
+    #
+    # ABS expects to ask `build_host_info` for the needed resources for a build,
+    # and to have that return a platform name.  ABS will then acquire the
+    # desired build host resources and will later run a vanagon build, passing
+    # those resource hostnames in specifically.
+    #
+    # `build_host_info` will normally use the `hardware` engine when a hardware
+    # platform is queried. The `always_be_scheduling` engine's behavior will
+    # be invoked instead when:
+    #
+    # `build_host_info ... --engine always_be_scheduling` is specified on the
+    # command-line.
+    #
+    # Configuration:
+    #
+    # Project platform configurations can specify the platform name to be returned
+    # via the `abs_resource_name` attribute. If this is not set but `vmpooler_template`
+    # is set, then the `vmpooler_template` value will be used. Otherwise, the
+    # platform name will be returned unchanged.
+    #
+    # Example 1:
+    #
+    # platform 'ubuntu-10.04-amd64' do |plat|
+    #   plat.vmpooler_template 'ubuntu-1004-amd64'
+    # end
+    #
+    # $ build_host_info puppet-agent ubuntu-10.04-amd64
+    # {"name":"ubuntu-10.04-amd64","engine":"pooler"}
+    #
+    # $ build_host_info puppet-agent ubuntu-10.04-amd64 --engine always_be_scheduling
+    # {"name":"ubuntu-10.04-amd64","engine":"always_be_scheduling"}
+    #
+    #
+    # Example 2:
+    #
+    # platform 'aix-5.3-ppc' do |plat|
+    #   plat.build_host ['aix53-builder-1.example.com']
+    #   plat.abs_resource_name 'aix-53-ppc'
+    # end
+    #
+    # $ build_host_info puppet-agent aix-5.3-ppc
+    # {"name":"aix53-builder-1.example.com","engine":"hardware"}
+    #
+    # $ build_host_info puppet-agent aix-5.3-ppc --engine always_be_scheduling
+    # {"name":"aix-53-ppc","engine":"always_be_scheduling"}
+    #
+    #
+    # Example 3:
+    #
+    # platform 'aix-5.3-ppc' do |plat|
+    #   plat.build_host ['aix53-builder-1.example.com']
+    #   plat.vmpooler_template
+    #   plat.abs_resource_name 'aix-53-ppc'
+    # end
+    #
+    # $ build_host_info puppet-agent aix-5.3-ppc
+    # {"name":"aix53-builder-1.example.com","engine":"hardware"}
+    #
+    # $ build_host_info puppet-agent aix-5.3-ppc --engine always_be_scheduling
+    # {"name":"aix-53-ppc","engine":"always_be_scheduling"}
+    class AlwaysBeScheduling < Base
+      def initialize(platform, target)
+        super
+
+        Vanagon::Driver.logger.debug "AlwaysBeScheduling engine invoked."
+      end
+
+      # Get the engine name
+      def name
+        'always_be_scheduling'
+      end
+
+      # return the platform name as the "host" name
+      def build_host_name
+        if @platform.abs_resource_name
+          @platform.abs_resource_name
+        elsif @platform.vmpooler_template
+          @platform.vmpooler_template
+        else
+          @platform.name
+        end
+      end
+    end
+  end
+end
--- a/spec/lib/vanagon/driver_spec.rb
+++ b/spec/lib/vanagon/driver_spec.rb
@@ -54,6 +54,39 @@ describe 'Vanagon::Driver' do
                               'engine' => 'hardware' })
     end
 
+    it 'returns the first build_host using the hardware engine when not using an engine, and not specifying an engine in the platform configuration' do
+      platform = eval_platform('aix-7.1-ppc', <<-END)
+        platform 'aix-7.1-ppc' do |plat|
+          plat.build_host ["pe-aix-71-01", "pe-aix-71-02"]
+        end
+      END
+
+      info = create_driver(platform).build_host_info
+
+      expect(info).to match({ 'name'   => 'pe-aix-71-01',
+                              'engine' => 'hardware' })
+    end
+
+    it 'returns the platform and the always_be_scheduling engine when using the always_be_scheduling engine' do
+      platform = eval_platform('aix-7.1-ppc', <<-END)
+        platform 'aix-7.1-ppc' do |plat|
+          plat.build_host ["pe-aix-71-01", "pe-aix-71-02"]
+        end
+      END
+
+      info = create_driver(platform, :engine => 'always_be_scheduling').build_host_info
+
+      expect(info).to match({ 'name'   => 'aix-7.1-ppc',
+                              'engine' => 'always_be_scheduling' })
+    end
+
+    it 'returns the vmpooler_template using the always_be_scheduling engine when using the always_be_scheduling engine' do
+      info = create_driver(redhat, :engine => 'always_be_scheduling').build_host_info
+
+      expect(info).to match({ 'name'   => 'centos-7-x86_64',
+                              'engine' => 'always_be_scheduling' })
+    end
+
     it 'returns the docker_image using the docker engine' do
       platform = eval_platform('el-7-x86_64', <<-END)
         platform 'el-7-x86_64' do |plat|
@@ -85,4 +118,3 @@ describe 'Vanagon::Driver' do
     end
   end
 end
-
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -1,6 +1,14 @@
 RSpec.configure do |c|
   c.before do
-    allow($stdout).to receive(:puts)
-    allow($stderr).to receive(:puts)
+    allow_any_instance_of(Vanagon::Component::Source::Git).to receive(:puts)
+    allow_any_instance_of(Vanagon::Component::Source::Http).to receive(:puts)
+    allow_any_instance_of(Vanagon::Component::Source::Local).to receive(:puts)
+
+    class Vanagon
+      module Utilities
+        def puts(*args)
+        end
+      end
+    end
   end
 end
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ rvm:
   - 2.2.0
 
 env:
-  - "CHECK='rspec spec --color --format documentation --order random'"
+  - "CHECK='rspec spec --color --format documentation --order random --require spec_helper.rb'"
   - "CHECK='rubocop -D'"
 
 matrix:
--- a/lib/vanagon/engine/hardware.rb
+++ b/lib/vanagon/engine/hardware.rb
@@ -27,7 +27,7 @@ class Vanagon
         host
       end
 
-      # Iterarte over the options and find a node open to lock.
+      # Iterate over the options and find a node open to lock.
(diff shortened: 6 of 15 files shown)


rick:
I lost at code jenga with rubocop:

```
lib/vanagon/driver.rb:36:5: C: Metrics/PerceivedComplexity: Perceived complexity for load_engine is too high. [8/7]
    def load_engine(engine_type, platform, target)
    ^^^
```

nwolfe:
(copying from comment from Ryan McKern on the JIRA ticket)

Can we already get this functionality by providing a hostname on the CLI?

```bash
$ build_host_info puppet-agent aix-5.3-ppc f98109fjkljk2.delivery.puppetlabs.net
```

rick:
@nwolfe saw that and had a brief conversation with @mckern.  We are going to be providing the ultimately ABS-allocated host back to vanagon that way; it's not clear if there's a shorter path to people being able to say, in their vanagon configs, "I need an aix-5.3-ppc, and I'm not specifying build hosts, can you just make build_host_info return the platform?".  I didn't see one, but I'm no vanagon expert.

rick:
Bumping, as this appears to have stalled in review. Maybe the worst week to get a PR review (what with PuppetConf happening), but worth a shot.  @mckern, thoughts?

mckern:
@rick this week is a 🗑 . Today is definitely a 🔥 :
screen shot 2016-10-18 at 10 05 28 am

rick:
No worries, man -- it's That Week™  Be well, :fire: :skull: some things.

rick:
Bumping this as I'd still :heart: a review.

stahnma:
I doubt this is something that you did, but still:

```
Deprecation Warnings:

Using `any_instance` from rspec-mocks' old `:should` syntax without explicitly enabling the syntax is deprecated. Use the new `:expect` syntax or explicitly enable `:should` instead. Called from /home/stahnma/vanagon/spec/lib/vanagon/component/source/http_spec.rb:42:in `block (3 levels) in <top (required)>'.


If you need more of the backtrace for any of these deprecations to
identify where to make the necessary changes, you can configure
`config.raise_errors_for_deprecations!`, and it will turn the
deprecation warnings into errors, giving you the full backtrace.

1 deprecation warning total
```

stahnma:
Could we get some more comments on the what the scheduled engine is attempting to do. If I were using this (not involved in this conversation), getting build_host_name to return `scheduled` to me would seem weird. 

I'm wondering if calling it abc or always-be-scheduling would be clearer than just scheduled. Mostly, I have minor UX concerns. The code will work, that looks pretty clear.

rick:
```
Using `any_instance` from rspec-mocks' old `:should` syntax without explicitly enabling the syntax is deprecated. Use the new `:expect` syntax or explicitly enable `:should` instead. Called from /home/stahnma/vanagon/spec/lib/vanagon/component/source/http_spec.rb:42:in `block (3 levels) in <top (required)>'.
```

Pushed a commit to update this to the new style.  The tests are generally kind of noisy.  I think that someone at some point tried to silence them by doing this:

https://github.com/puppetlabs/vanagon/blob/ba1402bf7e46a4262f79a9dba38c4e3951533c0f/spec/spec_helper.rb#L1-L6
``` ruby
RSpec.configure do |c|
  c.before do
    allow($stdout).to receive(:puts)
    allow($stderr).to receive(:puts)
  end
end
```

But it doesn't appear to be having the desired effect, as a few specific `puts` called from various `sources/` classes are getting through (among other things).

I don't see anything in `spec/` actually doing a `require` on `spec_helper`, so not sure if that was supposed to be automagically happening (yay, rspec) and changed, or if it never really worked.  I think one of the reasons I see it is that I'm just running `bundle exec rspec`, which just outputs dots, instead of the formatted version used in travis, which is verbose and hides the stray output.

lib/vanagon/driver.rb
```diff
@@ -46,6 +48,24 @@ def load_engine(engine_type, platform, target)
...
+      platform.engine && schedule_hardware_allocation?
+    end
+
+    # Should we allow the allocation of hosts via an external scheduler?
+    #
+    # REMIND: This is a feature flag method, triggered by an ENV setting
+    def schedule_hardware_allocation?
+      ENV['VANAGON_SCHEDULED_HARDWARE'] &&
+        !ENV['VANAGON_SCHEDULED_HARDWARE'].empty?
+    end
```
joshcooper: The desired engine can be passed to the `build_host_info` command:

```bash
$ bundle exec build_host_info puppet-agent el-7-x86_64 --engine pooler
{"name":"centos-7-x86_64","engine":"pooler"}
```

So you could use that instead of an ENV variable:

```shell
$ bundle exec build_host_info puppet-agent aix-7.1-ppc --engine scheduled
```

rick:
I've figured out how to suppress our noisy output (and I believe someone was either relying on a `.rspec` file to do something like `--require spec_helper.rb`, or that was a past rspec behavior that went away.

Anyway, I can push a commit that does the right thing there in a bit, but I notice now that this spec is failing sometimes (i.e., when `--order random` is in effect), which means there must be something ordering dependent about it:


https://github.com/puppetlabs/vanagon/blob/master/spec/lib/vanagon/engine/pooler_spec.rb#L33-L39
``` ruby
    it 'falls back to a file if the env var is not set' do
      expect(File).to receive(:expand_path).with('~/.vanagon-token').and_return(token_filename)
      expect(File).to receive(:exist?).with(token_filename).and_return(true)
      expect(File).to receive(:open).with(token_filename).and_return(token_file)
      expect(token_file).to receive(:read).and_return('abcd')
      expect(Vanagon::Engine::Pooler.new(platform).token).to eq('abcd')
    end
```

This sort of highly mockist white-box testing (where the chain of calls, args, and return values are all lined up and we trigger the dominoes in the last line) are pretty suspect to me: they have low value, they're brittle, and putting expectations on classes and modules (`File`) can be dicey.

I'll look at it further, though.

spec/lib/vanagon/driver_spec.rb
```diff
@@ -54,6 +54,40 @@ def create_driver(platform, options = {})
...
+      platform = eval_platform('aix-7.1-ppc', <<-END)
+        platform 'aix-7.1-ppc' do |plat|
+          plat.build_host ["pe-aix-71-01", "pe-aix-71-02"]
+          plat.engine "scheduled"
+        end
+      END
+
+      info = create_driver(platform).build_host_info
+
+      expect(info).to match({ 'name'   => 'aix-7.1-ppc',
```
joshcooper: Earlier we were talking about being consistent in our naming scheme for ABS, and using vmpooler as the gold standard for better or worse. It doesn't use dotted version numbers, e.g. `osx-1011-x86_64`, so I'm thinking this should be `aix-71-ppc`, which I think is worse, but would at least be consistently worse.

lib/vanagon/engine/scheduled.rb
```diff
@@ -0,0 +1,27 @@
+require 'vanagon/engine/base'
+require 'json'
+
+class Vanagon
+  class Engine
+    # Class to use when building on a shared resource device that is managed
+    # by an external scheduler.
+    class Scheduled < Base
```
joshcooper: What about using the `Scheduled` engine for both `pooler` and `hardware`? The [`build_host_name` method below](https://github.com/puppetlabs/vanagon/pull/417/files#diff-b48dd1920d5ccbd8b89b388eda6f547eR22) could return the `template_name` for pooler platforms and [something else](https://github.com/puppetlabs/vanagon/pull/417/files#r85019195) for hardware platforms.

If we only use `Scheduled` for hardware platforms, then `engine` can either be `pooler` (which is really the default/preferred allocator) or `scheduled` (which says to use the scheduler, but omits the allocator). IOW, engine means different things in different contexts, which gets confusing.

spec/lib/vanagon/driver_spec.rb
rick: Hm.  This is a good point. I was working from the platform names being used for hardware in vanagon, which are these dot-having names.

This *is* the place where we're expecting to emit abs-ready names, so I think it's useful to do the translation here while it's happening (as no other consumers exist for this).

lib/vanagon/driver.rb
rick: Gotcha!  Good point.

lib/vanagon/engine/scheduled.rb
rick: I'd be generally :+1: on using `Scheduled` for both `pooler` and `hardware`.  Lemme :zzz: on it and see what the ramifications are. I think it's probably pretty clean, though.

stahnma: Following along, things look good unless this change is still pending.

rick: Going to dig into this a bit. There's a checklist at the top of the PR, let's not merge until that's complete.

lib/vanagon/driver.rb
rick: Looking a bit closer, the cmdline options parsing for `build_host_info` passes in the value of the `--engine` flag, which ends up in our [`load_engine`](https://github.com/puppetlabs/vanagon/pull/417/files#diff-6d5e9c9d5b499a9e8d56ef00434459d9R36) method. Looks like that method just overwrites the `engine` value (before this PR).

Seems like we'd want to actually use that value if it's set, and **otherwise** go through the logic of setting a reasonable value.

spec/lib/vanagon/driver_spec.rb
stahnma: Yuck. :frowning_man: I'd really like to not make the naming of systems worse. I know we agreed not to change it, but I was hoping to contain the terrible part of it. Are there other options here or am I just doomed to be sad about our platform tagging and naming?

rick: Lemme see if I've got my head around everything:

We're probably all doomed to be sad about platform tagging and naming.  😿 

We did agree to use vmpooler names for interoperating with the ci.next scheduler (always-be-scheduling).  That means that we can either translate platform names in vanagon when they are outbound to always-be-scheduling or we can translate on the inbound side (note that the scheduler is sometimes a step or few removed from the output of `build_host_info` when we are running in a pipeline).

The work here started as "let's get `build_host_info` to not allocate and lock/unlock hardware" which implies "let's get vanagon on a path to not allocate / lock / unlock hardware". 

The logic would be to trigger `build_host_info` in such a way that we know ABS is asking for what platform a build is going to use, and if it's hardware to provide a hardware platform name and a count (rather than allocating and providing a hostname, as is the current flow).  The scheduler does the needful, and comes back with a specific hostname and provides that to vanagon when it's time to actually do the build.

We've so far never output a platform or template name for hardware from `build_host_info`, which means that whatever we do here, we're going to be changing behavior (under the scope of a specific engine being created) and will provide some version of a platform name.  I think @joshcooper's point (that I sympathize with) is that, if we're suddenly providing a platform name, and we know the recipient of that platform name wants it as a vmpooler-style platform name, we should probably actually output it in that format.

We are considering going a step further here and expanding the 'scheduler' engine (probably will be 'always-be-scheduling' to make things clearer) to also return template names for vmpooler builds, which, presumably would also be output in vmpooler format (as that's what always-be-scheduling would prefer to consume).  I think that's mostly a passthrough.

lib/vanagon/driver.rb
rick: I think I have a reasonable implementation for this in place, maybe.

lib/vanagon/engine/always_be_scheduling.rb
```diff
@@ -37,7 +37,21 @@ def name
 
       # return the platform name as the "host" name
       def build_host_name
-        @platform.name
+        platform_name_to_vmpooler_style(@platform.name)
+      end
+
+      def platform_name_to_vmpooler_style(platform)
+        translation = {
+          "aix-5.3-ppc" => "aix-53-ppc",
```
stahnma: can this be done generically? When AIX 7.2 or 8 comes out, it'd be pretty cool to not have to remember this is hard-coded in vanagon somewhere.

joshcooper: Or add another field to the platform definition (essentially what `vmpooler_template` is doing for pooler-based platforms)?

```ruby
platform "..." do |plat|
  ...
  plat.abs_resource_name "aix-53-ppc"
end
```

spec/lib/vanagon/driver_spec.rb
rick: I think this needs more discussion.  I have a candidate implementation for this up at @320891bd2b0.  Some thoughts (some mentioned in the commit message):

This is an explicit translation of known hardware platforms in use here (Puppet). An explicit translation list has a number of disadvantages:

 - When we add platforms we have to update it
 - It is necessarily focused on our conventions and installations
 - Which means it makes life hard for other vanagon users
 - It doesn't vary with time or versions of related software (so we can have difficulties with drift, or with trying to move from one implementation to another)

It would be possible to make the translation more programmatic (like via a regex), but that has most of the same pitfalls, even if we don't have to edit the file as often.

We've not been using hardware resources in ABS as of yet. We could translate on the inbound side if we wanted to (though for AIX, the vmpooler "style" is never going to actually touch vmpooler, so it's completely a "style").  We could not translate at all and just pass the platform through, even though it's not in the vmpooler "style".

We could bite the bullet (though I never want to make it be a dependency of much of anything) and finally implement a library that understands how to translate between platform standards.

Anyway, I think more discussion around this is needed. It's easy to revert a commit, so no worries there.

lib/vanagon/engine/always_be_scheduling.rb
nwolfe: We have all sorts of mappings like this in beaker-hostgenerator too.  At some point there's gotta be a shared library these tools can draw from, right?  Not sure.

rick: I'm in favor of having a shared library at some point. I'm not opposed to @joshcooper's idea of `abs_resource_name` (or similar) akin to `vmpooler_template`.  [usages of `vmpooler_template`](https://github.com/search?p=2&q=org%3Apuppetlabs+vmpooler_template&type=Code)

lib/vanagon/engine/always_be_scheduling.rb
```diff
@@ -0,0 +1,49 @@
...
+  class Engine
+    # This engine allows build resources to be managed by the "Always be
+    # Scheduling" (ABS) scheduler (https://github.com/puppetlabs/always-be-scheduling)
+    #
+    # ABS expects to ask `build_host_info` for the needed resources for a build,
+    # and to have that return a platform name.  ABS will then acquire the
+    # desired build host resources and will later run a vanagon build, passing
+    # those resource hostnames in specifically.
+    #
+    # `build_host_info` will normally use the hardware` engine when a hardware
```
nwolfe: s/hardware`/hardware

nwolfe: or s/hardware\`/\`hardware\`

rick:
:OK: This has been updated and could use another round of :eyes:

There are basically two final commits of information here:

 - A bunch of test-related changes: silencing extraneous test output, getting rid of some rspec deprecations, making some tests less order dependent / brittle
 - Work directly related to the `always_be_scheduling` engine and its test

After the review (and whatever changes might come from it) I'll update to bring this down to two commits and ask for one more :+1: pass before final merge.

spec/spec_helper.rb
```diff
@@ -4,3 +4,8 @@
     allow($stderr).to receive(:puts)
   end
 end
+
+module Kernel
+  def warn(*args)
+  end
```
joshcooper: can we stub something so we don't have to do this?

rick:
cc @stahnma @mckern looking for a review, then will split to 2 commits for merging.

rick:
Squashed to two commits. If CI is still happy, this is ready to merge.  :rocket:

PR #418, opened 2016-11-08
DO NOT MERGE (VANAGON-30) Add ability to specify package version numbers in build_…

ScottGarman:
…requires

This allows you to specify version numbers by appending "=<version>" to
the package name. This currently only works for (and is needed for) Linux
platforms with package managers which by default grab the latest version
of a package from an online repository.

I've verified this feature works by specifying older versions of pl-boost when building the target pl-yaml-cpp for the following platforms:

* (apt) debian-8-amd64
* (yum) el-7-x86_64
* (zypper) sles-12-x86_64
* (dnf) fedora-f22-x86_64

Additionally, I ran a dynamic vanagon builder job with a p-a modified to use this vanagon branch and built pl-yaml-cpp for a variety of platforms:

http://jenkins-release.delivery.puppetlabs.net/job/vanagon_generic_job/664/

### DIFF
--- a/lib/vanagon/driver.rb
+++ b/lib/vanagon/driver.rb
@@ -77,9 +77,43 @@ class Vanagon
       { "name" => @engine.build_host_name, "engine" => @engine.name }
     end
 
-    # Returns the set difference between the build_requires and the components to get a list of external dependencies that need to be installed.
+    # Normalize our pkgname=version syntax to work with the package
+    # manager being used for this platform.
+    def normalize_build_dependencies_syntax(build_deps)
+      if @platform.build_dependencies.command =~ /apt/
+        # apt uses '=' as the package/version separator, no need to do anything
+      elsif @platform.build_dependencies.command =~ /yum|dnf|zypper/
+        # yum/dnf/zypper uses '-' as the package/version separator
+        build_deps = build_deps.map { |pkg| pkg.sub('=', '-') }
+      else
+        raise Vanagon::Error, "Support for specifying package versions in build_requires is not available for #{@platform.name}"
+      end
+
+      build_deps
+    end
+
+    # Returns a list of external dependencies that need to be installed.
+    # If version numbers are specified, this method ensures the package
+    # names use the correct format for the platform's package manager.
     def list_build_dependencies
-      @project.components.map(&:build_requires).flatten.uniq - @project.components.map(&:name)
+      build_requires = @project.components.map(&:build_requires).flatten.uniq
+      component_names = @project.components.map(&:name)
+      build_deps = []
+      version_specified = false
+
+      build_requires.each do |pkg|
+        version_specified = true if pkg.include?('=')
+        # Skip any package names that match the name of a component
+        unless component_names.include?(pkg.split('=')[0])
+          build_deps << pkg
+        end
+      end
+
+      if version_specified
+        build_deps = normalize_build_dependencies_syntax(build_deps)
+      end
+
+      build_deps
     end
 
     def install_build_dependencies # rubocop:disable Metrics/AbcSize
--- a/lib/vanagon/component/dsl.rb
+++ b/lib/vanagon/component/dsl.rb
@@ -113,7 +113,8 @@ class Vanagon
       # build_requires adds a requirements to the list of build time dependencies
       # that will need to be fetched from an external source before this component
       # can be built. build_requires can also be satisfied by other components in
-      # the same project.
+      # the same project. An optional version number of a desired external package
+      # can be specified by appending '=<version>' to the package name.
       #
       # @param build_requirement [String] a library or other component that is required to build the current component
       def build_requires(build_requirement)


McdonaldSeanp:
👍

lib/vanagon/driver.rb
```diff
@@ -77,9 +77,43 @@ def build_host_info
...
+    # If version numbers are specified, this method ensures the package
+    # names use the correct format for the platform's package manager.
     def list_build_dependencies
-      @project.components.map(&:build_requires).flatten.uniq - @project.components.map(&:name)
+      build_requires = @project.components.map(&:build_requires).flatten.uniq
+      component_names = @project.components.map(&:name)
+      build_deps = []
+      version_specified = false
+
+      build_requires.each do |pkg|
```
McdonaldSeanp: nitpick: I'd rename "pkg" to something like build_requirement. "pkg" is a moniker we use regularly in the use of vanagon and I'd like to keep it out of the source code.

ScottGarman: I dunno, pkg is a nice short variable name appropriate for use within a block like that, and it literally represents the name (or name=version) of a specific package in a package repository. So naming it pkg seems more relevant to me than build_requirement.

McdonaldSeanp: That's fair

McdonaldSeanp:
👍

lib/vanagon/driver.rb
stahnma: Missing tests.

lib/vanagon/driver.rb
```diff
@@ -77,9 +77,43 @@ def build_host_info
...
-    # Returns the set difference between the build_requires and the components to get a list of external dependencies that need to be installed.
+    # Normalize our pkgname=version syntax to work with the package
+    # manager being used for this platform.
+    def normalize_build_dependencies_syntax(build_deps)
+      if @platform.build_dependencies.command =~ /apt/
+        # apt uses '=' as the package/version separator, no need to do anything
+      elsif @platform.build_dependencies.command =~ /yum|dnf|zypper/
+        # yum/dnf/zypper uses '-' as the package/version separator
+        build_deps = build_deps.map { |pkg| pkg.sub('=', '-') }
+      else
```
melissa: I would love to see nuget support in here as well

melissa: This may not be feasible due to how chocolaty passes in the version of a package to install. I can't remember off the top of my head though

melissa:
I agree with @stahnma on tests. I'm adding the 'don't merge' label until the above questions are addressed.

lib/vanagon/driver.rb
```diff
@@ -77,9 +77,43 @@ def build_host_info
...
+      component_names = @project.components.map(&:name)
+      build_deps = []
+      version_specified = false
+
+      build_requires.each do |pkg|
+        version_specified = true if pkg.include?('=')
+        # Skip any package names that match the name of a component
+        unless component_names.include?(pkg.split('=')[0])
+          build_deps << pkg
+        end
```
mckern: Realistically, would you get more milage out of extending the DSL to take a second parameter instead of splitting a string? I want to see some examples of the UX around this proposal, and I'd like to work through that side of it before we dig into the implementation.

ScottGarman: My original thought when approaching this was to add a second parameter to build_requires, but upon investigation this would have increased the complexity of the refactoring much more than the approach I ended up taking. Splitting a string turned out to be far simpler, and yet was also in line with the UX of specifying package versions with Linux based package managers, which either use an '=' or '-' to separate the package name and version. Given that the '-' was likely to be more ambiguous to parse, I settled on '='. What are the UX advantages we're missing out on by adopting this approach?

ScottGarman: Also, with regard to making this a 'y' release - is that actually necessary? The change is transparent to anyone not trying to make use of the additional feature, and is in that sense fully backward-compatible, isn't it?

stahnma: Why do we care if it's a Y release?

mckern: @stahnma there's some offline conversation we had around this yesterday but my position boils down to

- because it's right for users
- this is a change that cannot be expected to work in both directions (downgrading & upgrading Vanagon around an existing project)
  - specifically, it is a change in previous behavior that could potentially have unexpected side effects or cause unintentional behavior if pushed in a Z release

shrug:
Ping @ScottGarman - does this need updates?

ScottGarman:
I'm leaving this up for reference until I have an alternate pull request. In fact, I'm finding that the Linux package manager don't even support package ranges, so before I go any further I may start an email thread to discuss this more and include those who have commented on this already.

mckern:
Closing this since some technical shortcomings were found (specifically, package managers that don't understand how to install versioned packages at all).

PR #419, opened 2016-11-28
(maint) Fix bill-of-materials generation

haus:
The makefile dependency for the bill-of-materials was being incorrectly specified, which meant that any project using it would have broken dependencies for `@name-project`. This commit ensures the bill_of_materials target matches the dependency listed in the `@name-project` target.


haus:
Not sure if this will actually fix the issue @highb was seeing.

shrug:
👍

PR #420, opened 2016-12-06
Add note that build machines should be clean

traylenator:
When building multiple subsequent builds on a build
target things can go wrong. In particular the list
of files taken before and after the installation
phase can be quite easily incorrect.

Realistically the only way to be sure to be clean is
to reinstall the build machine.


shrug:
Thanks!

PR #421, opened 2016-12-16
(maint) use .vmfloaty.yml if .vanagon-token absent

johnmccabe:
This PR will look for the vmpooler auth token in `~/.vmfloaty.yml` if `~/.vanagon-token` is absent.

### DIFF
--- a/lib/vanagon/engine/pooler.rb
+++ b/lib/vanagon/engine/pooler.rb
@@ -1,4 +1,5 @@
 require 'vanagon/engine/base'
+require 'yaml'
 
 class Vanagon
   class Engine
@@ -36,12 +37,17 @@ class Vanagon
       end
 
       def token_from_file
-        token_file = File.expand_path("~/.vanagon-token")
+        vanagon_token_file = File.expand_path("~/.vanagon-token")
+        vmfloaty_token_file = File.expand_path("~/.vmfloaty.yml")
 
-        if File.exist?(token_file)
-          return File.open(token_file).read.chomp
+        if File.exist?(vanagon_token_file)
+          puts "Reading vmpooler token from: #{vanagon_token_file}"
+          return File.open(vanagon_token_file).read.chomp
+        elsif File.exist?(vmfloaty_token_file)
+          puts "Reading vmpooler token from: #{vmfloaty_token_file}"
+          conf = YAML.load_file(vmfloaty_token_file)
+          return conf['token']
         end
-
         nil
       end
 


mckern:
Broadly 👍 on this, but can you please add some documentation to the README for this feature?

lib/vanagon/engine/pooler.rb
```diff
@@ -36,12 +37,17 @@ def load_token
       end
 
       def token_from_file
-        token_file = File.expand_path("~/.vanagon-token")
+        vanagon_token_file = File.expand_path("~/.vanagon-token")
+        vmfloaty_token_file = File.expand_path("~/.vmfloaty.yml")
```
mckern: Conversation, not criticism: should we consider updating these to use `Dir.home` instead of `~`?

```ruby
vanagon_token_file = File.join(Dir.home, (".vanagon-token"))
vmfloaty_token_file = File.join(Dir.home, (".vmfloaty.yml"))
```

Thinking about it, `~` is a UNIX convention that most stuff *probably* understands, but `Dir.home` is pretty explicit. Both respect the value of `ENV['HOME']` but should we stop making UNIX-first assumptions about our tools?

lib/vanagon/engine/pooler.rb
```diff
@@ -36,12 +37,17 @@ def load_token
...
 
-        if File.exist?(token_file)
-          return File.open(token_file).read.chomp
+        if File.exist?(vanagon_token_file)
+          puts "Reading vmpooler token from: #{vanagon_token_file}"
+          return File.open(vanagon_token_file).read.chomp
```
mckern: I know this code predates this update, but this could probably be simplified with 

```ruby
File.read(vanagon_token_file).chomp
```

Also, should we consider using `#strip` instead of `#chomp` here? I'm not sure if we care about leading whitespace here. Again, just sort of thinking out loud about old code -- don't block on these questions.

lib/vanagon/engine/pooler.rb
```diff
@@ -36,12 +37,17 @@ def load_token
...
+        elsif File.exist?(vmfloaty_token_file)
+          puts "Reading vmpooler token from: #{vmfloaty_token_file}"
+          conf = YAML.load_file(vmfloaty_token_file)
+          return conf['token']
         end
```
mckern: Short term, this is probably good. But we should refactor this method a little bit and split each "load a kind of data from a specific source" chunk of code into their own methods.

johnmccabe:
@mckern I didn't get round to addressing your comments before going on hols, I'll raise a new PR with updates.

mckern:
(check later PRs)

johnmccabe:
Aha, cool, thx

PR #422, opened 2016-12-20
(CPR-393) puppetlabs-release-pc1 RPM on centos5

mwaggett:
This commit adds a dependency on `mkdir` and `touch`, as they are used in the
pre- and post-install script boilerplates, which then get used by every project.
In the future, we will supress pre/post scripts completely if there's nothing
specified by the project or the components.

### DIFF
--- a/resources/rpm/project.spec.erb
+++ b/resources/rpm/project.spec.erb
@@ -48,6 +48,15 @@ Autoreq: 0
 <%- get_requires.each do |requires| -%>
 Requires:  <%= requires %>
 <%- end -%>
+# All rpm packages built by vanagon have the pre-/post-install script
+# boilerplates (defined below). These require `mkdir` and `touch` but previously
+# did not specify a dependency on these.
+# In the future, we will supress pre/post scripts completely if there's nothing
+# specified by the project or the components.
+Requires(pre): /bin/mkdir
+Requires(pre): /bin/touch
+Requires(post): /bin/mkdir
+Requires(post): /bin/touch
 <%- if has_services? -%>
   <%- if @platform.servicetype == "systemd" -%>
     <%- if @platform.is_sles? -%>


shrug:
Do we use these in %post? I'm not seeing them there, unless we're pulling in things from the projects.

mwaggett:
Good point.. 
I was mostly just following @mckern's instructions and the notes we took yesterday talk about post-install, rather than pre-install, but I admit that my understanding it still a bit lacking.

resources/rpm/project.spec.erb
```diff
@@ -48,6 +48,15 @@ Autoreq: 0
...
 <%- end -%>
+# All rpm packages built by vanagon have the pre-/post-install script
+# boilerplates (defined below). These require `mkdir` and `touch` but previously
+# did not specify a dependency on these.
+# In the future, we will supress pre/post scripts completely if there's nothing
+# specified by the project or the components.
+Requires(pre): /bin/mkdir
+Requires(pre): /bin/touch
+Requires(post): /bin/mkdir
+Requires(post): /bin/touch
```
melissa: It looks like this isn't isolated to el5. Will mkdir and touch be in the same path for all the platforms we make this package available on?

mckern: Yes.

mckern:
@shrug We don't use it in the post, but my thoughts here were that this communicates where we expect those utilities to be used. And generally, sticking something in `%pre` that isn't also in `%post` seems like an oversight. We could probably limit it to the `%pre` section but I think it's academic -- won't really matter as long as dependencies are resolved.

resources/rpm/project.spec.erb
melissa: coolio

melissa:
I'm not merging so that @shrug has a chance to respond

PR #423, opened 2016-12-28
(maint) Add utility method uses_host_build_tools?

ScottGarman:
This method is used to detect platforms that use the host OS build
tools (instead of pl-build-tools packages).


mckern:
Not sure `host` is the word to use here. `vendor`, `distro`, or `upstream` might be better. Alternatively, should we instead express that a platform does not use the vendor dev toolchain at all with a `:skip_vendor_toolchain` style configuration option?

ScottGarman:
I do like `vendor` or `distro` better. Distro seems to be more of a Linux oriented term, so I can change this to vendor. Not sure if I follow what the benefits of using `skip_vendor_toolchain` would offer - perhaps I'm not totally clear on what you're implying?

mckern:
Top-down (this platform should not ever use a vendor's toolchain) vs bottom-up (this platform should only use our toolchain). @branan thoughts?

branan:
I agree with the use of `vendor`, for sure. As for which side should be the "default" ,I think that saying vendor toolchain is the default will make Vanagon more usable for those outside our walls.

mckern:
This is pedantic, but `build_tools` or `toolchain`? `toolchain` is the general nomenclature around your compiler & linker (and associated tools).

ScottGarman:
And since this includes the libboost and libyaml-cpp libraries from the vendor OS, I thought `toolchain` would be pedantically inappropriate.

mckern:
Yeah... we should probably do something about *that* too.

ScottGarman:
@MosesMendoza made a good point that the method/variable name could easily be misinterpreted as "this is an option to vendor the build tools", and suggested using "os" instead. We also agreed that "toolchain" is likely more appropriate here, with a proposed variable and  method name of "use_os_toolchain" / "platform.uses_os_toolchain?" @mckern would you be okay with that?

mckern:
`os` or `upstream`? I assume it'd look something like this in practice:

```ruby
# config/platforms/debian-8-armhf.rb
platform "debian-8-armhf" do |plat|
  plat.servicedir "/lib/systemd/system"
  plat.defaultdir "/etc/default"
  plat.servicetype "systemd"
  plat.codename "jessie"

  # I assume this is going to take a pure boolean value?
  plat.use_os_toolchain false

  plat.add_build_repository "http://pl-build-tools.delivery.puppetlabs.net/debian/pl-build-tools-release-jessie.deb"
  plat.provision_with "export DEBIAN_FRONTEND=noninteractive; apt-get update ; apt-get install -qq -y --no-install-recommends build-essential devscripts make quilt pkg-config debhelper rsync fakeroot binutils-arm-linux-gnueabihf"
  plat.install_build_dependencies_with "DEBIAN_FRONTEND=noninteractive; apt-get install -qq -qy --no-install-recommends "
  plat.vmpooler_template "debian-8-x86_64"
  plat.cross_compiled "true"
  plat.output_dir File.join("deb", plat.get_codename, "PC1")
end
```

```ruby
# config/components/curl.rb
# I assume that this will always return a pure boolean value?
unless platform.uses_os_toolchain?
  pkg.build_requires 'runtime'
  pkg.environment "PATH" => "/opt/pl-build-tools/bin:$$PATH:#{settings[:bindir]}"
end
```

lib/vanagon/platform.rb
```diff
@@ -111,6 +111,7 @@ def initialize(name) # rubocop:disable Metrics/AbcSize
       @sort ||= "sort"
       @copy ||= "cp"
       @cross_compiled ||= false
+      @vendor_build_tools ||= false
```
mckern: Regardless of what we call this, shouldn't this default to `true`? The idea is that "most platforms probably use native vendor-provided upstream toolchains", right?

lib/vanagon/platform.rb
```diff
@@ -271,5 +272,13 @@ def is_cross_compiled?
     def is_cross_compiled_linux?
       return (is_cross_compiled? && is_linux?)
     end
+
+    # Utility matcher to determine if the platform uses host OS provided build tools
+    # (toolchain, boost, etc).
+    #
+    # @return [true, false] true if the platform uses host OS build tools, false otherwise
+    def uses_vendor_build_tools?
+      return @vendor_build_tools
```
mckern: Ditch the explicit `return`, and consider always coercing this value into a pure boolean. Many DSL parameters are passed as strings, and the existing DSL method does no type checking for what is passed into it.

lib/vanagon/platform/dsl.rb
```diff
@@ -128,6 +128,13 @@ def cross_compiled(cxx_flag)
         @platform.cross_compiled = cxx_flag
       end
 
+      # Set the vendor_build_tools flag for the platform
+      #
+      # @param flag [Boolean] True if this platform uses host OS provided build tools
```
mckern: The documentation for this method seems off/incomplete -- am I reading it poorly or is it misleading?

ScottGarman:
@mckern I've agreed with your feedback and made changes accordingly. Mind taking a final look and offering a thumbs up if approved?

McdonaldSeanp:
@ScottGarman I would update the name of these methods to: `use_vendor_toolchain`

As we discussed, this should alleviate any onfusion about "vendor" as a verb, and still retain the clarity be specifying that this is the toolchain provided by the upstream vendor

mckern:
I'm calling a moratorium on name talk. We've had too much back and forth. We're gonna go ask someone in UX.

lib/vanagon/platform.rb
```diff
@@ -271,5 +272,15 @@ def is_cross_compiled?
...
     end
+
+    # Utility matcher to determine if the platform uses host OS provided build tools
+    # (toolchain, boost, etc).
+    #
+    # @return [true, false] whether platform uses host OS toolchain, determined by
+    # whether the platform config set the use_os_toolchain flag (defaults to true
+    # if not set)
+    def uses_os_toolchain?
+      !!@use_os_toolchain
```
mckern: This isn't going to work the way you hope it will due to the way that booleans work. Strings are always true unless you parse them first.

```ruby
class CrapulenceFirkin
  attr_accessor :boolean

  def initialize(boolean_value)
    @boolean = boolean_value
  end

  def pure_boolean
    puts (boolean).inspect
    puts (!boolean).inspect
    puts (!!boolean).inspect
    !!boolean
  end
end
```

Behold!

```ruby
class CrapulenceFirkin
  attr_accessor :boolean

  def initialize(boolean_value)
    @boolean = boolean_value
  end

  def pure_boolean
    puts (boolean).inspect
    puts (!boolean).inspect
    puts (!!boolean).inspect
    !!boolean
  end
end
```

Here's that toy class in use:

```ruby
>> truthiness = CrapulenceFirkin.new("true")
=> #<CrapulenceFirkin:0x007fcc62e54378 @boolean="true">
>> truthiness.boolean
=> "true"
>> truthiness.pure_boolean
"true" # <= original value is a string (true)
false  # <= first inversion is a pure boolean
true   # <= second inversion is a pure boolean corresponding to
       #    the original string being considered true.
=> true
>> falsiness = CrapulenceFirkin.new("false")
=> #<CrapulenceFirkin:0x007fcc6040fd38 @boolean="false">
>> falsiness.boolean
=> "false"
>> falsiness.pure_boolean
"false" # <= original value is a string
false   # <= first inversion is a pure boolean
true    # <= second inversion is a pure boolean corresponding to
        #    the original string being considered true...
        #    but the desired value is false, not true.
=> true
>>
```

If you parse/match the value, you can get what you're looking for. Here's an example I added to that toy class above:

```ruby
  def truthiness
    return false if %w(false f no n).include? @boolean
    !!@boolean
  end
```

And in use:

```ruby
>> falsiness = CrapulenceFirkin.new("false")
=> #<CrapulenceFirkin:0x007f8e5a969c50 @boolean="false">
>> falsiness.truthiness
=> false
>> falsiness.boolean
=> "false"
```

lib/vanagon/platform/dsl.rb
```diff
@@ -128,6 +128,13 @@ def cross_compiled(cxx_flag)
         @platform.cross_compiled = cxx_flag
       end
 
+      # Set the use_os_toolchain flag for the platform
+      #
+      # @param flag [Boolean] True if this platform uses host OS provided toolchain
+      def use_os_toolchain(flag)
+        @platform.use_os_toolchain = flag
```
mckern: Re: parsing values. You can do it here, or you can do it when it's called by `platform.rb` but I'd only mangle/parse/match it in one place. You choice.

lib/vanagon/platform.rb
mckern: (I would probably not use my naive example code  for `#truthiness` verbatim BTW)

lib/vanagon/platform.rb
```diff
@@ -271,5 +272,19 @@ def is_cross_compiled?
...
+    # Utility matcher to determine if the platform uses host OS provided build tools
+    # (toolchain, boost, etc).
+    #
+    # @return [true, false] whether platform uses host OS toolchain, determined by
+    # whether the platform config set the use_os_toolchain flag (defaults to true
+    # if not set)
+    def uses_os_toolchain?
+      # Ensure we're returning a boolean:
+      return true if @use_os_toolchain == true || @use_os_toolchain == "true"
+      return false if @use_os_toolchain == false || @use_os_toolchain == "false"
```
branan: Should this also handle the `:true` and `:false` symbols? What about `1` and `0`? nil?

PR #427, opened 2016-12-31
(MAINT) Add Simplecov support to Rakefile

mckern:
[Simplecov](https://github.com/colszowka/simplecov) allows us to start building rough metrics for how much code in Vanagon is actually tested. It's not bulletproof, but any measurement is probably better than no measurement.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -7,7 +7,7 @@ desc 'Test Vanagon'
 namespace :test do
   require 'rspec/core/rake_task'
   RSpec::Core::RakeTask.new do |task|
-    task.rspec_opts = %w(--format documentation --color --require spec_helper)
+    task.rspec_opts = %(--format documentation --color --require spec_helper)
   end
 
   desc 'Test Vanagon and calculate test coverage'
--- a/Gemfile
+++ b/Gemfile
@@ -24,7 +24,8 @@ group(:development, :test) do
   gem 'json'
   gem 'lock_manager', *lock_manager_location_for(ENV['LOCK_MANAGER_LOCATION'] || '>= 0')
   gem 'packaging', '~> 0.4.0', github: 'puppetlabs/packaging', branch: 'master'
-  gem 'rake'
+  gem 'simplecov', require: false
+  gem 'rake', require: false
   gem 'rspec', '~> 3.0', require: false
   gem 'rubocop', "~> 0.41.2", require: false
   gem 'yard', require: false
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -1,3 +1,12 @@
+if ENV["COVERAGE"]
+  require 'simplecov'
+  SimpleCov.start do
+    add_filter '.bundle'
+    add_filter 'spec'
+    add_filter 'vendor'
+  end
+end
+
 RSpec.configure do |c|
   c.before do
     allow_any_instance_of(Vanagon::Component::Source::Git).to receive(:puts)


Rakefile
```diff
@@ -7,7 +7,7 @@ desc 'Test Vanagon'
 namespace :test do
   require 'rspec/core/rake_task'
   RSpec::Core::RakeTask.new do |task|
-    task.rspec_opts = %w(--format documentation --color --require spec_helper)
+    task.rspec_opts = %(--format documentation --color --require spec_helper)
```
shrug: mmm?

mckern: This is safe; it didn't need to be an Array.

shrug:
👍

glennsarti:
👍 Changes seem sane...is there any local test output you can share?

mckern:
@glennsarti since there's no functionality change, there's no tests to change. But here's the terminal output of the `test:coverage` task now:

```
mckern@flexo vanagon (git:maint/add-simplecov) $ bundle exec rake test:coverage
/opt/local/var/rbenv/versions/2.3.3/bin/ruby -I/Users/mckern/Documents/Repositories/vanagon/.bundle/gems/ruby/2.3.0/gems/rspec-core-3.5.4/lib:/Users/mckern/Documents/Repositories/vanagon/.bundle/gems/ruby/2.3.0/gems/rspec-support-3.5.0/lib /Users/mckern/Documents/Repositories/vanagon/.bundle/gems/ruby/2.3.0/gems/rspec-core-3.5.4/exe/rspec --pattern spec/\*\*\{,/\*/\*\*\}/\*_spec.rb --format documentation --color --require spec_helper
Unable to load AWS SDK; skipping optional EC2 engine spec tests

Makefile::Rule
  a rule with no dependencies and an empty recipe
    creates an empty rule
  a rule with no dependencies and a simple recipe
    creates the rule with the recipe
  a rule with dependencies and no recipe
    creates the rule with the recipe
  a rule with dependencies and a recipe
    creates the rule with the recipe
  a rule with a multiline recipe
    inserts tabs after each newline in the recipe

Vanagon::Common::Pathname
  #has_overrides?
    is false for a pathname with just a path
    is true if the pathname has more than a path set
  #file
    creates a new Pathname instance, marked as a file
  #configfile
    creates a new Pathname instance, marked as a configuration file
  #initialize
    strips trailing slashes off of the path to normalize it
    removes extra / from the pathname to normalize it
  equality
    is not equal if the paths differ
    is not equal if there are different attributes set
    is equal if there are the same attributes set to the same values
    is equal if the paths are the same and the only attribute set
  #hash
    has the same hash is the attributes are the same
    has different hashes if any attribute is different
  uniqueness of pathnames
    should only add 1 Pathname object with the same attributes to a set
    should reduce an array to unique elements successfully

Vanagon::Common::User
  initialize
    group defaults to the name of the user
  equality
    is not equal if the names differ
    is not equal if there are different attributes set
    is equal if there are the same attributes set to the same values
    is equal if the name are the same and the only attribute set

Vanagon::Component::DSL
  #load_from_json
    sets the ref and url based on the json fixture
    raises an error on invalid methods/attributes in the json
  #configure
    sets configure to the value if configure is empty
    appends to the existing configure if not empty
  #build
    sets build to the value if build is empty
    appends to the existing build if not empty
  #check
    sets check to the value if check is empty
    appends to the existing check if not empty
  #install
    sets install to the value if install is empty
    appends to the existing install if not empty
  #apply_patch
    adds the patch to the list of patches
    can specify strip and fuzz
    can specify a directory where the patch should be applied
    can specify when to try to apply the patch
    will default the patch timing to after the source is unpacked
    will fail if the user wants to install the patch at an unsupported step
    can specify a directory where the patch should be applied
    can specify when to try to apply the patch
    will default the patch timing to after the source is unpacked
    will fail if the user wants to install the patch at an unsupported step
  #build_requires
    adds the build requirement to the list of build requirements
  #requires
    adds the runtime requirement to the list of requirements
  #provides
    adds the package provide to the list of provides
    supports versioned provides
  #replaces
    adds the package replacement to the list of replacements
    supports versioned replaces
  #conflicts
    adds the package conflict to the list of conflicts
    supports versioned conflicts
  #add_actions
    adds the corect preinstall action to the component
    fails with bad preinstall action
    fails with empty preinstall action
    adds the corect postinstall action to the component
    fails with bad postinstall action
    fails with empty postinstall action
    adds the corect preremove action to the component
    fails with bad preremove action
    fails with empty preremove action
    adds the corect postremove action to the component
    fails with bad postremove action
    fails with empty postremove action
  #install_service
    adds the correct command to the install for the component for sysv platforms
    reads from a file when the OS is AIX for services
    adds the correct command to the install for the component for systemd platforms
    adds the correct command to the install for smf services using a service_type
    adds the correct command to the install for smf services
  #install_file
    installs correctly using GNU install on AIX
    adds the correct commands to the install to copy the file
    adds an owner and group to the installation
  configfile handling
    on anything but solaris 10
      #configfile
        adds the file to the configfiles list
      #install_configfile
        adds the commands to install the configfile
        adds the file to the configfiles list
    on solaris 10, do something terrible
      #configfile
        adds the file to the configfiles list
      #install_configfile
        adds the commands to install the configfile
        adds the file to the configfiles list
  #build_dir
    sets the build_dir when given a relative path
    raises an error when given an absolute path
  #link
    adds the correct command to the install for the component
  #environment
    adds an override to the environment for a component
    merges against the existing environment
  #directory
    adds a directory with the desired path to the directory collection for the component
    adds a directory with the desired mode to the directory collection for the component
    adds a directory with the desired owner to the directory collection for the component
    adds a directory with the desired group to the directory collection for the component
    adds a directory with the desired attributes to the directory collection for the component

Vanagon::Component::Rules
  the component rule
    depends on the component-install rule
  the unpack rule
    should eq ["file-list-before-build"]
    extracts the source
    sets environment variables before running the unpack steps
    behaves like a rule that touches the target file
      touches the target file as the last step of the recipe
  the patch rule
    should eq ["leatherman-unpack"]
    does nothing when there are no patches
    applies each listed patch in order when patches are set
    behaves like a rule that touches the target file
      touches the target file as the last step of the recipe
  the configure rule
    should eq ["leatherman-patch"]
    runs all of the configure commands when given
    sets environment variables before running the configure steps
    when a build directory is set
      creates the build directory
    behaves like a rule that touches the target file
      touches the target file as the last step of the recipe
  the build rule
    should eq ["leatherman-configure"]
    does nothing when the build step is empty
    runs all of the build commands when given
    sets environment variables before running the build steps
    behaves like a rule that touches the target file
      touches the target file as the last step of the recipe
  the check rule
    should eq ["leatherman-build"]
    does nothing when the check step is empty
    does nothing when the project skipcheck flag is set
    runs all of the check commands when given
    sets environment variables before running the check steps
    behaves like a rule that touches the target file
      touches the target file as the last step of the recipe
  the install rule
    should eq ["leatherman-check"]
    does nothing when the install step is empty
    runs all of the install commands when given
    sets environment variables before running the install steps
    applies any after-install patches
    behaves like a rule that touches the target file
      touches the target file as the last step of the recipe
  the cleanup rule
    should eq ["leatherman-install"]
    runs the component source cleanup step
    behaves like a rule that touches the target file
      touches the target file as the last step of the recipe
  the clean rule
    should be empty
    runs a `make clean` in the build dir
    remotes the touch files for the configure, build, and install steps
  the clobber rule
    should eq ["leatherman-clean"]
    removes the source directory and unpack touchfile

Vanagon::Component::Source::Git
  #initialize
    raises error on initialization with an invalid repo
  #dirname
    returns the name of the repo
    returns the name of the repo and strips .git
  #ref
    returns a default value of HEAD when no explicit Git reference is provided
  #fetch
    raises an error on checkout failure with a bad SHA

Vanagon::Component::Source::Http
  #dirname
    returns the name of the tarball, minus extension for archives
    returns the current directory for non-archive files
  verify
    fails with a bad sum_type
    calls md5 digest when it's supposed to
    calls sha256 digest when it's supposed to
    calls sha512 digest when it's supposed to

Vanagon::Component::Source::File
  #fetch
    puts the source file in to the workdir
    puts the source directory in to the workdir
    preserves nested directories when copying folders
  #dirname
    returns the name of the tarball, minus extension for archives
  #ref
    returns the current directory for non-archive files
  #extension
    returns the extension for valid extensions

Vanagon::Component::Source
  .source
    fails on unrecognized URI schemes
    fails on invalid URIs
    takes a Git repo
      returns a Git object for git@ triplet repositories
      returns a Git object for git:// repositories
      returns a Git object for http:// repositories
      returns a Git object for https:// repositories
      rewrites git:// URLs
    takes a HTTP/HTTPS file
      returns an object of the correct type for http:// URLS
      returns an object of the correct type for https:// URLS
      applies rewrite rules to HTTP URLs
    takes a local file
      returns an object of the correct type for file:// URLS
  .rewrite
    replaces the first section of a url with a string if string is given
    applies the rule to the url if a proc is given as the rule
    applies the rule to the url if a proc is given as the rule
  .register_rewrite_rule
    only accepts Proc and String as rule types
    rejects invalid protocols
    registers the rule for the given protocol

Vanagon::Component
  #get_environment
    returns a makefile compatible environment
    merges against the existing environment
    returns : for an empty environment
  #get_build_dir
    uses the dirname when no build_dir was set
    joins the dirname and the build dir when a build_dir was set

Vanagon::Driver
  when resolving build host info
    returns the vmpooler_template using the pooler engine
    returns the vmpooler template with an explicit engine
    returns the first build_host using the hardware engine
    returns the first build_host using the hardware engine when not using an engine, and not specifying an engine in the platform configuration
    returns the platform and the always_be_scheduling engine when using the always_be_scheduling engine
    returns the vmpooler_template using the always_be_scheduling engine when using the always_be_scheduling engine
    returns the docker_image using the docker engine
    returns "local machine" using the local engine
    raises when using the base engine

Vanagon::Engine::AlwaysBeScheduling
  #validate_platform
    returns true if the platform has the required attributes
  #build_host_name
    by default returns the platform name with no translation
    returns vmpooler_template if vmpooler_template is specified
    returns abs_resource_name if abs_resource_name is specified
    prefers abs_resource_name to vmpooler_template if both are specified
  #name
    returns "always_be_scheduling" engine name

Vanagon::Engine::Base
  #select_target
    raises an error without a target
    returns a target if one is set
  #validate_platform
    raises an error if the platform is missing a required attribute
    returns true if the platform has the required attributes

Vanagon::Engine::Docker
  returns "docker" name
  #initialize
    fails without docker installed
  #validate_platform
    raises an error if the platform is missing a required attribute
    returns true if the platform has the required attributes

Vanagon::Engine::Hardware
  returns "hardware" name
  #select_target
    raises an error without a target
    returns a target if one is set
  #validate_platform
    raises an error if the platform is missing a required attribute
    returns true if the platform has the required attributes

Vanagon::Engine::Local
  returns "local" name
  #validate_platform
    succeeds
  #dispatch
    execs successfully
    returns the result if return_output is true

Vanagon::Engine::Pooler
  returns "pooler" name
  #load_token
    prefers an env var to a file
    falls back to a file if the env var is not set
    returns nil if there is no env var or file
  #validate_platform
    raises an error if the platform is missing a required attribute
    returns true if the platform has the required attributes

OpenStruct
  with JSON mixins
    responds to #to_json
    can be converted to a valid JSON object

Set
  with JSON mixins
    responds to #to_json
    can be converted to a valid JSON object

String
  responds to #undent

Vanagon::Extensions::String
  #undent
    trims trivial leading whitespace
    trims more complex whitespace
    trims leading tabs

Vanagon::OptParse
  options that don't take a value
    can create an option parser that accepts the skipcheck flag
    can create an option parser that accepts the preserve flag
    can create an option parser that accepts the verbose flag
    short options
      maps the short option p to preserve
      maps the short option v to verbose
  options that take a value
    can create an option parser that accepts the workdir option
    can create an option parser that accepts the configdir option
    can create an option parser that accepts the target option
    can create an option parser that accepts the engine option
    short options
      maps the short option w to workdir
      maps the short option c to configdir
      maps the short option t to target
      maps the short option e to engine

Vanagon::Platform::DSL
  #apt_repo
    grabs the file and adds .list to it
    installs a deb when given a deb
    installs a gpg key if given one
  #yum_repo
    grabs the file and adds .repo to it
    downloads the repo file to the correct yum location for wrlinux
    installs a rpm when given a rpm
      uses rpm everywhere
  #zypper_repo
    grabs the file and adds .repo to it
    installs a sles rpm when given a rpm
  #add_build_repository
    hands off to the platform specific method if defined
    raises an error if the platform does not define "add_repository"
  #abs_resource_name
    sets the instance variable on platform
  #vmpooler_template
    sets the instance variable on platform
    is called by vcloud_name as a deprecation

Vanagon::Platform::RPM::AIX
  #rpm_defines
    doesn't include dist on aix
  aix puts commands in weird places
    uses /opt/freeware/bin everwhere

Vanagon::Platform::RPM
  on el-5-i386 we should behave ourselves
    #rpm_defines
      removes dashes from the dist macro
  on fedora-21-x86_64 we should behave ourselves
    #rpm_defines
      removes dashes from the dist macro
  on cisco-wrlinux-7-x86_64 we should behave ourselves
    #rpm_defines
      removes dashes from the dist macro

Vanagon::Platform::Solaris_11
  #ips_version
    strips invalid characters from the version
    handles the release field correctly

Vanagon::Platform::Windows
  on windows-2012r2-x64
    #target_user
      sets the target_user to 'Administrator'
    #wix_product_version
      returns first three digits only
      returns only numbers
    #generate_msi_packaging_artifacts
      Copies Wix File from product specific directory to output directory
      Copies Wix File from Vanagon directory to work directory
      Picks Project Specific Wix File in favour of Generic Wix file
      Picks Project Specific Wix File in favour of Generic ERB file
      Picks Project Specific ERB File in favour of Generic Wix file
      Picks Project Specific ERB File in favour of Generic ERB file
      Copies Hierarchy of files from Product Specific Directory to output directory with ERB translation as necessary
      Copies Hierarchy of files from vanagon directory to output directory with ERB translation as necessary
      generate_wix_dirs
        returns one directory with install_service defaults
        returns one directory with non-default name
        returns nested directory correctly with \
        adds a second directory for the same input but different components
        returns correctly formatted multiple nested directories

Vanagon::Platform
  #os_name
    returns the os_name for the platform
  #os_version
    returns the the os_version for the platform
  #output_dir
    returns correct output dir
    adds the target repo in the right way
    does the right thing with empty strings
  #architecture
    returns the architecture for the platform

Vanagon::Project::DSL
  #version_from_git
    sets the version based on the git describe
    sets the version based on the git describe with multiple dashes
  #directory
    adds a directory to the list of directories
  #user
    sets a user for the project
  #target_repo
    sets the target_repo for the project
  #noarch
    sets noarch on the project to true
  #identifier
    sets the identifier for the project
  #cleanup_during_build
    sets @cleanup to true
    defaults to nil
  #write_version_file
    sets version_file for the project
  #release
    sets the release for the project
    defaults to 1
  #provides
    adds the package provide to the list of provides
    supports versioned provides
    gets rid of duplicates
  #replaces
    adds the package replacement to the list of replacements
    supports versioned replaces
    gets rid of duplicates
  #conflicts
    adds the package conflict to the list of conflicts
    supports versioned conflicts
    gets rid of duplicates
  #component
    stores the component in the project if the included components set is empty
    stores the component in the project if the component name is listed in the included components set
    does not store the component if the included components set is not empty and does not include the component name

Vanagon::Project
  #get_root_directories
    returns only the highest level directories
  #filter_component
    returns nil when given a component that doesn't exist
    returns only the component with no build deps
    returns component and one build dep
    returns only the component with build deps that are not components of the @project
    returns the component and build deps with both @project components and external build deps
    returns the component and multiple build deps
    returns the component and multiple build deps with external build deps

Vanagon::Utilities::ShellUtilities
  #andand joins commands with &&
  #andand_multiline joins commands with && and an escaped newline
  #cmdjoin
    returns a single value as-is
    turns an array with a single value into the wrapped value
    joins multiple commands with the separator string
    joins single strings and arrays of strings

Vanagon::Utilities
  #find_program_on_path
    finds commands on the PATH
    finds commands on the PATH
    raises an error if required is true and command is not found
    returns false if required is false and command is not found
  #local_command
    runs commands in an unpolluted environment
  #ssh_command
    adds the correct flags to the command if VANAGON_SSH_KEY is set
    returns just the path to ssh if VANAGON_SSH_KEY is not set
    sets the port to 22 when none is specified
    sets the port to 2222 when that is specified
    disables strict host checking
    sets known hosts file to /dev/null
    adds the correct flags to the command if VANAGON_SSH_AGENT is set
  #retry_with_timeout
    raises a Vanagon::Error if the command fails n times
    returns true if the command succeeds within n times
    raises a Vanagon::Error if the command times out

Finished in 5.59 seconds (files took 0.26834 seconds to load)
320 examples, 0 failures

mckern@flexo vanagon (git:maint/add-simplecov) $
```

glennsarti:
I'm 👍  for merge

mckern:
@glennsarti Should probably figure out how to consume Simplecov output programmatically. I'll work on that in a separate commit/PR.

glennsarti:
In that case...LGTM

PR #428, opened 2017-01-04
(MAINT) Fix tests that fail on macOS

mckern:
Thanks to macOS' [System Integrity Protection feature](https://en.wikipedia.org/wiki/System_Integrity_Protection), assuming that writing to `/tmp` will probably let you down. Instead, we should call `mktemp -d` or using Dir.mktmpdir from within Ruby. Any tests that hardcoded `/tmp` as a value have been updated.


shrug:
👍

PR #429, opened 2017-01-04
(MAINT) Travis-CI fixups for the new Rakefile

mckern:
Vanagon uses the same group for :development and :test, so there's no
benefit to running Bundler with the `--without-development` flag.

The stubbed-out Git config during `before_check` has been updated to be
slightly more informative

Travis will now run the `test:spec` task instead of calling rspec
directly.

Travis is now unpinned from specific Ruby versions, using the aliases
for the x.y releases we actually care about.

This commit also includes a quick fix for the Rakefile: making sure
that Rubocop displays the names of failing cops by default.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,21 +1,20 @@
 language: ruby
-bundler_args: --without development
 sudo: false
 before_script:
-  - 'git config --global user.email "you@example.com"'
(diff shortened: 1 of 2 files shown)


stahnma:
👍

PR #430, opened 2017-01-04
(MAINT) Fix undefined variable in Pooler engine

mckern:
When token loading was refactored, this variable was missed. That
caused anyone using HEAD some lamentations and wailing when they tried
to load pooler tokens. The spec test for this functionality needs to
be refactored, but this is a blocking issue for at least one CI
pipeline.


melissa:
will merge once travis finishes

PR #431, opened 2017-01-06
(MAINT) Improve spec coverage for token loading

mckern:
When token loading was refactored, the coverage was not sufficient.
Failures were silently masked due to overly aggressive stubbing out of
test resources. The coverage for the way that the pooler engine
handles environment variables, a ~/.vmpooler-token file, and a
~/.vmfloaty.yml file has been rewritten completely to exercise the
intended functionality.

### DIFF
--- a/spec/lib/vanagon/engine/pooler_spec.rb
+++ b/spec/lib/vanagon/engine/pooler_spec.rb
@@ -21,28 +21,88 @@ describe 'Vanagon::Engine::Pooler' do
   before :each do
     # suppress `#warn` output during tests
     allow_any_instance_of(Vanagon::Platform::DSL).to receive(:warn)
+
+    # stubbing ENV doesn't actually intercept ENV because ENV
+    # is special and you aren't. So we have to mangle the value
+    # of ENV around each of these tests if we want to prevent
+    # actual user credentials from being loaded, or attempt to
+    # validate the way that credentials are read.
+    ENV['REAL_HOME'] = ENV.delete 'HOME'
+    ENV['HOME'] = Dir.mktmpdir
+
+    # This should help maintain the ENV['HOME'] masquerade
+    allow(Dir).to receive(:home).and_return(ENV['HOME'])
   end
 
+  after :each do
+    # Altering ENV directly is a legitimate code-smell.
+    # We should at least clean up after ourselves.
+    ENV['HOME'] = ENV.delete 'REAL_HOME'
+  end
+
+  # We don't want to run the risk of reading legitimate user
+  # data from a user who is unfortunate enough to be running spec
+  # tests on their local machine. This means we have to intercept
+  # and stub out a very specific environment, where:
+  #   1) an environment variable exists
+  #   2) the env. var is unset, and only a ~/.vmpooler-token file exists
+  #   3) the env. var is unset, and only a ~/.vmfloaty.yml file exists
   describe "#load_token" do
-    after(:each) { ENV['VMPOOLER_TOKEN'] = nil }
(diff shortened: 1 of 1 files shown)


spec/lib/vanagon/engine/pooler_spec.rb
```diff
@@ -21,28 +21,86 @@
   before :each do
     # suppress `#warn` output during tests
     allow_any_instance_of(Vanagon::Platform::DSL).to receive(:warn)
+
+    # stubbing ENV doesn't actually intercept ENV because ENV
```
ericwilliamson: hmm..So we do this in one of our puppet modules spec tests...would it work for this use case?

```
allow(ENV).to receive(:[]).and_call_original
allow(ENV).to receive(:[]).with("PE_NODE_GROUP_CLASSIFICATION_ATTEMPTS").and_return("2")
```

https://github.com/puppetlabs/puppetlabs-pe_install/blob/ea403f29235fcd568a2f5f66d8422956d8c45f39/spec/unit/util/pe_node_groups.rb#L137-L138

mckern: Nope. That only works at what I'm affectionately referring to as the "Ruby" level. Anything that reads environment variables from what I'm frustratingly referring to as the "C" level will completely disregard that and return your real, true process environment. So when the code paths we're testing call `File.expand_path`, that's a pure C method and it returns a value from the C level.

```ruby
>> ENV.inspect
=> {"HOME"=>"/var/folders/jg/0fyykqj50wgc48f_6bygnblr0000gn/T/d20170105-47621-yontgt"}
>> puts File.join(ENV['HOME']), .vmfloaty.yml)
=> /var/folders/jg/0fyykqj50wgc48f_6bygnblr0000gn/T/d20170105-47621-yontgt/.vmfloaty.yml
>> puts File.expand_path('~/.vmfloaty.yml')
=> /Users/mckern/.vmfloaty.yml
>>
```

mckern: (Credit to @branan for helping me work that out)

spec/lib/vanagon/engine/pooler_spec.rb
```diff
@@ -21,28 +21,86 @@
...
+  #   3) the env. var is unset, and only a ~/.vmfloaty.yml file exists
   describe "#load_token" do
-    after(:each) { ENV['VMPOOLER_TOKEN'] = nil }
+    let(:environment_value) { 'cafebeef' }
+    let(:token_value) { 'decade' }
+    let(:pooler_token_file) { File.expand_path('~/.vanagon-token') }
+    let(:floaty_config) { File.expand_path('~/.vmfloaty.yml') }
+
+    it 'prefers the VMPOOLER_TOKEN environment variable to a config file' do
+      stub_const('ENV', { 'VMPOOLER_TOKEN' => environment_value } )
```
branan: Since we're already messing with ENV, I don't think we should necessarily stub_const here. This will create a "shadow" ENV object that will contain only this key/value, and not things like HOME that we set above.

mckern: Yeah, maybe. Since this test *only* focuses on the value of that env. var. though, does it matter? I'm open to suggestions that don't involve setting and unsetting any more env. values directly.

mckern: It does look like we could use this here:

```ruby
      allow(ENV).to receive(:[])
                      .with('VMPOOLER_TOKEN')
                      .and_return(environment_value)
```

PR #436, opened 2017-01-14
(VANAGON-44) Add --remote_dir flag

adreyer:
I'd like to be able to specify the remote_workdir on the build build box so that all dependencies and make files remain in place and I can rebuild specific components quickly. For example when working on puppet-job for windows this allows me to quickly provision a build box on which I can recompile in a few seconds.

Is this a feature anyone else wants?
Is there a better way to do this?

### DIFF
--- a/lib/vanagon/engine/pooler.rb
+++ b/lib/vanagon/engine/pooler.rb
@@ -7,7 +7,7 @@ class Vanagon
       attr_reader :token
 
       # The vmpooler_template is required to use the pooler engine
-      def initialize(platform, target = nil)
+      def initialize(platform, target = nil, opts = {})
         super
 
         @pooler = "http://vmpooler.delivery.puppetlabs.net"
--- a/README.md
+++ b/README.md
@@ -73,8 +73,13 @@ Build machines should be cleaned between builds.
 #### Flagged arguments (can be anywhere in the command)
 
 ##### -w DIR, --workdir DIR
-Specifies a directory where the sources should be placed and builds performed.
+Specifies a directory on the controller where the sources should be placed and builds performed.
 Defaults to a temporary directory created with Ruby's Dir.mktmpdir.
+
+##### --remote_dir
+Specify a permament directory on the build host to place sources and perform
+builds. Components can then be rebuild manually on the build host for faster iteraction. Sources may not be correctly updated if this directory already exists.
+Defaults to a temporary directory.
 
 ##### -c DIR, --configdir DIR
 Specifies where project configuration is found. Defaults to $pwd/configs.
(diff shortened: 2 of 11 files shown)


mckern:
How much of this is already expressed in one way or another by the (admittedly poorly documented) `devkit` command?

lib/vanagon/engine/pooler.rb
```diff
@@ -7,7 +7,7 @@ class Pooler < Base
       attr_reader :token
 
       # The vmpooler_template is required to use the pooler engine
-      def initialize(platform, target = nil)
+      def initialize(platform, target = nil, opts = {})
```
mckern: Since Vanagon depends on Ruby 2.1 or newer, it's safe to use keyword arguments. Would that be easier than continuing to use the old Options Hash pattern?

adreyer: I'm ok either way.

adreyer:
This is based on work I did last summer when trying to work on windows agents. Devkit did not solve for my use case then but I can't remember exactly why. Who uses devkit that I can talk to?

MikaelSmith:
Devkit is pretty orphaned and development seems to have improved on what it was trying to do with the `build` command already. I'd be fine dropping it.

Getting what I intended is probably `build <project> <hostname> --preserve --remote_workdir /build --only_build <component>`.

stahnma:
yo @mckern take a look here and make good things happen.

mckern:
@adreyer did @MikaelSmith's note help?

adreyer:
@mckern yeah it sounds like this is the correct path forward. I'll clean this up and remove the don't merge

lib/vanagon/engine/pooler.rb
adreyer: @mckern since none of these actually use specific kwargs this would be a case where I would just `**kwargs` in python to avoid having to update all these signatures when we add new options they don't care about. What should I do in ruby?

mckern: Works pretty much the same in Ruby. Here's an example I stole from someone else:

```ruby
def dual_argument_capturing_method(*args, **keyword_args)
  {args: args, keyword_args: keyword_args}
end

dual_argument_capturing_method(1, 2, 3, key: "value") # => {:args=>[1, 2, 3], :keyword_args=>{:key=>"value"}}
```

Still basically an options hash but a little bit less 1.9-era boilerplate. Ideally we'd just constrain our initialization so we don't have to provide a million arguments, but I'm OK shaving one yak at a time.

stahnma:
@adreyer is this ready to go (other than needs a rebase?) /cc @mckern

melissa:
@adreyer we're getting ready to cut a vanagon 0.10.0 release soon. If we can get this merged in this afternoon or early tomorrow, we can include it in that release

mckern:
Closing in favor of #452

PR #437, opened 2017-01-19
(VANAGON-40) Environmental Activism

mckern:
I'm trying something a little different for this feature addition... a "@rick used to work at github" style checklist and a longish-lived PR allowing others to track the status of a feature as it's developed.

- [X] Support for environment variables as sane objects
- [x] Doc & Tests for env. var class
- [x] DSL methods to define env. vars per `Project`
- [x] DSL methods to define env. vars per `Platform`
- [x] DSL methods to define env. vars per `Component`
- [x] Doc & test for shiny new DSL methods

### DIFF
--- a/lib/vanagon/platform/dsl.rb
+++ b/lib/vanagon/platform/dsl.rb
@@ -18,16 +18,16 @@ class Vanagon
       #
       # @param name [String] name of the platform
       # @return [Vanagon::Platform::DSL] A DSL object to describe the {Vanagon::Platform}
-      def initialize(name)
-        @name = name
+      def initialize(platform_name)
+        @name = platform_name
       end
 
       # Primary way of interacting with the DSL. Also a simple factory to get the right platform object.
       #
       # @param name [String] name of the platform
       # @param block [Proc] DSL definition of the platform to call
-      def platform(name, &block)
-        @platform = case name
+      def platform(platform_name, &block) # rubocop:disable Metrics/AbcSize
+        @platform = case platform_name
                     when /^aix-/
                       Vanagon::Platform::RPM::AIX.new(@name)
                     when /^(cisco-wrlinux|el|fedora)-/
@@ -51,6 +51,7 @@ class Vanagon
                     end
 
         yield(self)
+        environment 'VANAGON_PLATFORM', platform_name.tr('-', '.')
         @platform
       end
 
@@ -81,6 +82,13 @@ class Vanagon
 
       def respond_to_missing?(method_name, include_private = false)
         method_name.to_s.start_with?('get_') || super
+      end
+
+      # Adds an arbitrary environment variable to a Platform, which will be
+      # merged with any environment variables defined by the Project into the
+      # rendered Makefile
+      def environment(key, value)
+        @platform.environment[key] = value
       end
 
       # Set the path to make for the platform
@@ -127,9 +135,10 @@ class Vanagon
 
       # Set the cross_compiled flag for the platform
       #
-      # @param cxx_flag [Boolean] True if this is a cross-compiled platform
-      def cross_compiled(cxx_flag)
-        @platform.cross_compiled = cxx_flag
+      # @param xcc [Boolean] True if this is a cross-compiled platform
+      def cross_compiled(xcc_flag)
+        @platform.cross_compiled = !!xcc_flag
+        environment 'VANAGON_PLATFORM_XCC', @platform.cross_compiled.to_s
       end
 
       # Set the path to rpmbuild for the platform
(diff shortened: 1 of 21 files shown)


lib/vanagon/project.rb
```diff
@@ -46,6 +106,9 @@ def initialize(name, platform)
       @requires = []
       @directories = []
       @settings = {}
+      # Environments are like Hashes but with specific constraints
+      # around their keys and values.
+      @environment = Vanagon::Environment.new
```
melissa: I assume there are going to be two seperate environment objects during a build. One defined here in the project, and one defined in the platform. How are those two going to interact?

mckern:
```make
SHELL = bash

# The way I'm thinking about the break down, a Project will have 
# support for an Environment object which will contain a bunch 
# of top-level env. vars. which would be set at this level.
export VANAGON_STATSD_HOST := 127.0.0.1
export VANAGON_STATSD_PORT := 8125
export VANAGON_TEST_STRING := a test string
export VANAGON_PROJECT_NAME := vanagon-example
# And then that will end up merged with a Platform's Environment object,
# and the resulting complete Environment will be merged in and printed
# at the top scope of the Makefile. These would potentially be defined
# in a macOS platform definition, and they'd override any existing
# values of the same name in the Project's Environment or be appended
# if they don't already exist.
export CPPFLAGS = -O3 -I/opt/puppetlabs/puppet/include -march=core2 -msse4
export CC = pl-gcc-6

.PHONY: inherited
inherited: export VANAGON_TARGET = inherited
inherited: example
	echo "the target is $(@)"
	echo "the compiler is $(CC)"
	echo "and the CFLAGS are $(CPPFLAGS)"
	echo -e "----\n"

# And any component specific exports would end up defined in the Make
# target, and would override what's passed in from the top-level scope.
.PHONY: example
example: export VANAGON_TARGET = example
example: export CC = clang
example: export CPPFLAGS = -O1 -wall
example:
	echo "the target is $(@)"
	echo 'The compiler is $(CC)'
	echo "and the CFLAGS are $(CPPFLAGS)"
	echo -e "----\n"
```

```
mckern@flexo example $ make inherited
echo "the target is example"
the target is example
echo 'The compiler is clang'
The compiler is clang
echo "and the CFLAGS are -O1 -wall"
and the CFLAGS are -O1 -wall
echo -e "----\n"
----

echo "the target is inherited"
the target is inherited
echo "the compiler is pl-gcc-6"
the compiler is pl-gcc-6
echo "and the CFLAGS are -O3 -I/opt/puppetlabs/puppet/include -march=core2 -msse4"
and the CFLAGS are -O3 -I/opt/puppetlabs/puppet/include -march=core2 -msse4
echo -e "----\n"
----

mckern@flexo example $
```

melissa: 👍

mckern:
This is looking really good so far:

```make
export SHELL := $(shell which bash)

tempdir := $(shell mktemp -d -p /var/tmp 2>/dev/null || mktemp -d -t 'tmp')
workdir := $(PWD)

export VANAGON_PROJECT_VERSION = 1.8.3.1.gc9145783
export VANAGON_PROJECT_VENDOR = Puppet Labs <info@puppetlabs.com>
export VANAGON_PROJECT_NAME = puppet-agent
export VANAGON_PLATFORM_NAME = el-7-x86_64
```

mckern:
tumblr_n5b0rx6d2b1qfhhceo8_250

mckern:
Great news. I've built Puppet-Agent for `el-7-x86_64` and `debian-8-armhf` and it was 🙌 ! This work should be tested against AIX and Solaris, and if someone (like @rick maybe?) wanted to take a stab at the ABS engine, that'd be 👌 .

mckern:
Here's some [Makefiles from before the new Environment work](https://gist.github.com/mckern/e238d630ec86718d03bb1a3f9e66c57b) and some Makefiles rendered [after the new Environment work](https://gist.github.com/mckern/195b6fe2f1f82d5159a7da0407ed4ad8). Diffs pending.

melissa:
@mckern did you want to add updates to the readme before this PR is merged?

resources/Makefile.erb
```diff
@@ -1,15 +1,15 @@
-<% dirnames = get_directories.map {|d| d.path } %>
```
melissa: Why did you remove this line? Don't we need to set `dirnames`?

melissa: OH! I found it, nevermind

lib/vanagon/platform/dsl.rb
```diff
@@ -51,6 +51,7 @@ def platform(name, &block)
                     end
 
         yield(self)
+        environment 'VANAGON_PLATFORM', platform_name.tr('-', '.')
```
melissa: should we remove the `-` at a later point in time, since I think that's only really needed when we send metrics. Would it be confusing to have this variable be in the format `el.7.x86_64` rather than `el-7-x86_64`?

melissa:
@mckern do you have a makefile diff yet?

mckern:
@melissa the diffs: https://gist.github.com/mckern/d54f16aae3879e2f7760b0decad21cfa

mckern:
I think I am going to omit adding to the README in this PR. I'm working on a restructuring of the docs, and that might fit better there.

lib/vanagon/platform/dsl.rb
mckern: Well, if we use this value in a metrics aggregation system, `el.7.x86_64` would translate out as:

```
mckern@flexo Downloads $ tree -F el
el/
└── 7/
    └── x86_64/
        ├── stat_name_1
        ├── stat_name_2
        ├── stat_name_3
        ├── stat_name_4
        ├── stat_name_5
        └── stat_name_6
```

melissa:
can you test this with a windows makefile? I'm curious to see what the diff there would look like

mckern:
@melissa all relevant gists have been updated with a Windows 12 Makefile & associated diff.

lib/vanagon/platform/dsl.rb
melissa: Will this variable only be used for stats? Is there a possibility we'll use it for something else? I'm just concerned we've set the precedent for platform tags to have dashes, and changing that paradigm here is a little confusion. That being said, this is the introduction of a new variable, so it's not the worst. I'm not blocking on this, I just want to make sure it's a deliberate decision

mckern: This is intended to be used in metric rollup; nothing else has used this value inside the Makefile, so I'm not *super* worried about it. Open to leaving it alone, or even renaming this variable and using `VANAGON_PLATFORM` for the origina value. @underscorgan, @shrug, @branan feedback? Thoughts?

mckern:
@branan @ScottGarman @McdonaldSeanp do you want to take a look at/chime in on this? Gonna fish for a merge otherwise.

ScottGarman:
I've taken a cursory look at this and I'm generally in favor of it. Adding a section to the README file with some example use cases seems worthwhile. LGTM otherwise and I'd like to play with this when I have more time.

shrug:
Add a changelog update?

PR #438, opened 2017-01-19
(VANAGON-43) Initial pass at metrics gathering

melissa:


resources/Makefile.erb
```diff
@@ -1,11 +1,18 @@
 <% dirnames = get_directories.map {|d| d.path } %>
-export SHELL := $(shell which bash)
 
-tempdir := $(shell mktemp -d -p /var/tmp 2>/dev/null || mktemp -d -t 'tmp')
 workdir := $(PWD)
 
+export SHELL := "${workdir}/profiling_shell.sh"
+export VANAGON_STATSD_HOST ?= 10.32.162.156
+export VANAGON_STATSD_PORT ?= 8125
```
melissa: I need to change these to better defaults before this is merged

resources/metrics/profiling_shell.sh
```diff
@@ -0,0 +1,46 @@
...
+# want to check & preserve that.
+if [[ ${PIPESTATUS[0]} -ne 0 ]]; then
+  exit "${PIPESTATUS[0]}"
+fi
+
+# Finally, if everything completed successfully,
+# finesse the numbers and fire them off to our
+# statsd interface.
+if [[ "${VANAGON_TARGET}" ]] && [[ "${VANAGON_STATSD_HOST}" ]] && [[ "${VANAGON_STATSD_PORT}" ]]; then
+  type bc >/dev/null 2>&1 || { echo >&2 "BC is required for metrics collection.  Aborting."; exit 1; }
```
melissa: I want to figure out which platforms are missing bc, and this seemed like an easy way to do that. This line shouldn't make it into vanagon proper though

resources/Makefile.erb
```diff
@@ -1,11 +1,18 @@
...
+export VANAGON_PROJECT ?= <%= @name %>
+export VANAGON_PLATFORM ?= <%= @platform.name.tr('-','.') %>
```
melissa: This means we'll be able to break out build stats for platform type, version, and architecture. It might be more detail than we need, but it might be nice to have the option to drill down to this if we want to.

resources/metrics/profiling_shell.sh
mckern: Don't use `type`. Use `command`.

```bash
mckern@flexo ~ $ command -v rsync &> /dev/null; echo $?
0
mckern@flexo ~ $ command -v non-existant &> /dev/null; echo $?
1
mckern@flexo ~ $ 
```

resources/Makefile.erb
```diff
@@ -1,11 +1,18 @@
...
+
+tempdir := $(shell mktemp -d -p /var/tmp 2>/dev/null || mktemp -d -t 'tmp')
+
 all: file-list-before-build <%= package_name %>
 
+<%= package_name %>: export VANAGON_TARGET = generate-package
```
melissa: This is the only additional piece of info I thought might be useful to gather beyond the <project>-build, <project>-configure, etc. make targets

resources/metrics/profiling_shell.sh
mckern: Also, we could just add a quick check around the things we need so that lack of external commands (`bc` and `tee`) doesn't blow up the entire build -- a build w/o metrics is probably better than a build that fails because it can't generate metrics? I dunno. I could go either way on it.

melissa: I was going to pull the fail before this gets in, for sure. But I think just the default error should be sufficient moving forward. Like I said, I just wanted to figure out which platforms we need to add bc on

mckern: `bc` is generally considered part of a core UNIX installation, even if it's not part of the POSIX standard. Assume all Linux platforms get a new dependency on `bc` -- no harm specifying that there. macOS, Solaris, and AIX all have it by default (yes, I checked). Windows is the unknown quantity. Got a provisioned Windows builder we can check on?

melissa:
```
$ floaty ssh win-2012r2-x86_64
Warning: Permanently added 'kuvc9v64agwv9zw.delivery.puppetlabs.net,10.32.114.48' (ECDSA) to the list of known hosts.

Administrator@kuvc9v64agwv9zw ~
$ which bc
which: no bc in (/usr/local/bin:/usr/bin:/bin:/cygdrive/c/Windows/system32:/cygdrive/c/Windows:/cygdrive/c/Windows/System32/Wbem:/cygdrive/c/Windows/System32/WindowsPowerShell/v1.0:/cygdrive/c/Program Files/Git/cmd:/cygdrive/c/Packer/SysInternals)
```

mckern: I believe Cygwin provides `bc` as a package. Chocolatey provides `GnuWin`, which includes `bc` as part of it.

resources/Makefile.erb
```diff
@@ -1,11 +1,18 @@
 <% dirnames = get_directories.map {|d| d.path } %>
-export SHELL := $(shell which bash)
 
-tempdir := $(shell mktemp -d -p /var/tmp 2>/dev/null || mktemp -d -t 'tmp')
 workdir := $(PWD)
 
+export SHELL := "${workdir}/profiling_shell.sh"
```
mckern: we shouldn't use `export` here if we don't actually have to, and the value after the assignment doesn't need to be quoted. 

```make
# This should be more than sufficient to set a valid shell for Make to send commands to
SHELL = $(PWD)/profiling_shell.sh
```

resources/metrics/profiling_shell.sh
```diff
@@ -0,0 +1,45 @@
...
+# failing status the pipeline returned. We
+# want to check & preserve that.
+if [[ ${PIPESTATUS[0]} -ne 0 ]]; then
+  exit "${PIPESTATUS[0]}"
+fi
+
+# Finally, if everything completed successfully,
+# finesse the numbers and fire them off to our
+# statsd interface.
+if [[ "${VANAGON_TARGET}" ]] && [[ "${VANAGON_STATSD_HOST}" ]] && [[ "${VANAGON_STATSD_PORT}" ]]; then
```
melissa: @mckern with your work on setting up environments, we shouldn't need to set these in the makefile, correct? We should be able to set them at a project or platform level and have them picked up here without issue?

resources/Makefile.erb
```diff
@@ -1,11 +1,15 @@
...
+SHELL = $(PWD)/profiling_shell.sh
+export VANAGON_PROJECT ?= <%= @name %>
+export VANAGON_PLATFORM ?= <%= @platform.name.tr('-','.') %>
 
 tempdir := $(shell mktemp -d -p /var/tmp 2>/dev/null || mktemp -d -t 'tmp')
 workdir := $(PWD)
 
 all: file-list-before-build <%= package_name %>
 
+<%= package_name %>: export VANAGON_TARGET = create-package
```
melissa: Beyond the individual component metrics, this was the only metric I could think of that would be useful to collect. This doesn't include downloading or unpacking sources, prepping the build host, etc, as that all happens before the makefile is executed. Is there anything I'm leaving out that we want to include?

resources/metrics/profiling_shell.sh
```diff
@@ -0,0 +1,45 @@
...
+  __elapsed="$(echo "scale=0;${__seconds}*1000/1" | bc -l)"
+  echo -n "vanagon.${VANAGON_PROJECT}.${VANAGON_PLATFORM}.${VANAGON_TARGET}:${__elapsed}|ms" > "/dev/udp/${VANAGON_STATSD_HOST}/${VANAGON_STATSD_PORT}"
```
melissa: @Ziaunys this is what we have for metrics gathering. I know you had asked for it to be format independent, but this is our first pass at this work. Do you have any suggestions off the top for your head for easy changes we can make now for that work to be a little easier down the road?

shrug: Is that redirection supported everywhere?

mckern: What's "everywhere"? It's supported in Bash. This is a Bash script.

mckern: OH, are you talking about using `/dev/udp`? That's not actually a character or device file -- that's a Bash specific hook.

shrug: Seems to work on aix 5.1, works for me

resources/Makefile.erb
mckern: I thought sources were shipped across the wire wholesale, and if archived they got unpacked on the remote build target (not on the local coordinator)?

resources/metrics/profiling_shell.sh
mckern: Kinda. Env. vars. at the project & platform level should end up in the Makefile as a result of my work, and the shell that executes them should inherit them.

resources/Makefile.erb
melissa: I think I misspoke about the unpack. I believe the unpack happens in the makefile now that I think about it, but downloading the sources and installing build time dependencies happens outside of the makefile

resources/metrics/profiling_shell.sh
melissa: Awesome, we'll be able to make modifications later if need be as well

resources/metrics/profiling_shell.sh
melissa: @Ziaunys thoughts on this format for metrics collection? I have a feeling we'll expand it at a later date to accommodate different formats as well

Ziaunys: +1 This looks like a good first step. The format looks fine.

 I had a discussion with @melissa where we discussed how to make this more generic down the road. Our conclusion was that we should decouple the reporting mechanism from the steps in the build. For instance, the scripts would emit data that Vanagon collects on the client and then a Vanagon Ruby library can handle how and where the data is written. The benefit is that it is easier to write additional reporters because we don't need to write them in Bash.

I don't think we need to do any additional work here to support the longer-term approach as it doesn't really lock us in to anything. As long as the same data is emitted at the end of the day, it shouldn't matter.

melissa: I'm trying to think how this could be implemented. I keep hitting road blocks

melissa: I'm going to table that effort for now. Metrics will keep the same format when we update it in the future

PR #439, opened 2017-01-23
(VANAGON-28) Rename is_osx? helper to is_macos?

sarameisburger:
Make method name more canonical to what the system actually is.
Added a deprecation warning if is_osx? is still used and then calls is_macos?

### DIFF
--- a/lib/vanagon/platform.rb
+++ b/lib/vanagon/platform.rb
@@ -221,11 +221,22 @@ class Vanagon
       return !!@name.match(/^cisco-wrlinux-.*$/)
     end
 
-    # Utility matcher to determine is the platform is an osx variety
+    # Utility matcher to determine if the platform is an osx variety
     #
+    # @deprecated Please use is_macos? instead
     # @return [true, false] true if it is an osx variety, false otherwise
     def is_osx?
-      return !!@name.match(/^osx-.*$/)
+      warn "is_osx? is a deprecated method, please use #is_macos? instead."
+      is_macos?
+    end
+
+    # Utility matcher to determine if the platform is a macos or osx variety
+    # is_osx is a deprecated method that calls is_macos
+    # We still match for osx currently but this will change
+    #
+    # @return [true, false] true if it is a macos or osx variety, false otherwise
+    def is_macos?
+      !!(@name =~ /^macos-.*$/ || @name =~ /^osx-.*$/)
     end
 
     # Utility matcher to determine is the platform is a solaris variety


stahnma:
Did we have tests on this? Should have tests? 

Feedback from Melissa: Did I see something about hamburger helper on here? 🍔

mckern:
52456_640x428

mckern:
hamburger-helper-album-cover-hed-2016

lib/vanagon/platform.rb
```diff
@@ -221,10 +221,19 @@ def is_cisco_wrlinux?
...
     def is_osx?
+      warn "is_osx? is a deprecated method, please use is_macos? instead."
+      is_macos?
+    end
+
+    # Utility matcher to determine if the platform is a macos variety
+    #
+    # @return [true, false] true if it is a macos variety, false otherwise
+    def is_macos?
       return !!@name.match(/^osx-.*$/)
```
mckern: Thinking about this, we might also want to add a matcher on `macos`; we should preserve old behavior and add new behavior.

lib/vanagon/platform.rb
```diff
@@ -221,11 +221,22 @@ def is_cisco_wrlinux?
...
-    # Utility matcher to determine is the platform is an osx variety
+    # Utility matcher to determine if the platform is an osx variety
     #
+    # @deprecated Please use is_macos? instead
     # @return [true, false] true if it is an osx variety, false otherwise
     def is_osx?
-      return !!@name.match(/^osx-.*$/)
+      warn "is_osx? is a deprecated method, please use is_macos? instead."
+      return true if is_macos?
+      return true if @name =~ /^osx-.*$/
```
underscorgan: Are people going to see the deprecation warning and try to use `is_macos` when the platform is still osx something something? That sounds like something I would do, and this would break that.

lib/vanagon/platform.rb
```diff
@@ -221,11 +221,22 @@ def is_cisco_wrlinux?
...
+      false
```
mwaggett: Couldn't you just say `return is_macros? || (@name =~ /^osx-.*$/)` ?

lib/vanagon/platform.rb
mckern: Yeah, maybe. What do you think a good workflow should be?

mwaggett: What if the osx match was built into the `is_macos?` function?

mckern: It'd kind of overload the function and make it a little more work to untangle when it's removed, not just deprecated.

lib/vanagon/platform.rb
```diff
@@ -221,11 +221,22 @@ def is_cisco_wrlinux?
...
-      return !!@name.match(/^osx-.*$/)
+      warn "is_osx? is a deprecated method, please use #is_macos? instead."
+      is_macos?
+    end
+
+    # Utility matcher to determine if the platform is a macos or osx variety
+    # is_osx is a deprecated method that calls is_macos
+    # We still match for osx currently but this will change
+    #
+    # @return [true, false] true if it is a macos or osx variety, false otherwise
```
mckern: Nitpick: This could totally say `@return [Boolean] true if it is a macos or osx variety, false otherwise`; yardoc supports `Boolean` as a shortcut around `true, false` here. I don't think whoever wrote a lot of these knew that at the time. We can fix it later.

PR #441, opened 2017-01-26
(CPR-391) Fedora packages have odd distro tag

mwaggett:
This commit adjusts how `dist` is defined for rpms. We now only use `os_name`
and `os_version` to construct `dist` if it isn't otherwise specified. This will
allow us to change the fedora distro tags without disrupting the way other
distro tags are currently defined.

- [x] Implement `dist` accessors
- [x] Implement tests for new `dist` accessors
- [x] Validate old behavior is preserved

### DIFF
--- a/spec/lib/vanagon/platform/windows_spec.rb
+++ b/spec/lib/vanagon/platform/windows_spec.rb
@@ -1,7 +1,7 @@
+require 'tmpdir'
 require 'vanagon/platform'
 require 'vanagon/project'
 require 'vanagon/common'
-
 
 # These constants are defined for the purpose of the project/generic file merge tests
 # to point these directories to test areas under the /tmp directory.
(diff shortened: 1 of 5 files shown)


mckern:
Added piggybacking: I cleaned up some indentation on the Windows rspec tests while I was in here. They're isolated to their own commit.

spec/lib/vanagon/platform/windows_spec.rb
```diff
@@ -253,25 +255,26 @@
...
-</Directory>
-<Directory Name="twoUpAgain" Id="twoUpAgain_2_2">
-<Directory Id="SERVICETEST3BINDIR" />
-</Directory>
-</Directory>
-</Directory>
-HERE
-            )
+            expect(cur_plat._platform.generate_service_bin_dirs([comp._component.service, comp2._component.service, comp3._component.service].flatten.compact, proj._project))
+              .to eq( \
```
mwaggett: @mckern I don't know what the backslash here does, but you removed it in the rest of your cleanup, so not sure if you wanted it here or not.

mckern: Good catch. Someone intended it as a line-continuation character, but that isn't needed. I missed this one when I was un-bashifying that set of spec tests. Do you want to edit that commit, rebase, and update this PR? If not, I can do it later.

mwaggett: @mckern I can do it. 😄

PR #442, opened 2017-01-27
(FEATURE) Add `render` command

mckern:
`render` will generate just the Makefile for a given Vanagon project
and place it in a temporary directory, where it can be inspected
without interrupting/halting/aborting a project build. It will install
no dependencies, and do no engine/build target provisioning. This is
intended to help developers fine-tune any project changes that would
affect the Makefile before a project is built.

### DIFF
--- a/bin/render
+++ b/bin/render
@@ -1,0 +1,27 @@
+#!/usr/bin/env ruby
+load File.expand_path(File.join(File.dirname(__FILE__), "..", "lib", "vanagon.rb"))
+
+optparse = Vanagon::OptParse.new(
+  "#{File.basename(__FILE__)} <project-name> <platform-name> [options]",
+  [:workdir, :configdir, :engine]
+)
+options = optparse.parse! ARGV
+
+project = ARGV[0]
+platforms = ARGV[1]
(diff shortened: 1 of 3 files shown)


branan:
Why spit it out to a tempfile instead of just to stdout? Redirecting to a file is easy if needed, and sticking to unix paradigms makes using all the normal text searching and manipulation tools easy.

mckern:
Because it was the least code -- it's just a hacked up copy of the `#run` method. What'd I do for `inspect` the last time I got mad at Vanagon? We should do that.

`/me goes to see what he did last time`

mckern:
Don't merge this. Something's wrong.

haus:
oh god, `erb_file`. that method gives me all the sads.

mckern:
Should be all fixed up now:

```
mckern@flexo puppet-agent (git:detached*) $ bundle exec render puppet-agent el-7-x86_64
rendering Makefile
Generated: /var/folders/jg/0fyykqj50wgc48f_6bygnblr0000gn/T/d20170126-6875-1dtt3op/Makefile
```

PR #448, opened 2017-02-01
(RE-5646) add_source unpacks compressed second source

sarameisburger:
Before, if a second source was compressed it could not be unpacked because we assumed it was along the lines of a conf file. Now handles second sources that are archives and unpacks them appropriately.

PR #453, opened 2017-02-13
(RE-5646) add_source unpacks compressed second source & added spec tests

sarameisburger:
Before, if a second source was compressed it could not be unpacked because we assumed it was along the lines of a conf file. Now handles second sources that are archives and unpacks them appropriately. Spec tests were also added to ensure that the secondary source was added and unpacked properly.

Previous discussion of this PR can be found [here](https://github.com/puppetlabs/vanagon/pull/448)

### DIFF
--- a/lib/vanagon/component.rb
+++ b/lib/vanagon/component.rb
@@ -231,13 +231,18 @@ class Vanagon
     # Fetches secondary sources for the component. These are just dumped into the workdir currently.
     #
     # @param workdir [String] working directory to put the source into
-    def get_sources(workdir)
+    def get_sources(workdir) # rubocop:disable Metrics/AbcSize
+      # create an empty array if @extract_with doesn't already exist
+      @extract_with = Array(@extract_with)
       sources.each do |source|
         src = Vanagon::Component::Source.source(
           source.url, workdir: workdir, ref: source.ref, sum: source.sum
         )
         src.fetch
         src.verify
+        # set src.file to only be populated with the basename instead of entire file path
+        src.file = File.basename(src.url)
+        @extract_with << src.extract(platform.tar) if src.respond_to?(:extract)
       end
     end
 
--- a/lib/vanagon/component.rb
+++ b/lib/vanagon/component.rb
@@ -29,7 +29,8 @@ class Vanagon
     # that they were cloned to. For the outlying flat files, it'll
     # end up being defined explicitly as the string './'
     attr_accessor :dirname
-    # what special tool should be used to extract the primary source
+    # Which specific tool or command line invocations that
+    # should be used to extract a given component's primary source?
     attr_accessor :extract_with
     # how should this component be configured?
     attr_accessor :configure
@@ -153,7 +154,6 @@ class Vanagon
       @postinstall_actions = []
       @preremove_actions = []
       @postremove_actions = []
-      @extract_with = []
     end
 
     # Adds the given file to the list of files and returns @files.
@@ -191,7 +191,7 @@ class Vanagon
     end
 
     # Fetches the primary source for the component. As a side effect, also sets
-    # \@extract_with, @dirname and @version for the component for use in the
+    # @extract_with, @dirname and @version for the component for use in the
     # makefile template
     #
     # @param workdir [String] working directory to put the source into
@@ -201,7 +201,7 @@ class Vanagon
         @source = Vanagon::Component::Source.source(url, opts)
         source.fetch
         source.verify
-        extract_with_set(source.respond_to?(:extract) ? source.extract(platform.tar) : nil)
+        extract_with << source.extract(platform.tar) if source.respond_to? :extract
 
         @cleanup_source = source.cleanup if source.respond_to?(:cleanup)
         @dirname = source.dirname
@@ -217,7 +217,7 @@ class Vanagon
         @dirname = './'
 
         # If there is no source, there is nothing to do to extract
-        extract_with_set(': no source, so nothing to extract')
+        extract_with << ': no source, so nothing to extract'
       end
     end
 
@@ -242,13 +242,14 @@ class Vanagon
         src.verify
         # set src.file to only be populated with the basename instead of entire file path
         src.file = File.basename(src.url)
-        extract_with_set(src.extract(platform.tar)) if src.respond_to?(:extract)
+        extract_with << src.extract(platform.tar) if src.respond_to? :extract
       end
     end
 
-    # create setter method for extract_with to add sources to the array
-    def extract_with_set(extract)
-      @extract_with << extract
+    # @return [Array] the specific tool or command line invocations that
+    #   should be used to extract a given component's primary source
+    def extract_with
+      @extract_with ||= []
     end
 
     # Fetches patches if any are provided for the project.
--- a/lib/vanagon/component.rb
+++ b/lib/vanagon/component.rb
@@ -153,6 +153,7 @@ class Vanagon
       @postinstall_actions = []
       @preremove_actions = []
       @postremove_actions = []
+      @extract_with = []
     end
 
     # Adds the given file to the list of files and returns @files.
@@ -200,7 +201,8 @@ class Vanagon
         @source = Vanagon::Component::Source.source(url, opts)
         source.fetch
         source.verify
-        @extract_with = source.respond_to?(:extract) ? source.extract(platform.tar) : nil
+        extract_with_set(source.respond_to?(:extract) ? source.extract(platform.tar) : nil)
+
         @cleanup_source = source.cleanup if source.respond_to?(:cleanup)
         @dirname = source.dirname
 
@@ -215,7 +217,7 @@ class Vanagon
         @dirname = './'
 
         # If there is no source, there is nothing to do to extract
-        @extract_with = ': no source, so nothing to extract'
+        extract_with_set(': no source, so nothing to extract')
       end
     end
 
@@ -232,8 +234,6 @@ class Vanagon
     #
     # @param workdir [String] working directory to put the source into
     def get_sources(workdir) # rubocop:disable Metrics/AbcSize
-      # create an empty array if @extract_with doesn't already exist
-      @extract_with = Array(@extract_with)
       sources.each do |source|
         src = Vanagon::Component::Source.source(
           source.url, workdir: workdir, ref: source.ref, sum: source.sum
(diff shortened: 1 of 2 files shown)


mckern:
Rebase & squash before review?

underscorgan:
@sarameisburger it looks like in the squashing you picked up all my garbage commit messages, we should clean those up :O

lib/vanagon/component.rb
```diff
@@ -231,13 +231,18 @@ def get_build_dir
     # Fetches secondary sources for the component. These are just dumped into the workdir currently.
     #
     # @param workdir [String] working directory to put the source into
-    def get_sources(workdir)
+    def get_sources(workdir) # rubocop:disable Metrics/AbcSize
+      # create an empty array if @extract_with doesn't already exist
+      @extract_with = Array(@extract_with)
```
mckern: Using `[@extract_with]` would be a bit more idiomatic than `Array(@extract_with)`. Not a deal breaker, but worth pointing out. Also, use the accessor (`the #extract_with` method) instead of the instance variable (`@extract_with`).

lib/vanagon/component.rb
```diff
@@ -231,13 +231,18 @@ def get_build_dir
...
       sources.each do |source|
         src = Vanagon::Component::Source.source(
           source.url, workdir: workdir, ref: source.ref, sum: source.sum
         )
         src.fetch
         src.verify
+        # set src.file to only be populated with the basename instead of entire file path
+        src.file = File.basename(src.url)
+        @extract_with << src.extract(platform.tar) if src.respond_to?(:extract)
```
mckern: Same as above: accessor, not instance variable.

spec/lib/vanagon/component_spec.rb
```diff
@@ -45,20 +46,32 @@
...
 
     it "copies secondary sources into the workdir" do
-      subject.get_sources(@workdir)
+      component = subject
+      component.get_sources(@workdir)
       expect(File.exist?(File.join(@workdir, @file_name))).to be true
+      # make sure that our secondary source(s) made it into the workdir
+      expect(File.exist?(File.join(@workdir, "#{@fake_dir}.tar.gz"))).to be true
+      expect(component.extract_with.join(" && ")).to match "#{@fake_dir}.tar.gz"
     end
```
mckern: Instead of modifying the existing test, considering adding a new test that validates that the compressed files are expanded. Leave the existing test for uncompressed files alone, because the expectation is that secondary sources can be either kind of file.

mckern:
Why the new PR?

lib/vanagon/component.rb
underscorgan: May not be an issue since we should only get_sources once, but was mostly using `Array` because `foo = ['a']; Array(foo) = ['a']` whereas `[foo] = [['a']]`

spec/lib/vanagon/component_spec.rb
underscorgan: Ok, that could be moved into a new test, though the existing test would need some extra stubbing at least so it will properly have a platform. By 'validates the compressed files are expanded' do you mean as it is tested here, or test that the makefile rules run and the tarball is extracted?

sarameisburger:
@mckern there was a branch issue so I thought it would be safer to just make a new PR. I'll start on those other changes.

lib/vanagon/component.rb
mckern: I didn't realize you were coercing the value of `@extract_with` to an empty array, even if it's already an array. Wouldn't that be better handled by using the accessor than the instance variable, and using the accessor to ensure that `extract_with` returns what you need it to?

```ruby
# example only; didn't do the research
def extract_with
  @extract_with ||= []
end
```

mckern:
Branch issue? Like a merge conflict?

mckern:
As per our conversation on Monday, your new accessor could be simplified a bit by taking advantage of Ruby's implicit returns and automatic accessors:

```ruby
# This is WAY cut down from the normal class, but it should (hopefully)
# illustrate some of the stuff I talked about on Monday
class Vanagon
  class Component
    attr_writer :extract_with

    def get_source(workdir)
      if url
        # <---✂---✂ snip ✂---✂--->
        if source.respond_to?(:extract)
          extract_with << source.extract(platform.tar)
        end
      else
        # <---✂---✂ snip ✂---✂--->
        extract_with << ': no source, so nothing to extract'
      end
    end

     def get_sources(workdir) # rubocop:disable Metrics/AbcSize
      # <---✂---✂ snip ✂---✂--->
      extract_with << src.extract(platform.tar) if src.respond_to?(:extract)
    end
 
    # @return [Array]
    def extract_with
      @extract_with ||= []
    end
  end
```

spec/lib/vanagon/component_spec.rb
mckern: @underscorgan I may dig into this test today so we can tie this PR off

underscorgan:
I want to just run a quick test for this.

mckern:
No rush.

underscorgan:
Ok, rendering off of a hacky branch that adds some files in a tar as a secondary results in:
```
pe-activemq-unpack: file-list-before-build
  gunzip -c "apache-activemq-5.14.3-bin.tar.gz" | tar xf - && \
  : nothing to extract && \
  : nothing to extract && \
  gunzip -c "pe-activemq-properties.tar.gz" | tar xf - && \
  : nothing to extract && \
  : nothing to extract && \
  : nothing to extract && \
  : nothing to extract
  touch pe-activemq-unpack
```

which is what I expect. HOORAY 🎉

lib/vanagon/component.rb
```diff
@@ -29,7 +29,8 @@ class Component
     # that they were cloned to. For the outlying flat files, it'll
     # end up being defined explicitly as the string './'
     attr_accessor :dirname
-    # what special tool should be used to extract the primary source
+    # Which specific tool or command line invocations that
+    # should be used to extract a given component's primary source?
```
mwaggett: I don't understand this comment, mostly bc it seems ungrammatical.

lib/vanagon/component.rb
```diff
@@ -231,16 +233,25 @@ def get_build_dir
     # Fetches secondary sources for the component. These are just dumped into the workdir currently.
     #
     # @param workdir [String] working directory to put the source into
-    def get_sources(workdir)
+    def get_sources(workdir) # rubocop:disable Metrics/AbcSize
```
mwaggett: probably don't want this comment here?

melissa: We need this comment to tell rubocop that's it's okay this method is failing checks and to not error out because of it

lib/vanagon/component.rb
```diff
@@ -231,16 +233,25 @@ def get_build_dir
...
       sources.each do |source|
         src = Vanagon::Component::Source.source(
           source.url, workdir: workdir, ref: source.ref, sum: source.sum
         )
         src.fetch
         src.verify
+        # set src.file to only be populated with the basename instead of entire file path
+        src.file = File.basename(src.url)
+        extract_with << src.extract(platform.tar) if src.respond_to? :extract
       end
```
mwaggett: I'm confused as to why we have get_source AND get_sources. A lot of the behavior looks to be the same, so there could probably be some DRYing up of this.

underscorgan: Yes, I agree, though tend to think that will be a bigger change and probably outside the scope of this PR.

lib/vanagon/component.rb
mwaggett: oookay

lib/vanagon/component.rb
mwaggett: It just seems like one of those things where if we don't do it now, it'll never happen.. 
I may need more info/context as to what each of these methods are for, but in my mind, when I see a method `do_x` and another method `do_xs`, I expect the latter to be something along the lines of `for xs.each { |x| do_x }`. ya know?

underscorgan: so get_source is used for the primary source which ends up getting some special treatment as opposed to the secondary sources which are stored in get_sources.  The fact that they're being treated differently now doesn't mean they should be, but unraveling the differences that matter and ones that can be changed feels like it could be a huge yak.

For example, the primary source (get_source) is set via the project url, and all of the secondary are set via add_source.

I don't think we should block this work on figuring this out, but we can and should absolutely add a ticket in the VANAGON project.

mwaggett: In that case, I think we should have less-confusing method names. Maybe `get_secondary_sources` instead of `get_sources`?

underscorgan: Can that go in a separate PR after this one is merged? I tend to like to keep renaming/refactoring isolated from feature work so it's easier in the event things need to be reverted.

lib/vanagon/component.rb
underscorgan: would this be better?

```
# Which specific tools or command line invocations
# should be used to extract a given component's sources
```

lib/vanagon/component.rb
mwaggett: yeah. I didn't look closely enough and I think I thought `get_sources` was a new thing.

lib/vanagon/component.rb
mwaggett: what about
```
# The specific tools or command line invocations that
# should be used to extract a given component's sources
```
?

lib/vanagon/component.rb
underscorgan: ah fair

lib/vanagon/component.rb
underscorgan: 👍

underscorgan: updated

lib/vanagon/component.rb
```diff
@@ -215,7 +217,7 @@ def get_source(workdir) # rubocop:disable Metrics/AbcSize
         @dirname = './'
 
         # If there is no source, there is nothing to do to extract
-        @extract_with = ': no source, so nothing to extract'
+        extract_with << ': no source, so nothing to extract'
```
mckern: I wonder if I could ditch this line entirely. An empty `Array` is a safe array, and it'd cut down on chatter and verbosity in the Makefile.

man_presses_fingers_together_thinking_something_cg4p41662c

underscorgan: I think this could go away, but I think the makefile is still going to be a mess because of https://github.com/puppetlabs/vanagon/blob/ae1f319dbd48ed20f4e65929cf71b2d2208f67da/lib/vanagon/component/source/local.rb#L82

mckern: FYI I'm not ditching this in this PR. It'd require strenuous testing due to how Makefiles are rendered (a lot of `&&`ing).

spec/lib/vanagon/component_spec.rb
mckern: Disregard these comments since decompression happens in the Makefile and there's no useful framework for testing Makefile rules. This is as good as I think it's gonna get, minus a nitpick or two.

melissa:
when was this last tested?

mckern:
About 10 minutes ago.

PR #455, opened 2017-02-22
(MAINT) Fix egregious logic error in comp. DSL

mckern:
An array has no key or value, so this obviously won't work. I'm pretty
embarrassed that this was missed.


underscorgan:
???

PR #456, opened 2017-02-23
(maint) Allow owner, group, and mode to be passed to configfiles

underscorgan:
These options are already supported when creating files, but not
supported in the wrapper that creates configfiles. You want to be able
to set these options for config files.


underscorgan:
Tested with my in-progress activemq builds.

mckern:
Merging because how could `HEAD` of `master` be any MORE broken? We should probably test this across platforms but YOLO.

PR #457, opened 2017-02-28
(maint) Don't follow symlinks while making file lists

underscorgan:
If we're linking a directory we only want to add the link, not all of
the files inside the linked directory.

### DIFF
--- a/resources/Makefile.erb
+++ b/resources/Makefile.erb
@@ -17,14 +17,14 @@ file-list-before-build:
 <%- if dirnames.empty? -%>
 	touch file-list-before-build
 <%- else -%>
-	(<%= @platform.find %> -L "<%= dirnames.join('" "') %>" 2>/dev/null || <%= @platform.find %> "<%= dirnames.join('" "') %>" -follow 2>/dev/null) | <%= "xargs -I[] cygpath --mixed --long-name --absolute [] |" if @platform.is_windows? %> <%= @platform.sort %> | uniq > file-list-before-build
+	(<%= @platform.find %> -H "<%= dirnames.join('" "') %>" 2>/dev/null || <%= @platform.find %> "<%= dirnames.join('" "') %>" 2>/dev/null) | <%= "xargs -I[] cygpath --mixed --long-name --absolute [] |" if @platform.is_windows? %> <%= @platform.sort %> | uniq > file-list-before-build
 <%- end -%>
 
 file-list-after-build: <%= @components.map {|comp| comp.name }.join(" ") %>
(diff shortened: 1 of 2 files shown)


mckern:
What platforms has this been tested against?

underscorgan:
Building agent with 0.9.2 + 6f6da5d556661f692bb8b5744bc18cfc0c5f5168 + this at http://jenkins-release.delivery.puppetlabs.net/job/vanagon_generic_job/709/

Currently building the same sha of puppet-agent for el6 to verify there are no differences. This has been used to build pe-activemq which needed directory linking also.

mckern:
Can you test this on macOS and Windows? Don't care about the arch or anything -- just want to make sure we check some of our outlier platforms.

underscorgan:
@mckern just that they install + look ok? built fine with those platforms but haven't tested installing on weirder platforms yet.

underscorgan:
Verified that on el-6 p-a built with this vs Gemfile default had no differences (using `rpm -qplv`)

underscorgan:
Installed on osx, was able to get versions, run facter, run puppet apply. Testing on a windows now.

underscorgan:
Installed on windows and ran the same tests I ran on osx (versions for all the components, ran facter, puppet apply).

shrug:
Are there any cases where we link something into the build from outside... where the package might not own the link source?

underscorgan:
If we're linking something from outside where we don't own the source, should we own the subfiles? Would that lead to deleting something we don't own?

So like, if in some weird hypothetical packaging we need /etc/httpd linked locally and have
```
ln -s /etc/httpd /opt/myapp/httpd
```
With this the package would end up owning `/opt/myapp/httpd` but none of the other files under `/etc/httpd` and  that feels right? Though I may be thinking about this incorrectly.

underscorgan:
investigating some weird local behaviour

underscorgan:
rpmbuild 4.11 is stricter about file types (namely it fails if something says it's a dir but it isn't). 	f1a98ef updates our file-list-for-rpm munging to not say that links are dirs.

shrug:
👍  pending travis

melissa:
Can you test building razor with this?

underscorgan:
@melissa http://jenkins-release.delivery.puppetlabs.net/job/vanagon_generic_job/ it built!

`rpm -qplv` reports no differences on el7, so should be good!

melissa:
I would merge it, but I can't :( the button is greyed out

underscorgan:
@melissa lemme try some force-pushing to see if i can make travis kickoff

PR #459, opened 2017-03-01
(VANAGON-47) Vanagon 0.10.0 is embarrassingly broken

mckern:
- [X]  Windows builds of Puppet Agent (P-A) will no longer work at all because the new `profiling_shell.sh` doesn't run correctly on Windows.
- [X]  The interface for parsing the old implementation of `Component` environment variables is broken because `Arrays` don't have indices like `Hashes` do (fixed in previous PR).

There are refinements needed to metric retrieval now that they're being batched to single-field CSV files on the remote build target -- they will be in a follow-up PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,21 +1,32 @@
 ![Build Status](https://travis-ci.org/puppetlabs/vanagon.svg?branch=master)
-The Vanagon Project
-===
- * What is vanagon?
- * Runtime requirements
- * Configuration and Usage
- * Overview
- * Contributing
- * License
- * Maintainers
- * Support
+# The Vanagon Project
 
-What is vanagon?
----
+<!-- MarkdownTOC -->
+
+- [What is Vanagon?](#what-is-vanagon)
+- [Runtime Requirements](#runtime-requirements)
+  - [Local Host:](#local-host)
+  - [Remote Build Target:](#remote-build-target)
+- [Configuration and Usage](#configuration-and-usage)
+  - [`build` usage](#build-usage)
+  - [`inspect` usage](#inspect-usage)
+  - [`devkit` usage](#devkit-usage)
+- [Engines](#engines)
+  - [Amazon Ec2](#amazon-ec2)
+- [Contributing](#contributing)
+- [License](#license)
+- [Overview](#overview)
+- [Maintainers](#maintainers)
+
+<!-- /MarkdownTOC -->
+
+
+## What is Vanagon?
+
 Vanagon is a tool to build a single package out of a project, which can itself
 contain one or more components. This tooling is being used to develop the
 puppet-agent package, which contains components such as openssl, ruby, and
-augeas among others. For a simple example, please see the examples directory.
+augeas among others. For a simple example, please see the project in the `examples/` directory.
 
 Vanagon builds up a Makefile and packaging files (specfile for RPM,
 control/rules/etc for DEB) and copies them to a remote host, where make can be
@@ -57,15 +68,14 @@ Also, Vanagon ships with a number of engines which may include additional option
 - An ssh server ([homepage](https://www.openssh.com/)) is required by most engines
 - The command line tool `rsync` ([homepage](https://rsync.samba.org/)) (At least rsync 2.6.x)
 
-Configuration and Usage
----
+## Configuration and Usage
+
 Vanagon won't be much use without a project to build. Beyond that, you must
 define any platforms you want to build for. Vanagon ships with some simple
 binaries to use, but the one you probably care about is named 'build'.
 
 ### `build` usage
 The build command has positional arguments and position independent flags.
-
 
 #### Arguments (position dependent)
 
@@ -79,7 +89,7 @@ The name of the target platform to build `<project_name>` against; a file named
 directory. This can also be a comma separated list of platforms such as `platform1,platform2`;
 note that there are no spaces after the comma.
 
-##### target host [optional]
+##### target host <optional>
 Target host is an optional argument to override the host selection. Instead of using
 a random VM collected from the pooler (Vanagon's default build engine), the build will
 attempt connect to the target host over SSH as the `root` user.
@@ -92,7 +102,9 @@ than platforms, the extra platforms will be ignored.
 
 Build machines should be cleaned between builds.
 
-#### Flagged arguments (can be anywhere in the command)
+#### Flagged arguments
+
+**Note:** command flags can be used anywhere in the command.
 
 ##### -w DIR, --workdir DIR
 Specifies a directory on the local host where the sources should be placed and builds performed.
@@ -118,13 +130,15 @@ Currently supported engines are:
 
 #### Flags
 
+**Note:** command flags can be used anywhere in the command.
+
 ##### -p, --preserve
 Indicates that the host used for building the project should be left intact
 after the build instead of destroyed. The host is usually destroyed after a
 successful build, or left after a failed build.
 
-##### -v, --verbose (not yet implemented)
-Increase verbosity of output.
+##### -v, --verbose
+(Reserved for future implementation) Will increase the verbosity of output, when implemented.
 
 ##### -h, --help
 Display command-line help.
@@ -202,7 +216,7 @@ Platform can also be a comma separated list of platforms such as platform1,platf
 Specifies a directory where the sources should be placed and builds performed.
 Defaults to a temporary directory created with Ruby's Dir.mktmpdir.
 
-##### -c DIR, --configdir DIR
+##### -c DIR, --configdir DIR <optional>
 Specifies where project configuration is found. Defaults to $pwd/configs.
 
 ##### -e ENGINE, --engine ENGINE
@@ -249,7 +263,7 @@ As in `build` arguments.
 ##### platform name
 As in `build` arguments.
 
-##### component names [optional]
+##### component names <optional>
 Specifies specific components that should be built. If components are not
 specified, then all components in the project will be built. If components
 are specified as arguments, then any in the project that aren't specified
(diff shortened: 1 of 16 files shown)


mckern:
Travis is going to throw up on this because I haven't updated the Makefile tests. I'll do it after I sleep.

resources/metrics/profiling_shell.sh
```diff
@@ -1,45 +1,55 @@
...
+# directly in Bash without this wrapper. This will halt the job
+# without sending metrics for this job upstream.
+for value in "${__status[@]}"; do
+  if [[ ${value} -ne 0 ]]; then
+    exit "${value}"
+  fi
+done
+
+# Batch the runtime for this task
+echo "${__seconds}" >> "${__batchfile}"
```
melissa: when is this getting sent over to the statsd host?

resources/metrics/profiling_shell.sh
```diff
@@ -1,45 +1,55 @@
 #!/bin/bash
 
 # this variant of the Profiling shell wrapper
-# uses only the Bash keyword `time`, and the generally
+# uses only the Bash keyword `time`, `perl` to computer
```
shrug: 'compute' instead of 'computer'?

lib/vanagon/environment.rb
```diff
@@ -133,21 +133,41 @@ def to_s
...
+        by the shell. These variables will be mangled for you for now, but you should
+        update your project's parameters.
       eos
 
       warn warning.join("\n")
-      str.gsub(/\$\$([\w]+)/, '$(\1)')
+      str.gsub(pattern, '$(\1)')
     end
-    private :sanitize_value
+    private :sanitize_subshells
```
mwaggett: Should this be `:sanitize_variables`? Right now this duplicates line 152.

resources/metrics/profiling_shell.sh
```diff
@@ -1,45 +1,55 @@
...
-__stderr="${VANAGON_TARGET}.stderr.log"
+# relatively easily. While you're at it, store runtimes in batches.
+if [[ -n ${VANAGON_TARGET} ]]; then
+  __batchfile="${VANAGON_TARGET}.csv"
+  __stdout="${VANAGON_TARGET}.stdout.log"
+  __stderr="${VANAGON_TARGET}.stderr.log"
+else
+  # Alternatively, if there's no defined Target, just
+  # dump that stuff in the bitbucket
+  __batchfile="/dev/null"
```
shrug: How does this resolve on Windows? I assume cygwin?

shrug:
do you have a before/after makefile comparison?

underscorgan:
@mckern what have you built with this?

mckern:
Will address all feedback shortly, but this is gonna break AIX builds in a hilarious way: AIX is using GNU Make 3.80 (released in 2002) but we decided to target GNU Make 3.81 (released in 2006) features as our baseline; that should be safe because:

a) that's (at least) what many (most?) platforms ship out of the box
b) **2002 WAS 15 YEARS AGO**

So... time to dig into updating AIX GNU Make, and documenting this breaking change.

geoffnichols:
Left same comment in [puppet-agent/pull/1000](https://github.com/puppetlabs/puppet-agent/pull/1000), but looks like IBM may provide a GNU Make 4.1 rpm via their ["AIX Toolbox for Linux Applications"](http://www-03.ibm.com/systems/power/software/aix/linux/toolbox/alpha.html#M).

resources/metrics/profiling_shell.sh
mckern: It's not, right now. I'm working on the code to slurp it back. Sending it from the remote target just wasn't working right and it was getting to the point where getting `profiling_shell.sh` to actually run on Windows was more important. I'm working on code to slurp these back for post-processing now.

lib/vanagon/environment.rb
mckern: Good catch. Fixed.

resources/metrics/profiling_shell.sh
mckern: I dunno, I kinda like the hacky charm of this typo.

resources/metrics/profiling_shell.sh
mckern: Correct.

mckern:
@underscorgan I have built P-A for Linux and Windows. Will be doing more comprehensive testing ASAP now that the "`make` is too effing old` problem.

@shrug I will get you some diff's but be aware that this is going to be a breaking change for some projects that worked in 0.9.x/0.8.x. You want a P-A diff, or another project?

shrug:
If not a diff, then maybe a before/after makefile from a representative project, probably doesn't matter too much which one

README.md
```diff
@@ -210,6 +246,7 @@ build-time configuration, environment variables, and expected artifacts.
 
 ---
 
+<!-- deprecate devkit -->
```
melissa: What's the plan for this? Why not just add a depreciation note rather than a header?

mckern: It's as much a note to myself as anything else right now; I might remove the commit entirely but I didn't want to forget that this is something I wanted to do.

README.md
```diff
@@ -96,13 +129,15 @@ Currently supported engines are:
 
 #### Flags (can be anywhere in the command)
```
mwaggett: probably want to remove this redundancy

resources/metrics/profiling_shell.sh
mckern: This is now addressed; when `Platform.new('name-version-arch') { |plat| plat.profiling true }` is enabled then the remote `runtimes/` directory will be retrieved after a successful build. We can tune/tweak that behavior in the future if we decide we want to be able to retrieve metrics from failed builds.

README.md
mckern: Fixed!

lib/vanagon/component.rb
```diff
@@ -268,7 +268,7 @@ def get_environment
       if @environment.empty?
```
shrug: Still want an instance variable here?

resources/metrics/profiling_shell.sh
```diff
@@ -1,45 +1,68 @@
...
+# relatively easily. While you're at it, store runtimes in batches.
+if [[ -n ${VANAGON_TARGET} ]]; then
+  __batchfile="runtimes/${VANAGON_TARGET}.csv"
+  __stdout="logs/${VANAGON_TARGET}.stdout.log"
+  __stderr="logs/${VANAGON_TARGET}.stderr.log"
+  mkdir "logs" "runtimes"
+else
+  # Alternatively, if there's no defined Target, just
+  # dump that stuff in the bitbucket
+  __batchfile="/dev/null"
```
shrug: There was a previous comment, might have been lost - how does /dev/null resolve on windows? Do we depend in cygwin on the build environment?

mckern: It wasn't lost. I replied to it. It's resolved by cygwin's POSIX layer.

underscorgan:
throwing a don't merge on here until bugs with cross compiles are ironed out, just so we don't accidentally merge...

README.md
```diff
@@ -28,23 +39,44 @@ creates a master makefile for the project, and configures, builds, and installs
...
-for the target can be customized in the platform config for the target).
+## Runtime Requirements
+
+Vanagon carries two sets of requirements: requirements for the local host where Vanagon is run, and for the remote target where compilation and packaging happens.
+
+Also, Vanagon ships with a number of engines which may include additional optional dependencies if you wish to use them. These engines are currently considered experimental, and receive less attention than the `Hardware` or `vmpooler` engines do. If you find a bug in these engines, [please open a ticket](https://tickets.puppetlabs.com/browse/PA) and let us know.
+
+### Local Host:
+
+- A recent version of [Ruby](https://www.ruby-lang.org/en/) (At least Ruby 2.1.x or newer)
```
haus: Nit: "At least Ruby 2.1.x or newer" is redundant, "At least Ruby 2.1.x" or "newer than Ruby 2.1.x" would work.

README.md
```diff
@@ -28,23 +39,44 @@ creates a master makefile for the project, and configures, builds, and installs
...
+- [AWS SDK for Ruby](https://github.com/aws/aws-sdk-ruby), if you're using the EC2 engine
+- [Docker](https://www.docker.com/community-edition), if you're using the Docker engine
+
+### Remote Build Target:
+
+**Note:** package installation & builder configuration for the remote target can be customized in the `Platform` configuration that defines target provisioning instructions.
+
+- GNU Make ([homepage](https://www.gnu.org/software/make/)) (Vanagon specifically targets the feature set provided by [GNU Make 3.81](http://git.savannah.gnu.org/cgit/make.git/tree/NEWS?h=3.81&id=776d8b7bc2ff83f8ebf5d357ec89e3bbe6d83962) but newer versions are known to work -- older versions are specifically known to **not** work!)
+- Bash ([homepage](https://www.gnu.org/software/bash/)) is required by the Makefiles that Vanagon generates
+- A ssh server ([homepage](https://www.openssh.com/)) is required by most engines
```
haus: A => An

README.md
```diff
@@ -28,23 +39,44 @@ creates a master makefile for the project, and configures, builds, and installs
...
----
-Vanagon is self-contained. A recent version of ruby (2.1 or greater) should be
-all that is required. Beyond that, ssh, rsync and git are also required on the
-host, and ssh-server and rsync is required on the target (package installation
-for the target can be customized in the platform config for the target).
+## Runtime Requirements
+
+Vanagon carries two sets of requirements: requirements for the local host where Vanagon is run, and for the remote target where compilation and packaging happens.
+
+Also, Vanagon ships with a number of engines which may include additional optional dependencies if you wish to use them. These engines are currently considered experimental, and receive less attention than the `Hardware` or `vmpooler` engines do. If you find a bug in these engines, [please open a ticket](https://tickets.puppetlabs.com/browse/PA) and let us know.
```
underscorgan: this should be tickets.puppetlabs.com/browse/VANAGON

underscorgan: welp, apparently that project is private. Should it be? Should we use CPR instead of PA for easier ticket routing?

shrug: That's my fault, it should get opened up shortly

mckern: Gonna leave it for now. We can fix it when it's public.

PR #460, opened 2017-03-01
(maint) Allow passing of custom build variables

underscorgan:
For some projects you may need to set custom variables/defines/etc in
your RPM spec file or debian rules file. This allows you to pass
arbitrary strings that will get added to those files.

TODO:
- [x]  Unit tests
- [x]  Make sure this doesn't break other builds
- [x]  ~~Add functionality for non deb/rpm platforms~~ OR make it fail if used on unsupported platforms

### DIFF
--- a/lib/vanagon/project/dsl.rb
+++ b/lib/vanagon/project/dsl.rb
@@ -262,8 +262,7 @@ class Vanagon
 
       def package_override(var)
         platform = @project.platform
-        fail "I don't know how to set package overrides for #{platform.name}, teach me?" unless platform.is_rpm? || platform.is_deb?
-        @project.package_overrides << var
+        platform.package_override(self._project, var)
       end
     end
   end
--- a/lib/vanagon/project/dsl.rb
+++ b/lib/vanagon/project/dsl.rb
@@ -259,6 +259,12 @@ class Vanagon
       def retry_count(retry_count)
         @project.retry_count = retry_count
       end
+
+      def package_override(var)
+        platform = @project.platform
+        fail "I don't know how to set package overrides for #{platform.name}, teach me?" unless platform.is_rpm? || platform.is_deb?
+        @project.package_overrides << var
+      end
     end
   end
 end
--- a/spec/lib/vanagon/project/dsl_spec.rb
+++ b/spec/lib/vanagon/project/dsl_spec.rb
@@ -1,6 +1,7 @@
 require 'vanagon/project/dsl'
 require 'vanagon/driver'
 require 'vanagon/common'
+require 'vanagon/platform'
 
 describe 'Vanagon::Project::DSL' do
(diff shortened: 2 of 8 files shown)


underscorgan:
Testing an agent build at http://jenkins-release.delivery.puppetlabs.net/job/vanagon_generic_job/716/

spec/lib/vanagon/project/dsl_spec.rb
```diff
@@ -231,6 +232,35 @@
     end
   end
 
+  describe "#custom_build_var" do
+    let(:project_block) {
+"project 'test-fixture' do |proj|
+  proj.custom_build_var \"TEST_VAR='foo'\"
+end"
```
mwaggett: do these lines want to be indented?

underscorgan: Nope

mckern: You *could* indent them using the `String#undent` method that Vanagon provides, if you wanted to.

mckern:
Broadly, I like the work. Not sold on the name `custom_build_vars` though. Gonna think that over fora few minutes.

spec/lib/vanagon/project/dsl_spec.rb
underscorgan: I was mostly just following how we were doing this elsewhere, but *shrug*. I can change it if you want.

underscorgan:
@mckern 👍  naming is hard, definitely open to suggestions though.

underscorgan:
@mckern any thoughts?

mckern:
@underscorgan oh right! Something like `package_overrides`, maybe? Since there's no class around them to enforce structure and items will be written into the templates that build packages without sanity checking, that's probably more in-line with what they actually do.

underscorgan:
@mckern 👍 , thanks!

underscorgan:
@mckern how about now?

lib/vanagon/project/dsl.rb
```diff
@@ -268,6 +268,12 @@ def retry_count(retry_count)
         @project.retry_count = retry_count
         environment 'VANAGON_PROJECT_RETRY_COUNT', @project.retry_count
       end
+
+      def package_override(var)
+        platform = @project.platform
+        fail "I don't know how to set package overrides for #{platform.name}, teach me?" unless platform.is_rpm? || platform.is_deb?
```
mckern: This is good enough for a rev. 1 but I'd lose the `unless` conditional here and move `@project.package_overrides << var` directly into any platforms that support it ASAP. Or do you think we're better served just doing it now?

underscorgan: I think doing it now is ok, this isn't really critical.

PR #463, opened 2017-03-14
(VANAGON-47) RIP mayfly metrics -- remove profiling 

mckern:
Thanks to the transitive properties of target-specific variables, this feature will never truly work as intended: anything that doesn't define `VANAGON_TARGET` explicitly will end up inheriting it from anything in its call stack that does, and that will pollute metrics. We can still do what we want to do by using alternative build systems (extensions to `make`, mostly`) but this code has to go. This has the net-positive of (probably) deprecating https://github.com/puppetlabs/puppet-agent/pull/1000, since it actually maintains backwards compatability with P-A's `master` branch.


mckern:
(Note, I think I forgot to update a couple of spec tests. Will double check and fix those up if I did.)

melissa:
This is simply backing out the component environment work, yes? But it still maintains the project/platform environment work that was done?

mckern:
> This is simply backing out the component environment work, yes? But it still maintains the project/platform environment work that was done?

It backs out **most** of the Component environment work. Components will still use Environments now, instead of raw Hashes, and will still render as Make variables if they require subshells for resolution. They just won't be target-specific variables in the rendered Makefile anymore.

melissa:
And the vanagon build you linked in hipchat was built with this branch?

mckern:
Yup!

PR #466, opened 2017-03-15
(maint) Add capability to figure out version from branch

underscorgan:
This also includes some reorganizing of Platform/Platform::DSL so we can use our handy version_from_branch to add versioned repos from our project definition!

TODO:
- [x] Add YARD docs
- [x] Add spec tests
- [x] Test building the world. - http://jenkins-release.delivery.puppetlabs.net/job/vanagon_generic_job/740/

### DIFF
--- a/lib/vanagon/project/dsl.rb
+++ b/lib/vanagon/project/dsl.rb
@@ -159,6 +159,23 @@ class Vanagon
         warn "Directory '#{dirname}' cannot be versioned by git. Maybe it hasn't been tagged yet?"
       end
 
+      # Get the version string from a git branch name. This will look for a '.'
+      # delimited string of numbers of any length and return that as the version.
+      # For example, 'maint/1.7.0/fixing-some-bugs' will return '1.7.0' and '4.8.x'
+      # will return '4.8'.
+      #
+      # @return version string parsed from branch name, fails if unable to find version
+      def version_from_branch
+        branch = Git.open(File.expand_path("..", Vanagon::Driver.configdir)).current_branch
+        if branch =~ /(\d+(\.\d+)+)/
+          return $1
+        else
+          fail "Can't find a version in your branch, make sure it matches <number>.<number>, like maint/1.7.0/fixing-some-bugs"
+        end
+      rescue Git::GitExecuteError => e
+        fail "Something went wrong trying to find your git branch.\n#{e}"
+      end
+
       # Sets the vendor for the project. Used in packaging artifacts.
       #
       # @param vend [String] vendor or author of the project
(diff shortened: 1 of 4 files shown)


lib/vanagon/project/dsl.rb
```diff
@@ -159,6 +159,17 @@ def version_from_git
         warn "Directory '#{dirname}' cannot be versioned by git. Maybe it hasn't been tagged yet?"
       end
 
+      def version_from_branch
+        branch = Git.open(File.expand_path("..", Vanagon::Driver.configdir)).current_branch
+        if branch =~ /(\d+(\.\d+)+)/
+          return $1
+        else
+          warn "Can't find a version in your branch, make sure it matches \d+(\.\d+)+"
```
mckern: Nitpick: instead of describing it as a regex, use a plain description/example of acceptable formats?

lib/vanagon/project/dsl.rb
```diff
@@ -159,6 +159,17 @@ def version_from_git
...
+        end
+      rescue Git::GitExecuteError => e
+        warn "Something went wrong trying to find your git branch.\n#{e}"
```
mckern: Should this be a warning, or should this be a failure? Without this, won't artifacts become unversioned?

lib/vanagon/project/dsl.rb
```diff
@@ -159,6 +159,17 @@ def version_from_git
         warn "Directory '#{dirname}' cannot be versioned by git. Maybe it hasn't been tagged yet?"
       end
 
+      def version_from_branch
+        branch = Git.open(File.expand_path("..", Vanagon::Driver.configdir)).current_branch
+        if branch =~ /(\d+(\.\d+)+)/
+          return $1
+        else
+          warn "Can't find a version in your branch, make sure it matches <number>.<number>, like maint/1.7.0/fixing-some-bugs"
```
underscorgan: @mckern this better?

lib/vanagon/project/dsl.rb
```diff
@@ -159,6 +159,17 @@ def version_from_git
...
+        end
+      rescue Git::GitExecuteError => e
+        warn "Something went wrong trying to find your git branch.\n#{e}"
```
underscorgan: @mckern hm, not sure. I was mostly modeling off of `version_from_git` which only warns, but that means you need to have extra protections as someone using the method.

lib/vanagon/project/dsl.rb
mckern: Yes. The test should have good branch names and bad branch names and validate that it doesn't crap out with unexpectedly similar names (thanks to the magic of regex).

lib/vanagon/project/dsl.rb
mckern: Hrm. OK.

underscorgan: Eh, I've had a change of heart. I'm going to make these fail. If you're trying to get a version from git we should expect that to work.

PR #468, opened 2017-03-24
(maint) Ensure dirname override will be honored

melissa:
We allow the user to set dirname in case they are dealing with a source
that unpacks into a directory that's different than the name of the
archive. However, we don't honor this. The code always overwrites that
user-set variable with what it thinks it should be. That's lame.

This commit fixes how that variable is treated. If the user does in fact
supply `dirname` in the component config, then that variable assignment
will be honored. Otherwise, it defaults to use the same name as the
original archive.

### DIFF
--- a/lib/vanagon/component.rb
+++ b/lib/vanagon/component.rb
@@ -204,7 +204,7 @@ class Vanagon
         extract_with << source.extract(platform.tar) if source.respond_to? :extract
 
         @cleanup_source = source.cleanup if source.respond_to?(:cleanup)
-        @dirname = source.dirname
+        @dirname ||= source.dirname
 
         # Git based sources probably won't set the version, so we load it if it hasn't been already set
         if source.respond_to?(:version)
--- a/lib/vanagon/component/dsl.rb
+++ b/lib/vanagon/component/dsl.rb
@@ -315,6 +315,29 @@ class Vanagon
         end
       end
 
+      # Set a source dir
+      #
+      # The build dir will be created when the source archive is unpacked. This
+      # should be used when the unpacked directory name does not match the
+      # source archive name.
+      #
+      # @example
+      #   pkg.url "http://the-internet.com/a-silly-name-that-unpacks-into-not-this.tar.gz"
+      #   pkg.dirname "really-cool-directory"
+      #   pkg.configure { ["cmake .."] }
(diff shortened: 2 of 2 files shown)


lib/vanagon/component/dsl.rb
```diff
@@ -315,6 +315,29 @@ def build_dir(path)
...
+      #   pkg.install { ["make install"] }
+      #
+      # @param path [String] The build directory to use for building the project
+      def dirname(path)
+        if Pathname.new(path).relative?
+          @component.dirname = path
+        else
+          raise Vanagon::Error, "dirname should be a relative path, but '#{path}' looks to be absolute."
+        end
+      end
```
melissa: I know `dirname` should have an automagically generated setter, but when this method wasn't there, it barfed on me. Sooooo, I included it. We probably also don't need this checking. I just copied it from `build_dir`, but it probably doesn't hurt since it should be a relative path, not absolute.

lib/vanagon/component.rb
```diff
@@ -204,7 +204,7 @@ def get_source(workdir) # rubocop:disable Metrics/AbcSize
         extract_with << source.extract(platform.tar) if source.respond_to? :extract
 
         @cleanup_source = source.cleanup if source.respond_to?(:cleanup)
-        @dirname = source.dirname
+        @dirname ||= source.dirname
```
melissa: Well, this is to set `@dirname` if it isn't set by the user. So if it's already populated, then leave it as is. Otherwise, use the dirname provided by the source. I'm not sure the best way to do that other than the `||=`?

lib/vanagon/component/dsl.rb
mckern: I like explicit accessors if there's work to be done during the getting or setting. No concerns there.

lib/vanagon/component.rb
mckern: The naive answer is "move it into its own getter":

```ruby
def dirname
  @dirname ||= source.dirname
end
```

But thinking that through, it looks like source initialization is wonky and brittle. I planned to unwind this soon for the `render` command so maybe I'll fast track that. Your way probably works for now but I think that a lot of `get_source()` is going to be broken up and spread around soon.

PR #469, opened 2017-03-25
(MAINT) Git tests shouldn't hit the network

mckern:
Since we outsource git repo handling to a 3rd party library, we don't have to test every single aspect of our wrapper. Just ensuring that it talks to the underlying Git library should be sufficient.

The benefit is that unit test run time drops from about 7 seconds to consistently less than 1 second. A win is a win, right?

### DIFF
--- a/spec/lib/vanagon/component/source/git_spec.rb
+++ b/spec/lib/vanagon/component/source/git_spec.rb
@@ -1,32 +1,45 @@
 require 'vanagon/component/source/git'
 
 describe "Vanagon::Component::Source::Git" do
-  let(:klass) { Vanagon::Component::Source::Git }
-  let(:url) { 'git://github.com/puppetlabs/facter' }
-  let(:ref_tag) { 'refs/tags/2.2.0' }
-  let(:bad_sha) { 'FEEDBEEF' }
-  let(:workdir) { ENV["TMPDIR"] || '/tmp' }
+  # before(:all) blocks are run once before all of the examples in a group
+  before :all do
+    @klass = Vanagon::Component::Source::Git
+    # This repo will not be cloned over the network
+    @url = 'git://github.com/puppetlabs/facter.git'
+    # This path will not be created on disk
+    @local_url = "file://#{Dir.tmpdir}/puppet-agent"
+    @ref_tag = 'refs/tags/2.2.0'
+    @workdir = nil
+  end
 
-  after(:each) { %x(rm -rf #{workdir}/facter) }
+  # before(:each) blocks are run before each example
+  before :each do
+    allow(Git)
+      .to receive(:ls_remote)
+            .and_return(true)
+  end
 
   describe "#initialize" do
     it "raises error on initialization with an invalid repo" do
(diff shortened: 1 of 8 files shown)


mckern:
In Travis, the test run time drops from ~2:30 to about ~0:37 (on this PR at least).

spec/lib/vanagon/component/source/git_spec.rb
```diff
@@ -1,49 +1,53 @@
...
         .to eq('HEAD')
     end
   end
-
-  describe "#fetch" do
-    it "raises an error on checkout failure with a bad SHA" do
-      expect { klass.new("#{url}", ref: bad_sha, workdir: workdir).fetch }
-        .to raise_error Vanagon::CheckoutFailed
-    end
-  end
```
mwaggett: Why remove this test?

spec/lib/vanagon/component/source/git_spec.rb
```diff
@@ -1,49 +1,53 @@
...
 
   describe "#dirname" do
     it "returns the name of the repo" do
-      git_source = klass.new(url, ref: ref_tag, workdir: workdir)
+      git_source = @klass.new(@url, ref: @ref_tag, workdir: @workdir)
       expect(git_source.dirname)
         .to eq('facter')
     end
 
     it "returns the name of the repo and strips .git" do
```
melissa: what's the difference between the functionality this test and the test on line 33 are exercising?

spec/lib/vanagon/component/source/git_spec.rb
mckern: This would effectively be caught by the same wrapper as `"#initialize raises error on initialization with an invalid repo"` since this is all being handled by the underlying Git library now. When Vanagon provided all of its own Git implementation, this was needed but since it's outsourced to a 3rd party library we should only be testing our connectivity to the library and any functionality we provide on top of the library. So there's room for new tests, but redundant old tests should go.

mwaggett: mmkay, well in that case, you can delete `@bad_sha` on line 9, since it isn't used anywhere anymore.

spec/lib/vanagon/component/source/git_spec.rb
mckern: Which test (by name) are you referring to? Line 33 is inside this test (`"#dirname returns the name of the repo"`).

melissa: `it "returns the name of the repo" do` looks like it does exactly the same this as `it "returns the name of the repo and strips .git" do`, just with a different title

mckern: Ah, that's my fault. The second test used to mangle the repo URL (which did not include `.git`) to include `.git`. That makes sense, since the implementation in `Vanagon::Component::Source::Git` is:

```ruby
# The dirname to reference when building from the repo
#
# @return [String] the directory where the repo was cloned
def dirname
  File.basename(url.path, ".git")
end
```

PR #473, opened 2017-04-07
(VANAGON-56) Introducing Mirror URLs

mckern:
The rewrite engine in Vanagon is not popular. It's made hitting
internal mirrors easier for Puppet but it's made building Puppet Agent
harder for many of our customers. I've been planning on adding Mirror
URL support for a while now... and that day is today.

Mirror URLs will deprecate rewrite rules, and provide flexibility for
multiple mirrors (if so desired). There's a new DSL method, and some
backend code, and deprecated-but-preserved backwards compatibility for
any projects that aren't likely to be able to move forward quickly.

Checklist:

- [X] Method deprecation warnings
- [X] Tests that validate that we preserved the old methods
- [X] Tests that validate the new behavior
- [x] Updated examples

### DIFF
--- a/lib/vanagon/component.rb
+++ b/lib/vanagon/component.rb
@@ -1,6 +1,7 @@
-require 'vanagon/component/source'
 require 'vanagon/component/dsl'
 require 'vanagon/component/rules'
+require 'vanagon/component/source'
+require 'vanagon/component/source/rewrite'
 
 class Vanagon
   class Component
@@ -18,6 +19,7 @@ class Vanagon
     attr_accessor :sources
     attr_accessor :patches
     attr_accessor :url
+    attr_accessor :mirrors
     attr_accessor :license
 
     # holds an OpenStruct describing all of the particular details about
@@ -190,16 +192,74 @@ class Vanagon
       @files.select(&:configfile?)
     end
 
+    # @return [Set] a list of unique mirror URIs that should be used to
+    #   retrieve the upstream source before attempting to retrieve from
+    #   whatever URI was defined for #uri.
+    def mirrors
+      @mirrors ||= Set.new [deprecated_rewrite_url].compact
+    end
+
+    # Retrieve upstream source file from a mirror, by randomly iterating
+    # through #mirrors until there's no more mirrors left. Will #warn
+    # if the mirror's URI cannot be resolved or if the URI cannot
+    # be retrieved. Does not suppress any errors from
+    # Vanagon::Component::Source.
+    #
+    # @return [Boolean] return True if the source can be retrieved,
+    #   or False otherwise. This is because because each subclass of
+    #   Vanagon::Component::Source returns an inconsistent value
+    #   if #fetch is successful.
+    def fetch_mirrors(options)
+      mirrors.to_a.shuffle.each do |mirror|
+        begin
+          $stderr.puts %(Attempting to fetch from mirror URL "#{mirror}")
+          @source = Vanagon::Component::Source.source(mirror, options)
+          return true if source.fetch
+        rescue SocketError
+          # SocketError means that there was no DNS/name resolution
+          # for whatever remote protocol the mirror tried to use.
+          warn %(Unable to resolve mirror URL "#{mirror}")
+        rescue RuntimeError
+          # Source retrieval does not consistently return a meaningful
+          # namespaced error message, which means we're brute-force rescuing
+          # RuntimeError. Not a good look, and we should fix this.
+          warn %(Unable to retrieve mirror URL "#{mirror}")
+        end
+      end
+      false
+    end
+
+    # Retrieve upstream source file from the canonical URL.
+    # Does not suppress any errors from Vanagon::Component::Source.
+    #
+    # @return [Boolean] return True if the source can be retrieved,
+    #   or False otherwise
+    def fetch_url(options)
+      $stderr.puts %(Attempting to fetch from canonical URL "#{url}")
+      @source = Vanagon::Component::Source.source(url, options)
+      # Explicitly coerce the return value of #source.fetch,
+      # because each subclass of Vanagon::Component::Source returns
+      # an inconsistent value if #fetch is successful.
+      !!source.fetch
+    end
+
+    # This method is deprecated and private. It will return
+    # a rewritten URL if there's any value for #url.
+    def deprecated_rewrite_url
+      return nil unless url
+      Vanagon::Component::Source::Rewrite.parse_and_rewrite(url)
+    end
+    private :deprecated_rewrite_url
+
     # Fetches the primary source for the component. As a side effect, also sets
     # @extract_with, @dirname and @version for the component for use in the
     # makefile template
     #
     # @param workdir [String] working directory to put the source into
-    def get_source(workdir) # rubocop:disable Metrics/AbcSize
+    def get_source(workdir) # rubocop:disable Metrics/AbcSize, Metrics/PerceivedComplexity
       opts = options.merge({ workdir: workdir })
-      if url
-        @source = Vanagon::Component::Source.source(url, opts)
-        source.fetch
+      if url || !mirrors.empty?
+        fetch_mirrors(opts) || fetch_url(opts)
         source.verify
         extract_with << source.extract(platform.tar) if source.respond_to? :extract
 
--- a/lib/vanagon/component/dsl.rb
+++ b/lib/vanagon/component/dsl.rb
@@ -276,11 +276,20 @@ class Vanagon
         @component.version = ver
       end
 
-      # Sets the url for the source of this component
+      # Sets the canonical URL or URI for the upstream source of this component
       #
-      # @param the_url [String] the url to the source for this component
-      def url(the_url)
-        @component.url = the_url
+      # @param uri [String, URI] a URL or URI describing a canonical location
+      #   for a component's source code or artifact
+      def url(uri)
+        @component.url = uri.to_s
+      end
+
+      # Sets a mirror url for the source of this component
+      #
+      # @param url [String] a mirror url to use as the source for this component.
+      #   Can be called more than once to add multiple mirror URLs.
+      def mirror(url)
+        @component.mirrors << url
       end
 
       def sum(value)
--- a/lib/vanagon/component/source/rewrite.rb
+++ b/lib/vanagon/component/source/rewrite.rb
@@ -1,0 +1,85 @@
+require 'vanagon/component/source'
+
+class Vanagon
+  class Component
+    class Source
+      # This class has been extracted from Vanagon::Component::Source for the
+      # sake of isolation and in service of its pending removal. Rewrite rules
+      # should be considered deprecated. The removal will be carried out before
+      # Vanagon 1.0.0 is released.
+      class Rewrite
+        @rewrite_rules = {}
+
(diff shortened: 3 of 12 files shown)


mckern:
This should not be merged until VANAGON-55 is resolved so we can cut a Z release to fix that. Merging this will be reason enough to cut a new Y release, no questions asked.

lib/vanagon/component.rb
```diff
@@ -190,16 +192,74 @@ def configfiles
...
+    # if the mirror's URI cannot be resolved or if the URI cannot
+    # be retrieved. Does not suppress any errors from
+    # Vanagon::Component::Source.
+    #
+    # @return [Boolean] return True if the source can be retrieved,
+    #   or False otherwise. This is because because each subclass of
+    #   Vanagon::Component::Source returns an inconsistent value
+    #   if #fetch is successful.
+    def fetch_mirrors(options)
+      mirrors.to_a.shuffle.each do |mirror|
```
mwaggett: What's the purpose of shuffling?

lib/vanagon/component/dsl.rb
```diff
@@ -276,11 +276,20 @@ def version(ver)
         @component.version = ver
       end
 
-      # Sets the url for the source of this component
+      # Sets the canonical URL or URI for the upstream source of this component
       #
-      # @param the_url [String] the url to the source for this component
-      def url(the_url)
-        @component.url = the_url
+      # @param addr [String, URI] a URL or URI describing a canonical location
```
mwaggett: This param name doesn't match the one in the function below.

spec/lib/vanagon/component_spec.rb
```diff
@@ -44,6 +44,55 @@
     end
   end
 
+  describe "#get_source" do
+    before :each do
+      @workdir = Dir.mktmpdir
+      @fake_tar = "file://spec/fixtures/files/#{@fake_dir}.tar.gz"
```
mwaggett: Do you mean `@workdir` instead of `@fake_dir`?

lib/vanagon/component.rb
mckern: Mirrors have no precedence, and hitting one should be equivilent to hitting any of them. So using `Array#shuffle` ensures that the order is randomized, but that the entire array is still worked through.

mwaggett: You don't think it makes sense to try mirrors in a certain order?

mckern: Nope. It introduces unexpected coupling.

mckern: I'll try to explain it here instead of IRL, because it occurs to me that "nope" is too glib an answer to be useful. I'm basing my position here on my experiences with load balancing infrastructures. A round-robin approach (which shuffling an array basically is) is a reliable way to ensure that you're not accidentally pinning to one server while the rest of them rot and fall over. By shuffling the array ourselves, the hope is to mitigate the risk of us accidentally pinning to one URL that works because we uploaded stuff to the right place but none of our other URLs work because we goofed.

spec/lib/vanagon/component_spec.rb
mckern: Yeah, leftover copypasta from previous tests. Will push up a fix soon.

lib/vanagon/component.rb
mwaggett: okay, that makes sense.

stahnma:
👍

branan:
:see_no_evil:

lib/vanagon/component/source/rewrite.rb
```diff
@@ -0,0 +1,85 @@
...
+      # sake of isolation and in service of its pending removal. Rewrite rules
+      # should be considered deprecated. The removal will be carried out before
+      # Vanagon 1.0.0 is released.
+      class Rewrite
+        @rewrite_rules = {}
+
+        class << self
+          attr_reader :rewrite_rules
+
+          # @deprecated Please use the component DSL method #mirror(<URI>)
```
melissa: Can you remind me what this does? It only spits this out in the docs, right? Users aren't going to get a deprecation method when they actually run the code?

mckern: correct, though a deprecation notice will be emitted by some of these methods now.

lib/vanagon/component.rb
```diff
@@ -190,16 +192,74 @@ def configfiles
...
+        begin
+          puts %(Attempting to fetch from mirror URL "#{mirror}")
```
mckern: I should probably update all of the uses of `puts` here to be inline with #474

melissa: yes, I'd agree with that

melissa:
I don't see anything that sticks out as a concern. I'd like to see this in action first, and I'd like to see the update to the uses of `puts` as well

melissa:
Question:

Rust has an update rule in `configs/projects/pl-rust.rb`
```
proj.register_rewrite_rule 'http', 'http://buildsources.delivery.puppetlabs.net'
```

which impacts a whole bunch of urls defined in `configs/components/rust.rb`. When we update this component config to use the mirror rules rather than the rewrite, we'll have to define multiple urls/mirrors for each thing, right?

This is the original
```
component 'rust' do |pkg, settings, platform|
  pkg.version '1.16.0'
  if platform.is_macos?
    pkg.md5sum 'a47f24355bf2165c632d4263031ac231'
    pkg.url "https://static.rust-lang.org/dist/rust-#{pkg.get_version}-x86_64-apple-darwin.tar.gz"
  elsif platform.is_windows?
    pkg.environment('CYGWIN', settings[:cygwin])

    if platform.architecture == 'x64'
      pkg.md5sum 'e938a4aeb6ff4e8a99be7eea82be3cc8'
      pkg.url "https://static.rust-lang.org/dist/rust-#{pkg.get_version}-x86_64-pc-windows-gnu.tar.gz"
    else
      pkg.md5sum '69d9f7d47bb961e773b23be829c1b0fc'
      pkg.url "https://static.rust-lang.org/dist/rust-#{pkg.get_version}-i686-pc-windows-gnu.tar.gz"
    end
  elsif platform.is_linux?
    if platform.architecture =~ /x86_64|amd64/
      pkg.md5sum '5effa7068c26151029432466235089a9'
      pkg.url "https://static.rust-lang.org/dist/rust-#{pkg.get_version}-x86_64-unknown-linux-gnu.tar.gz"
    elsif platform.architecture == 'i386'
      pkg.md5sum '52393f468be6797d8bf66bbc51d5ad6e'
      pkg.url "https://static.rust-lang.org/dist/rust-#{pkg.get_version}-i686-unknown-linux-gnu.tar.gz"
    end
  end

  pkg.install do
    if platform.is_windows?
      # install.sh fails when doing a final check because of $(uname -s)
      [ "./install.sh --prefix=#{settings[:prefix]} --without=rust-docs || true",
        "chmod -R 755 $(cygpath -u #{settings[:prefix]})/lib"
      ]
    else
      [ "./install.sh --prefix=#{settings[:prefix]} --without=rust-docs" ]
    end
  end

end
```

I assume that with this new paradigm, it'll look something like this?
```
component 'rust' do |pkg, settings, platform|
  pkg.version '1.16.0'
  if platform.is_macos?
    pkg.md5sum 'a47f24355bf2165c632d4263031ac231'
    pkg.url "http://buildsources.delivery.puppetlabs.net/rust-#{pkg.get_version}-x86_64-apple-darwin.tar.gz"
    pkg.url "https://static.rust-lang.org/dist/rust-#{pkg.get_version}-x86_64-apple-darwin.tar.gz"
  elsif platform.is_windows?
    pkg.environment('CYGWIN', settings[:cygwin])

    if platform.architecture == 'x64'
      pkg.md5sum 'e938a4aeb6ff4e8a99be7eea82be3cc8'
      pkg.url "http://buildsources.delivery.puppetlabs.net/rust-#{pkg.get_version}-x86_64-pc-windows-gnu.tar.gz"
      pkg.url "https://static.rust-lang.org/dist/rust-#{pkg.get_version}-x86_64-pc-windows-gnu.tar.gz"
    else
      pkg.md5sum '69d9f7d47bb961e773b23be829c1b0fc'
      pkg.url "http://buildsources.delivery.puppetlabs.net/rust-#{pkg.get_version}-i686-pc-windows-gnu.tar.gz"
      pkg.mirror "https://static.rust-lang.org/dist/rust-#{pkg.get_version}-i686-pc-windows-gnu.tar.gz"
    end
  elsif platform.is_linux?
    if platform.architecture =~ /x86_64|amd64/
      pkg.md5sum '5effa7068c26151029432466235089a9'
      pkg.url "http://buildsources.delivery.puppetlabs.net/rust-#{pkg.get_version}-x86_64-unknown-linux-gnu.tar.gz"
      pkg.mirror "https://static.rust-lang.org/dist/rust-#{pkg.get_version}-x86_64-unknown-linux-gnu.tar.gz"
    elsif platform.architecture == 'i386'
      pkg.md5sum '52393f468be6797d8bf66bbc51d5ad6e'
      pkg.url "http://buildsources.delivery.puppetlabs.net/rust-#{pkg.get_version}-i686-unknown-linux-gnu.tar.gz"
      pkg.mirror "https://static.rust-lang.org/dist/rust-#{pkg.get_version}-i686-unknown-linux-gnu.tar.gz"
    end
  end

  pkg.install do
    if platform.is_windows?
      # install.sh fails when doing a final check because of $(uname -s)
      [ "./install.sh --prefix=#{settings[:prefix]} --without=rust-docs || true",
        "chmod -R 755 $(cygpath -u #{settings[:prefix]})/lib"
      ]
    else
      [ "./install.sh --prefix=#{settings[:prefix]} --without=rust-docs" ]
    end
  end

end
```

Does it matter if the extras are set to `pkg.url` or `pkg.mirror`? I see you have both in your example components

underscorgan:
What have you built with this? I'd like to see some tests in the dynamic_vanagon_builder of at least agent and something that isn't agent.

I like the change though.

mckern:
@underscorgan 

```
mckern@flexo puppet-agent (git:1.10.0-release) $ VANAGON_SSH_KEY=~/.ssh/jenkins_rsa bundle exec build puppet-agent debian-8-armhf

✂-✂-✂-✂

cp /var/tmp/tmp.6kQ6L1WLa4/*.deb ./output/deb/jessie/PC1
Warning: Permanently added 's1ku07d718gd3nh.delivery.puppetlabs.net,10.32.112.228' (ECDSA) to the list of known hosts.
s1ku07d718gd3nh.delivery.puppetlabs.net has been destroyed
mckern@flexo puppet-agent (git:1.10.0-release) $
```

mckern:
(Gotta fix `$stdout` and `$stderr` usage)

mckern:
@melissa you actually found a typo. `#url` should be called once and only once. `#mirror` can be called any number of times. So that means that components will have to be updated, but I think that's preferable since the URLs are a property of components (not projects).

melissa:
Oh, awesome! That makes a lot more sense :) So, question still stands, the above example with rust is correct, but with the additional `pkg.url` as `pkg.mirror`?

mckern:
@melissa Yes. Rust is a weird special case because it's a precompiled, platform specific artifact. Vanagon was only intended/designed to build from source, not consume static binary blobs, so the Rust situation is a pretty clear outlier.

PR #475, opened 2017-04-07
(VANAGON-55) Fix AIX RPMs requirements

mckern:
AIX doesn't track the entire filesystem in the RPM database, so
there's a number of core system files that are not exposed to RPM and
should not be used as a constraint for dependencies. `/bin/touch` and
`/bin/mkdir` are two of them. We should stop requiring those files as
dependencies for AIX RPMs.

### DIFF
--- a/resources/rpm/project.spec.erb
+++ b/resources/rpm/project.spec.erb
@@ -58,10 +58,7 @@ Requires:  <%= requires %>
 # did not specify a dependency on these.
 # In the future, we will supress pre/post scripts completely if there's nothing
 # specified by the project or the components.
-<%- if @platform.is_aix? -%>
-Requires: /bin/mkdir
-Requires: /bin/touch
-<%- else -%>
+<%- unless @platform.is_aix? -%>
 Requires(pre): /bin/mkdir
 Requires(pre): /bin/touch
 Requires(post): /bin/mkdir


resources/rpm/project.spec.erb
```diff
@@ -58,10 +58,7 @@ Requires:  <%= requires %>
 # did not specify a dependency on these.
 # In the future, we will supress pre/post scripts completely if there's nothing
 # specified by the project or the components.
-<%- if @platform.is_aix? -%>
-Requires: /bin/mkdir
-Requires: /bin/touch
-<%- else -%>
+<%- unless @platform.is_aix? -%>
```
underscorgan: Is there a different package we should be depending on for aix or can we just assume we have `mkdir` and `touch`

shrug: I *think* those are provided by the base OS install and we shouldn't need to worry about them.

shrug: complicated by the fact that the RPM database knows nothing about system files or anything installed by a native AIX package

mwaggett:
Has this been tested in any way?

resources/rpm/project.spec.erb
underscorgan: wheeeeeeee

melissa:
Is it breaking things to leave this as is? What inspired you to dig into this?

McdonaldSeanp:
@melissa yes, it's breaking things. Currently builds on AIX are failing when using vanagon 0.11 because of this.

McdonaldSeanp:
👍

mckern:
@melissa and @underscorgan from the Agent hipchat room:

> AIX bolts RPM on as an extra package manager, like homebrew on macOS. subsequently, not all files available on an AIX platform are tracked in the rpm DB. that doesn't mean files are not available on a system, just that they were not installed by RPM so they are not tracked by RPM. when an RPM package expresses a dependency, RPM does not check to see if a file actually exists on a system -- it just looks it up in the RPM DB and determines if something provides it, and if it's already installed.
>
> AIX is a classic UNIX, which means that was you think of as "coreutils" (and which would likely be provided by a coreutils package in a linux distro) is just a standard part of the base OS and therefore not tracked by a package manager at all. Solaris, HPUX, and FreeBSD all behave identically there. so, while yes, we are going to need the files `/bin/touch` and /bin/mkdir` available for the pre/post scripts to work, on AIX you do not need to express this dependency (and in fact SHOULD NOT) because those files are absolutely going to be in place no matter what (as long as a user has not done some bullshit like `sudo rm /bin/*` and then refused to quit in shame for their mistake).

PR #476, opened 2017-04-10
(maint) Fix for workdir with symlinks

underscorgan:
Fun fact: When building on OSX your tmpdir is going to end up in /var
which is actually a symlink to /private/var. If you don't chdir / set
your workdir to the realpath of your tmpdir git will fail with

fatal: Could not chdir to
'../../../../../../../../../../../private/var/folders...'

if you are using the `--git-dir` option. The ruby-git library we use
appends `--git-dir` to each git command. So, when initializing the git
objects we need to make sure we're using the `File.realpath` so we don't
trigger this weird issue when trying to update submodules.

Big thanks to @mckern for helping me track this issue down.

### DIFF
--- a/lib/vanagon/component/source/git.rb
+++ b/lib/vanagon/component/source/git.rb
@@ -45,7 +45,7 @@ class Vanagon
           # Ensure that #url returns a URI object
           @url = URI.parse(url.to_s)
           @ref = opts[:ref]
-          @workdir = workdir
+          @workdir = File.realpath(workdir)
 
           # We can test for Repo existence without cloning
           raise Vanagon::InvalidRepo, "#{url} not a valid Git repo" unless valid_remote?
--- a/spec/lib/vanagon/component/source/git_spec.rb
+++ b/spec/lib/vanagon/component/source/git_spec.rb
@@ -17,6 +17,8 @@ describe "Vanagon::Component::Source::Git" do
     allow(Git)
       .to receive(:ls_remote)
             .and_return(true)
+
+    allow(File).to receive(:realpath).and_return(@workdir)
   end
 
   describe "#initialize" do
@@ -28,6 +30,13 @@ describe "Vanagon::Component::Source::Git" do
 
       expect { @klass.new(@url, ref: @ref_tag, workdir: @workdir) }
         .to raise_error(Vanagon::InvalidRepo)
+    end
+
+    it "uses the realpath of the workdir if we're in a symlinked dir" do
+      expect(File).to receive(:realpath).and_return("/tmp/bar")
+      git_source = @klass.new(@local_url, ref: @ref_tag, workdir: "/tmp/foo")
+      expect(git_source.workdir)
+        .to eq('/tmp/bar')
     end
   end
 


lib/vanagon/component/source/git.rb
```diff
@@ -45,7 +45,7 @@ def initialize(url, workdir:, **options)
           # Ensure that #url returns a URI object
           @url = URI.parse(url.to_s)
           @ref = opts[:ref]
-          @workdir = workdir
+          @workdir = File.realpath(workdir) if workdir
```
shrug: Can workdir be false?

underscorgan: I don't think so. Everywhere this code is called requires a workdir to be passed, I was mostly running into issues with unit tests where workdir was set to `nil`. I should probably just update the tests to have an actual workdir rather than just setting workdir to nil.

underscorgan: Yeah, traced through the codepaths and this will always be `@workdir` from the vanagon driver. Updating!

spec/lib/vanagon/component/source/git_spec.rb
```diff
@@ -17,6 +17,8 @@
     allow(Git)
       .to receive(:ls_remote)
             .and_return(true)
+
+    allow(File).to receive(:realpath).and_return(@workdir)
```
mckern: Did you strictly need this stub?

mckern: Yes, yes we do.

```
  4) Vanagon::Component::Source::Git #ref returns a default value of HEAD when no explicit Git reference is provided
     Failure/Error: @workdir = File.realpath(workdir)

     TypeError:
       no implicit conversion of nil into String
     # ./lib/vanagon/component/source/git.rb:48:in `realpath'
     # ./lib/vanagon/component/source/git.rb:48:in `initialize'
     # ./spec/lib/vanagon/component/source/git_spec.rb:59:in `new'
     # ./spec/lib/vanagon/component/source/git_spec.rb:59:in `block (3 levels) in <top (required)>'
```

spec/lib/vanagon/component/source/git_spec.rb
```diff
@@ -29,6 +31,13 @@
       expect { @klass.new(@url, ref: @ref_tag, workdir: @workdir) }
         .to raise_error(Vanagon::InvalidRepo)
     end
+
+    it "uses the realpath of the workdir if we're in a symlinked dir" do
+      expect(File).to receive(:realpath).and_return("/tmp/bar")
```
mckern: I know we're doing a bit of dodgy path hardcoding here, but is it possible to write this test without hardcoding assumptions about the availability of "/tmp"? I don't have any quick answers here but I might in the morning?

mckern:
Nothing but nitpicks from me. This is a good solution. If I can't think of a way to get the hardcoded `/tmp` paths out of that one unit test, I'll approve and merge.

spec/lib/vanagon/component/source/git_spec.rb
underscorgan: I was mostly using that as a miscellaneous string to make sure that if realpath('something') != 'something' the workdir gets set to realpath('something')

¯\_(ツ)_/¯

PR #481, opened 2017-05-09
(maint) Add ability to generate source artifacts for rpm/deb

underscorgan:
Sometimes it's useful to be able to generate .src.rpm and
.dsc/changes/etc. This adds the ability to do that.

### DIFF
--- a/lib/vanagon/platform.rb
+++ b/lib/vanagon/platform.rb
@@ -60,6 +60,9 @@ class Vanagon
     # Stores the local path where retrieved artifacts will be saved.
     attr_accessor :output_dir
 
+    # Stores the local path where source artifacts will be saved.
+    attr_accessor :source_output_dir
+
     # Username to use when connecting to a build target
     attr_accessor :target_user
 
@@ -237,6 +240,17 @@ class Vanagon
     # @return [String] relative path to where packages should be output to
     def output_dir(target_repo = "")
       @output_dir ||= File.join(@os_name, @os_version, target_repo, @architecture)
+    end
+
+    # Get the source dir for packages. Don't change it if it was already defined
+    # by the platform config. Defaults to output_dir unless specified otherwise
+    # (RPM specifies this)
+    #
+    # @param target_repo [String] optional repo target for built source packages
+    # defined at the project level
+    # @return [String] relative path to where source packages should be output to
+    def source_output_dir(target_repo = "")
+      @source_output_dir ||= output_dir(target_repo)
     end
 
     # Get the value of @dist, or derive it from the value of @os_name and @os_version.
--- a/lib/vanagon/platform/rpm.rb
+++ b/lib/vanagon/platform/rpm.rb
@@ -8,15 +8,24 @@ class Vanagon
       #
       # @param project [Vanagon::Project] project to build an rpm package of
       # @return [Array] list of commands required to build an rpm package for the given project from a tarball
-      def generate_package(project)
+      def generate_package(project) # rubocop:disable Metrics/AbcSize
         target_dir = project.repo ? output_dir(project.repo) : output_dir
+        target_source_output_dir = project.repo ? source_output_dir(project.repo) : source_output_dir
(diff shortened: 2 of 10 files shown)


underscorgan:
Still need to add docs and tests.

lib/vanagon/platform.rb
```diff
@@ -60,6 +60,9 @@ class Platform
     # Stores the local path where retrieved artifacts will be saved.
     attr_accessor :output_dir
 
+    # Stores the local path where source artifacts will be saved.
+    attr_accessor :source_dir
```
shrug: Hmm.... I kinda think this is a little bit confusing, too easy to mix up with build sources. maybe source_output_dir? or something a bit more descriptive?

underscorgan: ah, good point. will fix

lib/vanagon/platform/rpm.rb
```diff
@@ -40,6 +49,13 @@ def output_dir(target_repo = "products")
         super
       end
 
+      # Method to derive the directory for source artifacts
+      #
+      # @param target_repo [String] repo the source artifacts are targeting
+      def source_dir(target_repo = "products")
+        @source_dir ||= File.join(@os_name, @os_version, target_repo, 'SRPMS')
```
mwaggett: Do we want `target_repo` to come first, since we're changing paths?

lib/vanagon/platform/rpm.rb
```diff
@@ -40,6 +49,13 @@ def output_dir(target_repo = "products")
         super
       end
 
+      # Method to derive the directory for source artifacts
+      #
+      # @param target_repo [String] repo the source artifacts are targeting
+      def source_output_dir(target_repo = "products")
+        @source_output_dir ||= File.join(@os_name, @os_version, target_repo, 'SRPMS')
```
mwaggett: Do we want `target_repo` first, since we're changing paths?

lib/vanagon/platform/rpm.rb
underscorgan: No, this is happening before those vanagon changes. This should get updated when those changes go in.

lib/vanagon/platform/rpm.rb
underscorgan: Nope, those changes haven't been made yet, this will need to get updated when that happens.

mwaggett: I'm just literally working on those changes now.

underscorgan: But I need this to build sources for things that haven't moved yet. This is going to need to go in first.

mwaggett: kk

underscorgan: how are those changes going to know whether or not the order should change? not everything is moving

mwaggett: I'm gonna check if the target repo is PC1.

underscorgan: that's going to break PE things

mwaggett: What do those paths look like?

underscorgan: most of those don't have a target repo set and for RPMs that defaults to target repo of 'products' (so, like http://builds.puppetlabs.lan/puppet-enterprise-vanagon/pe-activemq-2016.4.5.14.3-3/artifacts/el/7/products/x86_64/) and debs defaults to no target repo (like http://builds.puppetlabs.lan/puppet-enterprise-vanagon/pe-activemq-2016.4.5.14.3-3/artifacts/deb/precise/)

mwaggett: Do we want to change those paths too though? I feel like we should..

underscorgan: I don't know. If we do we definitely need the change to be coordinated, the paths are for sure hardcoded in enterprise-dist.

PR #482, opened 2017-05-10
(WIP) (Working off Melissa's PR) (RE-8732) For puppet5, change the repo path to something more sensible

mwaggett:


lib/vanagon/platform.rb
```diff
@@ -239,7 +239,11 @@ def [](key)
     #   at the project level
     # @return [String] relative path to where packages should be output to
     def output_dir(target_repo = "")
-      @output_dir ||= File.join(@os_name, @os_version, target_repo, @architecture)
+      if target_repo == "PC1"
+        @output_dir ||= File.join(@os_name, @os_version, target_repo, @architecture)
+      else
+        @output_dir ||= File.join(target_repo, @os_name, @os_version, @architecture)
+      end
```
mckern: I wouldn't hardcode on `PC1` here. Remember, Vanagon is used outside of Puppet. Once this is in, there's a strong chance that it'll never come out or be updated again (or at least until the next big change).

lib/vanagon/utilities.rb
```diff
@@ -281,5 +281,16 @@ def erb_file(erbfile, outfile = nil, remove_orig = false, opts = { :binding => b
...
+
+    # Final
+    # - 1.0.0
+    # - 3.0.5.1
+    # - 4.3.0-5
+    # Nonfinal
+    # - 1.0.0-34-g1sadjf9283
+    # - 4.2.1.2344.gfdqw83rh
+    def final_version?(version)
+      !version.match(/g[a-f0-9]{7}$/)
```
mckern: I've been thinking about this a little bit. Would it be easier to try to parse out whether or not the version matches a specific `git describe` format -- and does this only apply to Git sources? If it does, shouldn't the logic for determining if something is final or not live there? 
I'm happy to talk my thoughts here through with you if you're interested.

lib/vanagon/platform.rb
mwaggett: yeah, I know it's not ideal. Haven't thought of a better solution yet, but also haven't given it much thought.

lib/vanagon/utilities.rb
mwaggett: yeah, @melissa wrote that part and I was already thinking of changing it because this seems really specific and I'd rather match for formats we DO want rather than ones we don't.

melissa: This isn't just for git sources, it's for the project as a whole. If the package is versioned from the git tag, then this matches, and if it's versioned from a string, I would assume it's always final (i.e., puppet-enterprise-vanagon projects). There reason I checked for the git describe is that we always have this in the version because it's in the automation. However, if we're building something from a tag or from a specific version (i.e., `proj.version '1.0.0'`, we have no control over what tag or version the user is going to specify. This could be `postgres-2017.4.0`, `1.0.0-1`, `1.0.0-rc0`, `random-string`, anything really. I wanted to keep this as simple as possible rather than trying to predict what crazy version strings we might encounter in the future.

mwaggett: hmm. I guess that makes sense, but it still feels weird to me to just be matching on one specific thing we don't want to call final. For example, the fact that '1.0.0-g1234567' is not a final version, but '1.0.0-g123456789' IS is strange. But I'm definitely open to ideas for how to do this better.

melissa: `g123456789` is a short sha that's generated based on the long sha of the commit `git rev-parse HEAD` will show you the long sha. It's always hexadecimal characters, hence the  `[a-f0-9]{7}`. The way vanagon gets that short sha is always going to be 7 characters in length. The way git reports it always has a leading `g`.

melissa: I suppose if you wanted a different check, you could say, is this versioned from a git tag or is the version hard-coded? If it's versioned from a tag, is the build we're doing current on a tag or not?

mckern: What are the possible formats that you'd expect to see this method parsing? Since you're writing a parser, just laying out as many of the expected formats you think you'll have to parse might provide some insight into how the parser should work.

mwaggett:
We've decided that all this logic should live in packaging. Closing.

PR #483, opened 2017-05-12
(maint) Fully qualify Git::GitExecuteError

joshcooper:
Since Git::GitExecuteError is referenced within a class named Git, we
need to fully qualify it, otherwise, we get:

    error uninitialized constant Vanagon::Component::Source::Git::GitExecuteError


mckern:
Oh noes, I can no longer merge.

PR #486, opened 2017-06-05
(VANAGON-49) install_file isn't respecting owner/group/mode

underscorgan:
Currently the owner/group/mode is set in the inidividual post-install
steps based on platform. However, some platforms do not have this
implemented. We can work around this by setting owner/group/mode when we
install files and directories on the build environment.

This seems to at least fix an issue with mode not being set for osx.
More in-depth testing should be done though.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -38,7 +38,7 @@ Lint/ShadowingOuterLocalVariable:
 Lint/LiteralInInterpolation:
   Enabled: true
 
-Style/MultilineMethodCallIndentation:
+Layout/MultilineMethodCallIndentation:
   EnforcedStyle: indented
   IndentationWidth: 2
 
@@ -57,7 +57,7 @@ Lint/AssignmentInCondition:
   Enabled: false
 
 # DISABLED - not useful
-Style/SpaceBeforeComment:
+Layout/SpaceBeforeComment:
   Enabled: false
 
 # DISABLED - not useful
@@ -136,7 +136,7 @@ Lint/UselessAssignment:
 Lint/Void:
   Enabled: false
 
-Style/AccessModifierIndentation:
+Layout/AccessModifierIndentation:
   Enabled: false
 
 Style/AccessorMethodName:
@@ -145,13 +145,13 @@ Style/AccessorMethodName:
 Style/Alias:
   Enabled: false
 
-Style/AlignArray:
+Layout/AlignArray:
   Enabled: false
 
-Style/AlignHash:
+Layout/AlignHash:
   Enabled: false
 
-Style/AlignParameters:
+Layout/AlignParameters:
   Enabled: false
 
 Metrics/BlockNesting:
@@ -169,7 +169,7 @@ Style/BracesAroundHashParameters:
 Style/CaseEquality:
   Enabled: false
 
-Style/CaseIndentation:
+Layout/CaseIndentation:
   Enabled: false
 
 Style/CharacterLiteral:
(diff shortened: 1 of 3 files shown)


underscorgan:
heeey @branan @McdonaldSeanp here's the start of the work for VANAGON-49. I need to make some unit-test updates and test that it doesn't break platforms other than osx, but using this and building from p-a b66a83414bbad1d24c91e43a13b4a26acf680f18 the wrapper.sh script had mode 0755 on the build machine and when the .dmg was installed.

McdonaldSeanp:
@underscorgan so, would this mean that group/mode/user would change in both the installation during build and the postinstall scripts for Linux platforms?

underscorgan:
Currently these are only changed in the postinst for debian and solaris 10, but yes, in those cases it would happen both during the build installation and the postinst. I don't think that's going to end up causing any issues. Mostly, I'm trying to fix the file modes not being set correctly on the builder issue, and due to how osx is packaged that also fixes osx not respecting modes. :woman_shrugging: 

It does look like this is causing issues for windows builds since those commands aren't defined, so need to look into that a little more.

McdonaldSeanp:
@underscorgan yeah, you might want to just ignore windows for this functionality for now. There really isn't a 1:1 mapping of file perms and user/group ownership between Unix and windows, so for instance the number value used for file mode will mean nothing to windows.

branan:
Can we add (yet another) warning when a Vanagon project tries to set these parameters on Windows? Ideally it would fail, since ignoring requested perms is potentially a security issue, but if we want to maintain compatibility I think a warning is worthwhile.

McdonaldSeanp:
I agree with @branan here, and also, my 2 cents about windows has become a little more opinionated. 

We should *not* attempt to force POSIX style file perms on to windows packages. The permission system on windows is much more complex than we can handle with the simple posix style strings. There are serious caveats to attempting to make that work (for instance, you actually have to specifically deny users access to files in windows) and I do not think vanagon should make any attempt to try and reconcile all of these issues.

For some background for others, check out the comments in https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/windows/security.rb for some docs and description of the problems trying to map posix to windows.

McdonaldSeanp:
I believe the better course of action on windows is to continue to improve the WiX workflow to make it easier to add WiX files representing specific files that are installed.

McdonaldSeanp:
Also, in my agreement with branan, I think that vanagon should actually fail when attempting to use posix file perms on windows packages. IMO it is definitely a security risk to allow builds to pass when setting perms is doing nothing.

underscorgan:
So, I was able to build on windows by using the full path to chmod, chgrp, and chown. This would set the files in the build environment but sounds like this wouldn't affect final packaging.  It sounds like you don't want this?

WRT fail vs warn I'm torn. Failing is definitely backwards incompatible (and will break existing puppet-agent code), but also I feel there is enough output from vanagon that you're likely to miss a warning. Failing seems better from a security-conscious standpoint, but I also feel like it could overly complicate code for projects building on windows and unix-y things with install_file and directory needing to get wrapped in

```
if platform.is_windows?
    ...
else
    ...
done
```

blocks any time you want to set ownership/mode, which feels gross.

McdonaldSeanp:
So: WiX heat.exe will not pick up source file perms when it's creating the WiX files that create the MSI packages. This means any changes we make to source files will have no effect on the files eventually laid down in the installer.

My suggestion with that information is to just ignore windows for this case.

underscorgan:
@McdonaldSeanp @branan thoughts? I know I'm missing tests for warning if it's windows / a good squashing, but does this look reasonable?

branan:
Yeah, this is fine for now. We should definitely look in to how to add ACL support for Windows - both for on-disk and wix.

underscorgan:
OK, so @demophoon had mentioned this yesterday but I thought maybe it would work, but users and groups are not set up on the build machines, so I'm going to modify this to only set the mode for now.

shrug:
So should we merge this as is and ticket further improvements? will this unblock @McdonaldSeanp and @branan as-is?

underscorgan:
@shrug I think it should. Running a test again to verify that this will make mode set for osx.

underscorgan:
On the build box
```
e6e273ozg3ogiy8:T root# find . -type f -name wrapper.sh
./tmp.apijonnU/puppet-agent-4.99.0.201.gbe33cf3/opt/puppetlabs/puppet/bin/wrapper.sh
./tmp.KANp8teI/osx/build/root/puppet-agent-4.99.0.201.gbe33cf3/opt/puppetlabs/puppet/bin/wrapper.sh
e6e273ozg3ogiy8:T root# ls -la ./tmp.KANp8teI/osx/build/root/puppet-agent-4.99.0.201.gbe33cf3/opt/puppetlabs/puppet/bin/wrapper.sh
-rwxr-xr-x  1 root  wheel  195 Jun  9 13:57 ./tmp.KANp8teI/osx/build/root/puppet-agent-4.99.0.201.gbe33cf3/opt/puppetlabs/puppet/bin/wrapper.sh
```

On install:
```
evjt7u159bud13v:bin root# ls -la wrapper.sh
-rwxr-xr-x  1 root  wheel  195 Jun  9 13:57 wrapper.sh
evjt7u159bud13v:bin root# pwd
/opt/puppetlabs/puppet/bin
```

underscorgan:
@shrug @branan @McdonaldSeanp tests look good, this work for y'alls?

PR #488, opened 2017-06-29
(VANAGON-62) Update AIX services on package upgrade

branan:
Previously, we created AIX services once at install, but did
not have correct logic to update them if they changed. This
adds the necessary logic to reconfigure AIX services.


McdonaldSeanp:
👍 lgtm

PR #490, opened 2017-06-29
(VANAGON-63) Generate additional build metadata

demophoon:
This commit adds in the ability for vanagon to generate additional build
data for downstream package promotion consumption.

### DIFF
--- a/lib/vanagon.rb
+++ b/lib/vanagon.rb
@@ -1,5 +1,9 @@
+require 'time'
+
 LIBDIR = File.expand_path(File.dirname(__FILE__))
 VANAGON_ROOT = File.join(File.expand_path(File.dirname(__FILE__)), "..")
+BUILD_TIME = Time.now.iso8601
+VANAGON_VERSION = Gem.loaded_specs["vanagon"].version.to_s
 
 $:.unshift(LIBDIR) unless
   $:.include?(File.dirname(__FILE__)) || $:.include?(LIBDIR)
--- a/lib/vanagon/component.rb
+++ b/lib/vanagon/component.rb
@@ -290,6 +290,10 @@ class Vanagon
       end
     end
 
+    def get_dependency_hash
+      { name => { 'version' => version, 'ref' => options[:ref] }.delete_if { |_, v| !v } }
+    end
+
     # Fetches secondary sources for the component. These are just dumped into the workdir currently.
     #
     # @param workdir [String] working directory to put the source into
--- a/lib/vanagon/project.rb
+++ b/lib/vanagon/project.rb
@@ -476,5 +476,44 @@ class Vanagon
     def generate_packaging_artifacts(workdir)
       @platform.generate_packaging_artifacts(workdir, @name, binding)
     end
+
+    # Generate a json hash which lists all of the dependant components of the
+    # project.
+    #
+    # @return [Hash] where the top level keys are components and their values
+    #   are hashes with additional information on the component.
+    def generate_dependencies_info
+      @components.each_with_object({}) do |component, hsh|
+        hsh.merge!(component.get_dependency_hash)
+      end
+    end
+
+    # Generate a hash which contains relevant information regarding components
+    # of a package, what vanagon built the package, time of build, as well as
+    # version of the thing we were building.
+    #
+    # @return [Hash] of information which is useful to know about how a package
+    #   was built and what went into the package.
+    def build_manifest_json
+      manifest = {
+        "packaging_type" => {
+          "vanagon" => VANAGON_VERSION,
+        },
+        "version" => version,
+        "components" => generate_dependencies_info,
+        "build_time" => BUILD_TIME,
+      }
+    end
+
+    # Writes a json file at `ext/build_metadata.json` containing information
+    # about what went into a built artifact
+    #
+    # @return [Hash] of build information
+    def save_manifest_json
+      manifest = build_manifest_json
+      File.open(File.join('ext', 'build_metadata.json'), 'w') do |f|
+        f.write(JSON.pretty_generate(manifest))
+      end
+    end
   end
 end
--- a/lib/vanagon/driver.rb
+++ b/lib/vanagon/driver.rb
@@ -124,6 +124,7 @@ class Vanagon
       @project.make_makefile(@workdir)
       @project.make_bill_of_materials(@workdir)
       @project.generate_packaging_artifacts(@workdir)
+      @project.save_manifest_json
       @engine.ship_workdir(@workdir)
       @engine.dispatch("(cd #{@engine.remote_workdir}; #{@platform.make})")
       @engine.retrieve_built_artifact
--- a/spec/lib/vanagon/project_spec.rb
+++ b/spec/lib/vanagon/project_spec.rb
@@ -115,4 +115,55 @@ describe 'Vanagon::Project' do
       expect(@proj.filter_component(comp1.name)).to eq([comp1, comp2, comp3])
     end
   end
+
+  describe '#generate_dependencies_info' do
+    before(:each) do
+      @proj = Vanagon::Project.new('test-project', {})
+    end
+
+    it "returns a hash of components and their versions" do
+      comp1 = Vanagon::Component.new('test-component1', {}, {})
+      comp1.version = '1.0.0'
+      comp2 = Vanagon::Component.new('test-component2', {}, {})
+      comp2.version = '2.0.0'
+      comp2.options[:ref] = '123abcd'
+      comp3 = Vanagon::Component.new('test-component3', {}, {})
+      @proj.components << comp1
+      @proj.components << comp2
+      @proj.components << comp3
+
+      expect(@proj.generate_dependencies_info()).to eq({
+        'test-component1' => { 'version' => '1.0.0' },
+        'test-component2' => { 'version' => '2.0.0', 'ref' => '123abcd' },
+        'test-component3' => {},
+      })
+    end
+  end
+
+  describe '#build_manifest_json' do
+    before(:each) do
+      class Vanagon
+        class Project
+          BUILD_TIME = '2017-07-10T13:34:25-07:00'
+          VANAGON_VERSION = '0.0.0-rspec'
+        end
+      end
+
+      @proj = Vanagon::Project.new('test-project', {})
+    end
+
+    it 'should generate a hash with the expected build metadata' do
+      comp1 = Vanagon::Component.new('test-component1', {}, {})
+      comp1.version = '1.0.0'
+      @proj.components << comp1
+      @proj.version = '123abcde'
+
+      expect(@proj.build_manifest_json()).to eq({
+        'packaging_type' => { 'vanagon' => '0.0.0-rspec' },
+        'version' => '123abcde',
+        'components' => { 'test-component1' => { 'version' => '1.0.0' } },
+        'build_time' => '2017-07-10T13:34:25-07:00',
+      })
+    end
+  end
 end


demophoon:
Sample build output: http://builds.delivery.puppetlabs.net/pe-modules/a320b6d1c727d1962ade8eeba172434b8c03003b/artifacts/?C=M&O=D

mckern:
Vanagon PRs should have Vanagon ticket numbers in them, shouldn't they? There's a Rubocop failure here which is trivial, but I'll have a couple of comments inline maybe?

lib/vanagon/project.rb
```diff
@@ -476,5 +477,36 @@ def generate_package
...
+        components_hash = component.options
+        [:version, :url].each do |key|
+          if val = component.send(key)
+            components_hash[key] = val
+          end
+        end
+        [
+          component.name,
+          components_hash,
+        ]
```
mckern: This can be collapsed into a single line

mckern: Can this logic be moved into the Components, so that it becomes something you can query them for instead of something you have to derive here?

mckern: (I have some nitpicks about this code, but if it's something that can be moved then maybe those nitpicks will resolve themselves in the move)

lib/vanagon/project.rb
```diff
@@ -476,5 +477,36 @@ def generate_package
...
+      end.to_h
+    end
+
+    def generate_manifest_json
+      build_time = Time.now
```
mckern: This should be moved way out into top most scope, shouldn't it? It'd almost make sense as a global that can be referenced anywhere in the codebase. It's a handy piece of data to have access to.

lib/vanagon/project.rb
mckern: (specifically, there's no documentation about what this method should return. It looks like you're going to build a multidimensional Array and then cast it to a Hash. You'd get something like this back:

```ruby
>> [["a", {b: nil, c: nil}],["b", {c: nil, d: nil}]].to_h
=> {
  "a" => {
    :b => nil,
    :c => nil
  },
  "b" => {
    :c => nil,
    :d => nil
  }
}
```

That makes sense and I think it's a solid approach. But the path to getting there is kind of winding since you're constructing everything in the Project and offloading nothing to the Components themselves.)

lib/vanagon/project.rb
```diff
@@ -476,5 +477,36 @@ def generate_package
...
+      manifest = {
+        "packaging_type" => {
+          "vanagon" => Gem.loaded_specs["vanagon"].version.to_s,
```
mckern: What kind of cranky version numbers does this produce if you're using Vanagon as a developer, all tangled up in Bundler and being loaded from a local directory?

lib/vanagon/project.rb
```diff
@@ -476,5 +477,36 @@ def generate_package
...
+        },
+        "git_sha" => @version,
```
mckern: 1. use the accessor (`#version`, not `@version`)
2. is `#version` always going to be a Git SHA?
3. what will happen if `#version` is neither a String nor a Git SHA?

lib/vanagon/project.rb
```diff
@@ -476,5 +477,36 @@ def generate_package
...
+        "components" => generate_dependencies_info,
+        "build_time" => build_time.rfc2822,
```
mckern: Why not just store `build_time` as RFC2822? Also, maybe consider ISO8601 instead of RFC2822:

```ruby
>> Time.now.iso8601
=> "2017-07-03T23:31:42-07:00"
```
(RFC2822 is for email timestamp headers. ISO8601 is a legit worldwide standard.)

lib/vanagon/project.rb
```diff
@@ -476,5 +477,36 @@ def generate_package
...
+      }
+      File.open(File.join('ext', 'build_metadata.json'), 'w') do |f|
```
mckern: When and where is `./ext` created? And should this reference `VANAGON_ROOT` when it constructs this path?

demophoon: The `./ext` directory should exist in the puppetlabs projects. Maybe I should create it here in case it doesn't exist in non-pl projects? Or conditionally put this file in `ext` if that directory does exist.

It is currently there because the packaging gem will pick up the file in this location and ship it to the build server.

lib/vanagon/project.rb
```diff
@@ -476,5 +476,33 @@ def generate_package
...
+    # Writes a json file at `ext/build_metadata.json` containing information
+    # about what went into a built artifact
+    def generate_manifest_json
+      manifest = {
+        "packaging_type" => {
+          "vanagon" => Gem.loaded_specs["vanagon"].version.to_s,
+        },
+        "git_sha" => version,
+        "components" => generate_dependencies_info,
+        "build_time" => BUILD_TIME.iso8601,
```
nwolfe: We used that other time format in the corresponding ezbake changes.

I'm wondering if it will be awkward to use different formats to represent the same thing.

Not sure.

mckern: If you're using `Time.parse()` it'll ultimately end up fine. Guidance would be to prefer ISO date formats to RFC date formats.

demophoon: Yeah, `Time.parse` works with both formats. I originally picked RFC-2822 because it was a little more human friendly than ISO-8601 but I can honestly go either way here. So long as there is a standard for it :man_shrugging:

lib/vanagon/component.rb
```diff
@@ -290,6 +290,14 @@ def get_build_dir
       end
     end
 
+    def get_dependency_hash
+      obj = @options
+      [:version, :name].each do |var|
+        obj[var.to_s] = self.instance_variable_get("@#{var}")
```
nwolfe: Isn't this going to modify the `@options` variable?  Is that intended?

Also this doesn't seem necessary to metaprogram since it's just 2 variables.

I'd rather just be direct and do:
```ruby
obj = @options
obj['version'] = @version
obj['name'] = @name
{@name => obj}
```

But then the `obj` variable seems unnecessary, so I'd consider just doing this:
```ruby
@options['version'] = @version
@options['name'] = @name
{@name => @options}
```

The `obj` only seems necessary if we're going to make a clone of `@options` so we don't modify it here.

lib/vanagon/project.rb
mckern: The RFC specifically pertains to timestamps related to internet protocols (which is why its RFC; that's a very specific prefix for a very specific organization). ISO means it's an internationally defined standard for use outside of the very specific cases that the RFC would apply to.

mckern:
@demophoon I'm going to tag out. You just shout when you think this is polished up enough for review.

lib/vanagon/component.rb
demophoon: I was only doing that just incase later on we want to get more data out.

lib/vanagon/project.rb
nwolfe: I see.

I'm not worried about the formats themselves, just the fact that we're using different ones to represent the same thing.

If we knew that it was better to use ISO over RFC then we should've used ISO in the ezbake change.

How are we going to be using these timestamps?  
Aren't we going to be comparing them?  
Will we ever be comparing the timestamp from ezbake with the timestamp from vanagon (i.e. we're comparing a timestamp in ISO format with a timestamp in RFC format)? 
Is that going to work properly, or will we need to convert one to the other?

Not trying to ask leading questions here; maybe all this will work seamlessly 😄

mckern: You should *never* be comparing a timestamp directly. It's much safer to parse the timestamps into objects, and then use the Time library to compare them. By parsing them into `Time` objects, you'll get comparisons for free:

```
>> rfc_time = Time.now.rfc2822
=> "Thu, 06 Jul 2017 13:48:12 -0700"
>> iso_time = Time.now.iso8601
=> "2017-07-06T13:48:22-07:00"
>> Time.parse(iso_time) - Time.parse(rfc_time)
=> 10.0
>> Time.parse(iso_time) > Time.parse(rfc_time)
=> true
>> Time.parse(iso_time) < Time.parse(rfc_time)
=> false
```

lib/vanagon/component.rb
nwolfe: > I was only doing that just incase later on we want to get more data out.

Sure, but it's also not hard to add another line that plucks out the value from `self` and puts it into the map.

I don't like metaprogramming this simple thing; it has more negatives than positives.

I think we should simplify this.

lib/vanagon/project.rb
nwolfe: Yeah, we were never going to compare the strings themselves, but I wasn't sure if we'd need to modify the `Time` objects before comparing.  Glad to see we don't!

Thanks @mckern

mckern: Anyway, I agree that the RFC format is more human readable. If the concern is to build something a human can read easily that's valid. If the concern is to build something that a machine compares and parses, it doesn't matter and the argument is semantic at best. Like I said, it's guidance. But do what's best for your users at the end of the day.

lib/vanagon/component.rb
```diff
@@ -290,6 +290,14 @@ def get_build_dir
       end
     end
 
+    def get_dependency_hash
+      obj = @options
+      %i[version name].each do |var|
+        obj[var.to_s] = self.instance_variable_get("@#{var}")
```
nwolfe: I don't think we should metaprogram this kind of stuff.

Metaprogramming is really only useful when you don't know what "the string" is, like something that's input from the user/outside.

Just access the 2 variables directly.

mckern: I'm inclined to agree with Nate. No reason to even add iteration here:

```ruby
def get_dependency_hash
  { name => { options[:name], options[:version] } }
end
```

That's pretty gosh darned readable

mckern: Oh I missed a bit. BRB.

demophoon: I actually just remembered why I went this route in the first place. There are some components in some projects which do not have versions/shas associated with them. The code which had been written before still would have failed had either of those existed. That was the reason why I had been iterating over the list of symbols. It should have been if the value was not nil then add it to the hash I'm okay with removing the metaprogramming stuff 
 and handling that in a different way.

lib/vanagon.rb
```diff
@@ -1,5 +1,8 @@
+require 'time'
+
 LIBDIR = File.expand_path(File.dirname(__FILE__))
 VANAGON_ROOT = File.join(File.expand_path(File.dirname(__FILE__)), "..")
+BUILD_TIME = Time.now
```
mckern: Quick note: maybe cast this to a string up front, so you don't have to worry about converting it down the line. You don't do anything directly with the object, so don't hang onto a Time object longer than you really have to.

demophoon: 👍

lib/vanagon/project.rb
```diff
@@ -476,5 +476,33 @@ def generate_package
     def generate_packaging_artifacts(workdir)
       @platform.generate_packaging_artifacts(workdir, @name, binding)
     end
+
+    # Generate a json file which lists all of the dependant components of the project
```
nwolfe: This looks like the docstring for the `generate_manifest_json` method?

This method doesn't actually generate a JSON file, but a Hash, right?

demophoon: Correct, my bad.

lib/vanagon/component.rb
```diff
@@ -290,6 +290,13 @@ def get_build_dir
       end
     end
 
+    def get_dependency_hash
+      deps = {}
+      deps['version'] = @version if @version
+      deps['ref'] = @options[:ref] if @options[:ref]
+      { name => deps }
+    end
```
mckern: Use the accessors, not the instance variables. That's what they're there for. And no need to check first, since the value of those keys in a Hash would be `nil` anyway:

```ruby
{ name => { 'version' => version, 'ref' => ref } }
```

if you _really, **really**_ want to strip out empty keys, try this:

```ruby
{ name => { 'version' => version, 'ref' => ref }.delete_if { |k, v| v.empty? } }
```

(I generally wouldn't worry about whether the keys are in there even if they're empty, since, again, any key in a Hash can be called and will return `nil` if it's not defined or if it's empty.)

mckern: Actually, this form would work better since not every object will respond to `#empty?`:

```ruby
>> hsh = { 'arugula' => { 'flavorful': true, 'affordable': nil }.delete_if { |k, v| !v } }
=> {
  "arugula" => {
    :flavorful => true
  }
}
>> puts JSON.pretty_generate hsh
{
  "arugula": {
    "flavorful": true
  }
}
=> nil
>> hsh = { 'arugula' => { 'flavorful': true, 'affordable': nil }.delete_if { |k, v| !v } }
```

lib/vanagon/component.rb
```diff
@@ -290,6 +290,10 @@ def get_build_dir
       end
     end
 
+    def get_dependency_hash
+      { name => { 'version' => version, 'ref' => options[:ref] }.delete_if { |k, v| !v } }
```
nwolfe: What is this magic?

Why not just build a simple hash like we had it?

We need 2 keys in a hash, that's it.

I don't even know what this does by looking at it.

Whatever :man_shrugging:

mckern: This is a simpler hash than the previous construction. There's no intermediate hash being allocated, and there's no conditional population. I'm not sure what magic you're referring to?

nwolfe: Optimize for human readability until this becomes a performance bottleneck.

There's still iteration, there's still conditionals.

lib/vanagon/project.rb
```diff
@@ -476,5 +476,34 @@ def generate_package
...
+    #
+    # @return [Hash] where the top level keys are components and their values
+    #   are hashes with additional information on the component.
+    def generate_dependencies_info
+      all_components = {}
+      @components.each do |component|
+        all_components = all_components.merge(component.get_dependency_hash)
+      end
+      all_components
+    end
```
mckern: You can make this more concise using `#each_with_object`:

```ruby
components.each_with_object({}) do |component, hsh|
  hsh.merge! all_components.merge(component.get_dependency_hash)
end
```

lib/vanagon/component.rb
mckern: I wouldn't even purge empty values since JSON serialization understands `nil` as `null`. That'd just look like this:

```ruby
>> puts JSON.pretty_generate({ 'rando' => { 'version' => '1.2.3', 'ref' => nil }})
{
  "rando": {
    "version": "1.2.3",
    "ref": null
  }
}
>> JSON.parse(JSON.pretty_generate({ 'rando' => { 'version' => '1.2.3', 'ref' => nil }}))
=> {
  "rando" => {
    "version" => "1.2.3",
        "ref" => nil
  }
}
```
----

Also, I strongly disagree that the suggested oneliner is unreadable. That's a very, very common Ruby construction. It could be spread over a few lines if it really bothers you:

```ruby
{ name => { 'version' => version, 'ref' => options[:ref] }
  .delete_if { |k, v| !v } }

# or

{ 
  name => {
    'version' => version, 
    'ref' => options[:ref]
  }.delete_if { |k, v| !v }
}

File: lib/vanagon/project.rb [hunk 15]
```diff
@@ -476,5 +476,34 @@ def generate_package
...
+      manifest = {
+        "packaging_type" => {
+          "vanagon" => Gem.loaded_specs["vanagon"].version.to_s,
+        },
+        "git_sha" => version,
+        "components" => generate_dependencies_info,
+        "build_time" => BUILD_TIME,
+      }
+      File.open(File.join('ext', 'build_metadata.json'), 'w') do |f|
+        f.write(manifest.to_json)
```
mckern: Do you want this pretty printed?

lib/vanagon/component.rb
nwolfe: I understand the desire to generalize this, but this entire method is self-contained and only operating on 2 elements.

If this method was taking input from outside and needed to be agnostic to the actual values its operating on, something that's general-purpose like this would be useful.

Arguments about performance, and how many hashes it may make, and how many conditionals it uses are valid, but at the end of the day there's no way around some of these things.

We still need conditionals to only insert elements if they're non-nil.  The simplest conditional is an `if` statement, so start with that.

Regarding how many hashes are created in the process, do we know the implementation of `delete_if`?  Maybe it creates an internal hash to modify so it doesn't get concurrent modification problems when attempting to iterate and remove elements at the same time.

Using `!` to test for nil isn't exactly obvious either, since that's usually something used for boolean values.  Why not use the `.nil?` method in this case?

Just because we can write it tersely doesn't mean the performance is better or that it's easier to understand what's going on.

> I wouldn't even purge empty values since JSON serialization understands nil as null

I wouldn't necessarily assume that this method will be only used for generating JSON.  We almost had it generate YAML, and it may be used by other methods that may do something different with the hash.

I would keep all concerns specific to JSON in the one method that actually writes it out as JSON.

> Also, I strongly disagree that the suggested oneliner is unreadable. That's a very, very common Ruby construction.

I strongly disagree that this is the most understandable way to write this.  Being a common Ruby construct doesn't make it more understandable.

And after all that, while there are many ways we can achieve the desired hash in Ruby, we need to move forward with the rest of the PEGI effort and not be tempted to rewrite this over and over.

I'd say we've all probably learned something from these discussions, and we're all the better for it, but time is of the essence!

We're nearly done with this change, so let's stay focused and finish whatever else is necessary to get this across the line.

lib/vanagon/project.rb
nwolfe: Looks like we could do `f.write JSON.pretty_generate(manifest)` to pretty-print it.

lib/vanagon/component.rb
mckern: Nate, that's a lot of words to say "I disagree with you and I want you to stop pushing it."

> I wouldn't necessarily assume that this method will be only used for generating JSON. We almost had it generate YAML, and it may be used by other methods that may do something different with the hash.

```ruby
>> puts YAML.dump({ 'rando' => { 'version' => '1.2.3', 'ref' => nil }})
---
rando:
  version: 1.2.3
  ref:
=> nil
>> YAML.load(YAML.dump({ 'rando' => { 'version' => '1.2.3', 'ref' => nil }}))
=> {
  "rando" => {
    "version" => "1.2.3",
        "ref" => nil
  }
}
```

They're serialization formats; by definition generators have to have some mechanism to deal with dumping values like `nil` into values that the parser than understand. Just let the values be `nil`.

> Using ! to test for nil isn't exactly obvious either, since that's usually something used for boolean values. Why not use the .nil? method in this case?

I didn't suggest just calling `#nil?` because of the way everything in Ruby is truthy except `nil` and `false`. But like I said, I wouldn't even bother trying to keep empty keys out. Filter them however you like if you have your heart set on filtering.

> I'd say we've all probably learned something from these discussions, and we're all the better for it, but time is of the essence!
>
> We're nearly done with this change, so let's stay focused and finish whatever else is necessary to get this across the line.

If you've got time to argue about it, you've got time to do it right. Things that happen in Vanagon seldom get revisited once they're in. Vanagon is consumed by entities outside of Puppet. A feature may start out with a Puppet-specific need, but that feature will still be available for folks outside of Puppet to use. So that means that there's a real cost for getting even private internal API stuff like this not-quite-right. I'm asking Britt (and by extension you) to consider that cost.

lib/vanagon/project.rb
mckern: That's what I was getting at, if you want it pretty-printed. I didn't want to assume.

lib/vanagon/component.rb
mckern: > Regarding how many hashes are created in the process, do we know the implementation of delete_if? Maybe it creates an internal hash to modify so it doesn't get concurrent modification problems when attempting to iterate and remove elements at the same time.

It's part of Ruby core, so it's written in C. That means it's going to be significantly faster than anything you write in Ruby to do the same thing. But since you asked:
 
From the [doc](https://ruby-doc.org/core-2.2.0/Array.html#method-i-delete_if)

```c
static VALUE
rb_ary_delete_if(VALUE ary)
{
    RETURN_SIZED_ENUMERATOR(ary, 0, 0, ary_enum_length);
    ary_reject_bang(ary);
    return ary;
}
```

nwolfe: Haha we could argue about Ruby all day 😄 

I've succumbed to the state of the code, so I think we've resolved this.

Have a good weekend @mckern 

I look forward to seeing you at the carts again soon

lib/vanagon/project.rb
demophoon: 👍 Didn't know about `each_with_object`.

lib/vanagon/project.rb
demophoon: This is done in the latest commit

lib/vanagon/component.rb
demophoon: I think i'm going to leave this as is. I do not want to get into the lower level implementation details between what is the best way of creating a simple hash like this. If we have enough components that choosing between the different ways to create a hash will adversely affect performance then there are probably bigger and better problems to solve before getting to this one.

lib/vanagon/project.rb
```diff
@@ -476,5 +476,32 @@ def generate_package
...
     end
+
+    # Generate a json hash which lists all of the dependant components of the
+    # project.
+    #
+    # @return [Hash] where the top level keys are components and their values
+    #   are hashes with additional information on the component.
+    def generate_dependencies_info
+      @components.each_with_object({}) do |component, hsh|
+        hsh.merge(component.get_dependency_hash)
```
mckern: You tested that `#merge` will do the right thing, and that you don't need `#merge!`?

mckern:
Looks good. Rub some tests on it, maybe?

lib/vanagon/project.rb
demophoon: Found that with the new test, yay tests

lib/vanagon/component.rb
mckern: Unrelated, I think Rubocop will complain about an unused key in the iterators. It'll ask you to use `_k` or `_` instead of `k`. It's strictly a style nitpick and I wanted to make sure you knew about it if Rubocop hadn't squawked yet.

lib/vanagon/project.rb
```diff
@@ -476,5 +476,32 @@ def generate_package
...
+    # about what went into a built artifact
+    def generate_manifest_json
+      manifest = {
+        "packaging_type" => {
+          "vanagon" => Gem.loaded_specs["vanagon"].version.to_s,
+        },
+        "git_sha" => version,
+        "components" => generate_dependencies_info,
+        "build_time" => BUILD_TIME,
+      }
```
mckern: Suggestion: split up the structure generation and the write operation into two discrete methods. They're both fine as-is, just decouple them.

lib/vanagon/component.rb
demophoon: Changed

demophoon:
@nwolfe @mckern Added a few tests as well as separated out the file writing from the hash generation.

Since we are going to need to make another change to ezbake to switch from the RFC to ISO timestamp i've also changed `git_sha` key in the metadata hash to `version` which will support projects which are not backed by git in vanagon but should still get the point across in ezbake land.

PR #491, opened 2017-06-30
(maint) Add gem components

DavidS:
This shortens the text required for defining a gem component significantly.

### DIFF
--- a/lib/vanagon/component/dsl.rb
+++ b/lib/vanagon/component/dsl.rb
@@ -25,6 +25,30 @@ class Vanagon
         yield(self, @component.settings, @component.platform)
       end
 
+      # Define a gem component with default values. Automatically sets `version`, `url`, `build_requires`,
+      # and `install`. Fixes up the PATH for Windows. Requires `ruby_version`, `gem_path_env`, and
+      # `gem_install` settings.
+      #
+      # @param name [String] name of the gem
+      # @param version [String] name of the componennt
+      # @param block [Proc] DSL definition of the component to call
+      def gem(name, version, &block)
+        self.version version
+        url "https://rubygems.org/gems/#{name}-#{version}.gem"
+
+        build_requires "ruby-#{@component.settings[:ruby_version]}"
+
+        if @component.platform.is_windows?
+          environment "PATH", @component.settings[:gem_path_env]
+        end
+
+        install do
+          "#{@component.settings[:gem_install]} #{name}-#{version}.gem"
+        end
+
+        yield(self, @component.settings, @component.platform) if block_given?
+      end
+
       # Accessor for the component.
       #
       # @return [Vanagon::Component] the component the DSL methods will be acting against


DavidS:
See https://github.com/puppetlabs/puppet-sdk-vanagon/compare/master...DavidS:wip-vanagon-gem-component for an example usage.

DavidS:
PS: I'll be on PTO for the next two weeks, so please don't expect any quick responses!

lib/vanagon/component/dsl.rb
```diff
@@ -25,6 +25,30 @@ def component(name, &block)
...
+      # `gem_install` settings.
+      #
+      # @param name [String] name of the gem
+      # @param version [String] name of the componennt
+      # @param block [Proc] DSL definition of the component to call
+      def gem(name, version, &block)
+        self.version version
+        url "https://rubygems.org/gems/#{name}-#{version}.gem"
+
+        build_requires "ruby-#{@component.settings[:ruby_version]}"
```
mckern: Use the accessor `#_component` instead of the instance variable `@component`

lib/vanagon/component/dsl.rb
```diff
@@ -25,6 +25,30 @@ def component(name, &block)
...
+      # Define a gem component with default values. Automatically sets `version`, `url`, `build_requires`,
+      # and `install`. Fixes up the PATH for Windows. Requires `ruby_version`, `gem_path_env`, and
+      # `gem_install` settings.
+      #
+      # @param name [String] name of the gem
+      # @param version [String] name of the componennt
+      # @param block [Proc] DSL definition of the component to call
+      def gem(name, version, &block)
+        self.version version
+        url "https://rubygems.org/gems/#{name}-#{version}.gem"
```
mckern: I think that this will set the component source type's URL as an [HTTP component](https://github.com/puppetlabs/vanagon/blob/master/lib/vanagon/component/source.rb#L27-L34), which should also then require a checksum. Does that not happen?

mckern:
Can you provide any examples of using this for the PR? Also, it'll need tests & at least one example in the `examples/components` directory. I'm not entirely sold on its utility but I'm also not opposed; I just wonder about what sort of precedent it opens up for future convenience methods in the DSL.

(As an aside, the name is kind of misleading, since a Gem component will still just be resolved as an HTTP component. `gem install <name>` won't really be run; `curl -O <url>; gem install <name>.gem` will be, unless that code path has changed significantly in the last ~60 days. Probably worth at least a little conversation or background about how you got here and who this helps.)

DavidS:
I just arrived back in the office after PTO. Here are a few quick answers. I'll look into updating the code for rubocop, and add tests and examples later this week.

**Examples:** as linked above, https://github.com/puppetlabs/puppet-sdk-vanagon/compare/master...DavidS:wip-vanagon-gem-component is the example I used to test this. It cuts the gem definition from 15 to 3 lines.

**Convenience methods:** Yes, please! Also, a library of pre-fabbed components.

**Background:** The example at the top also answers how I got here. Updating/adding bundled gems for that package is quite painful. Copying gem files to buildsources, and cloning/editing the current required codez even more so. And, we'll need more of that in the future.

mckern:
> Examples: as linked above, puppetlabs/puppet-sdk-vanagon@master...DavidS:wip-vanagon-gem-component is the example I used to test this. It cuts the gem definition from 15 to 3 lines.

No, I mean there is an `examples/` directory. Please add an example to it in this PR.

> Convenience methods: Yes, please! Also, a library of pre-fabbed components.

I have opinions here, but someone from Release Engineering should probably chime in before I get really deep into them. The TLDR is that because Vanagon is a general purpose tool that's used outside of Puppet, it'd probably make more sense for it to help enable caching prebuilt components that users have defined per-project instead of making prefabricated components available to users. Otherwise you're 2/3 of the way to maintaining a distro.

> Updating/adding bundled gems for that package is quite painful. Copying gem files to buildsources, and cloning/editing the current required codez even more so. And, we'll need more of that in the future.

The mirror URL functionality should have alleviated at least some of this pain. And copying gems to the internal mirror at Puppet is automated with a chatbot script. Ask Kerminator about the `buildsources` command in the `release-new` channel? Alternatively, there is a quick-and-dirty CGI endpoint which can be easily extended. I'm not sure that addressing the pain points of mirroring is best solved with Vanagon PRs.

shrug:
I believe this adds some implicit dependencies on external infrastructure that we don't necessarily want? Currently all of our build deps are hosted internally, for example we've historically had bad luck with the availability of rubygems.org.

DavidS:
thank you all for your comments, but I'm currently not able to give this the required focus, especially with the wider context questions raised by @highb . I'm looking forward to your visit in Belfast to revisit this in an easier setting.

PR #492, opened 2017-07-11
(RE-8839) Stage packages in platform tag directory 

melissa:
This commit updates how built packages are staged. Rather than being
opinionated on repo or path structure for specific platforms, this
commit changes the default behavior of vanagon to simply stage the
packages under the platform tag (platform-version-architecture). This
allows us to concentrate the automation that is opinionated about path
structure and repo formats into puppetlabs/packaging, which comes into
play when shipping out the packages.


melissa:
with `proj.target_repo` unset, we get:
```
output/ubuntu-16.04-amd64/puppet-client-tools_1.1.1.61.ge3d0c98-1xenial_amd64.deb
output/sles-12-x86_64/puppet-client-tools-1.1.1.61.ge3d0c98-1.sles12.x86_64.rpm
output/el-7-x86_64/puppet-client-tools-1.1.1.61.ge3d0c98-1.el7.x86_64.rpm
```

with `proj.target_repo "PC1"` in the project config, we get:
```
output/el/7/PC1/x86_64/puppet-client-tools-1.1.1.61.ge3d0c98-1.el7.x86_64.rpm
output/deb/xenial/PC1/puppet-client-tools_1.1.1.61.ge3d0c98-1xenial_amd64.deb
```

underscorgan:
I think this looks reasonable. Can you update tests for both old + new behavior?

melissa:
Absolutely. I just wanted to make sure this was sufficient

lib/vanagon/platform/rpm.rb
```diff
@@ -45,15 +45,15 @@ def package_name(project)
         "#{project.name}-#{project.version}-#{project.release}.#{project.noarch ? 'noarch' : @architecture}.rpm"
       end
 
-      def output_dir(target_repo = "products")
-        super
-      end
```
underscorgan: We should keep in mind that products is hardcoded around a bunch, so this will break pe stuff at least.

lib/vanagon/platform/rpm.rb
```diff
@@ -45,15 +45,15 @@ def package_name(project)
...
-
       # Method to derive the directory for source artifacts
       #
       # @param target_repo [String] repo the source artifacts are targeting
-      def source_output_dir(target_repo = "products")
-        @source_output_dir ||= File.join(@os_name, @os_version, target_repo, 'SRPMS')
+      def source_output_dir(target_repo = nil)
+        if target_repo.nil?
+          super
```
underscorgan: what do these end up looking like when there's no target_repo? are we going to end up with the srpm/source tags like http://enterprise.delivery.puppetlabs.net/2017.3/repos/ ?

melissa: I figured I'd default to the platform tag with the builder architecture, but I can change this to default to the srpm/source directories like we do for the enterprise repos. That's quite logical to do

underscorgan: I think it makes sense to keep those separated out

lib/vanagon/platform/rpm.rb
melissa: that's going to be a huge pain in the butt :(

lib/vanagon/platform/rpm.rb
melissa: done

shrug:
Blocked on beaker work (BKR-1118)

lib/vanagon/platform/rpm.rb
melissa: BUT WORTH IT TO FIX. It looks like that's mainly going to be a pain with promotions. We'll have to coordinate this and the fix to the promotion code in enterprise-dist

underscorgan: oh yes, for sure! This was mostly a note so future us didn't forget we broke this last time :)

spec/lib/vanagon/platform_spec.rb
```diff
@@ -24,10 +24,10 @@
...
         :os_version                     => "5",
         :architecture                   => "i386",
-        :output_dir                     => "el/5/products/i386",
+        :output_dir                     => "el-5-i386",
         :output_dir_with_target         => "el/5/thing/i386",
         :output_dir_empty_string        => "el/5/i386",
-        :source_output_dir              => "el/5/products/SRPMS",
+        :source_output_dir              => "el-5-srpms",
         :source_output_dir_with_target  => "el/5/thing/SRPMS",
         :source_output_dir_empty_string => "el/5/SRPMS",
```
melissa: @underscorgan  e already have quite a few tests surrounding old and new behavior, luckily!

melissa:
I built puppet-agent both with and without `proj.target_repo = 'pc1'` set. Both did the correct thing.

Built with `proj.target_repo = 'pc1'`:
http://builds.puppetlabs.lan/puppet-agent/e0e73ccd9f571e81e5421da87a7c4562f4d45df9/
Built without `proj.target_repo = 'pc1'`:
http://builds.puppetlabs.lan/puppet-agent/e0e73ccd9f571e81e5421da87a7c4562f4d45df9/

melissa:
Even though bumping packaging 1.0.x is a big step here, it looks like everything I did in packaging that gets used in vanagon (sign, ship, & repos) is backwards compatible.

lib/vanagon/platform/rpm.rb
```diff
@@ -45,15 +45,15 @@ def package_name(project)
...
       # Method to derive the directory for source artifacts
       #
       # @param target_repo [String] repo the source artifacts are targeting
-      def source_output_dir(target_repo = "products")
-        @source_output_dir ||= File.join(@os_name, @os_version, target_repo, 'SRPMS')
+      def source_output_dir(target_repo = nil)
+        if target_repo.nil?
+          @source_output_dir ||= "#{@os_name}-#{@os_version}-srpms"
+        else
+          @source_output_dir ||= File.join(@os_name, @os_version, target_repo, 'SRPMS')
```
mwaggett: I thought we were moving to having the target_repo above the OS info.

Gemfile
```diff
@@ -23,7 +23,7 @@ end
 group(:development, :test) do
   gem 'json'
   gem 'lock_manager', *lock_manager_location_for(ENV['LOCK_MANAGER_LOCATION'] || '>= 0')
-  gem 'packaging', github: 'puppetlabs/packaging', branch: 'master'
+  gem 'packaging', github: 'puppetlabs/packaging', branch: '1.0.x'
```
mwaggett: Is the 1.0.x branch ready for use?

lib/vanagon/platform.rb
```diff
@@ -237,8 +237,12 @@ def [](key)
     # @param target_repo [String] optional repo target for built packages defined
     #   at the project level
     # @return [String] relative path to where packages should be output to
-    def output_dir(target_repo = "")
-      @output_dir ||= File.join(@os_name, @os_version, target_repo, @architecture)
+    def output_dir(target_repo = nil)
+      if target_repo.nil?
+        @output_dir ||= "#{@os_name}-#{@os_version}-#{@architecture}"
+      else
+        @output_dir ||= File.join(@os_name, @os_version, target_repo, @architecture)
```
mwaggett: I thought we were moving to having the target_repo above the OS info.

melissa: This is for backward compatibility.

lib/vanagon/platform/rpm.rb
melissa: This is for backward compatibility.

Gemfile
melissa: In this case, yes. We've been using it, haven't we?

mwaggett: yes, but with some hackery and only for puppet5, as far as I'm aware.

lib/vanagon/platform.rb
mwaggett: so theoretically, a project would specify the appropriate output_dir?

Gemfile
melissa: Not really. Yes, there's puppet5 specific stuff, but a lot of it is backward compatible. See the comment I made on this PR earlier today.

lib/vanagon/platform.rb
melissa: No. If a project needs to continue building with the old paths, they simple have to specify `proj.target_repo='pc1'` in the project config. That will build and stage the artifacts in the exact same format we've always been building and staging projects. See the comment on this PR I made earlier today. If a user needed to further modify the directory built packages are staged in, then yes, they could set `output_dir`.

melissa: If, say, a project hasn't updated the beaker version they are using, they could still update to this change of vanagon, keep everything else the same, and it would all still work just as it always has.

mwaggett: yep, okay, I see now. I thought I had read the comments before starting to look at the code, but I think I mixed up which PRs I had read. Sorry about that!

mwaggett:
@melissa your links for builds with and without target_repo set are the same

melissa:
So they are. Updated

lib/vanagon/platform/rpm.rb
melissa: If I recall correctly, there are already pe projects that either have `target_repo` set to `pc1` or empty. So we already have a variety of path layouts to deal with here. How do we handle that currently? I can investigate.

underscorgan: I think the issue is that if target_repo isn't set (it isn't for all of them) it's enterprise-dist is hardcoded to look for it in 'products'

spec/lib/vanagon/platform_spec.rb
underscorgan: \o/

lib/vanagon/platform/rpm.rb
underscorgan: But yeah, I think we can audit existing PE vanagon projects (there aren't that many) to make sure target_repo is set to *something*, it's probably a bug not a feature that those ended up in products before :)

melissa: OR, get this, we could also fix enterprise-dist....

melissa: That might be too much effort though

PR #498, opened 2017-09-05
(VANAGON-59) Add support for RPM and Debian Triggers

sarameisburger:
Adding support for triggers in Vanagon

### DIFF
--- a/lib/vanagon/component.rb
+++ b/lib/vanagon/component.rb
@@ -92,6 +92,15 @@ class Vanagon
     # preinstall_actions is a two-dimensional Array, describing scripts that
     # should be executed before a given component is installed.
     attr_accessor :preinstall_actions
+    # install_triggers is a one-dimensional Array of OpenStructs, describing scripts that
+    # should be executed when a package is installed or upgraded
+    attr_accessor :install_triggers
+    # interest_triggers is a one-dimensional Array of OpenStructs, describing scripts that
+    # should be executed when a package identifies an interest trigger
+    attr_accessor :interest_triggers
+    # activate_triggers is a one-dimentional Array of Strings, describing scripts that
+    # should be executed when a package identifies an activate trigger
+    attr_accessor :activate_triggers
     # postinstall_actions is a two-dimensional Array, describing scripts that
     # should be executed after a given component is installed.
     attr_accessor :postinstall_actions
@@ -152,6 +161,9 @@ class Vanagon
       @environment = Vanagon::Environment.new
       @sources = []
       @preinstall_actions = []
+      @install_triggers = []
+      @interest_triggers = []
+      @activate_triggers = []
       @postinstall_actions = []
       @preremove_actions = []
       @postremove_actions = []
--- a/lib/vanagon/component/dsl.rb
+++ b/lib/vanagon/component/dsl.rb
@@ -416,19 +416,59 @@ class Vanagon
         eos
       end
 
+      # Checks that the array of pkg_state is valid (install AND/OR upgrade).
+      # Returns vanagon error if invalid
+      #
+      # @param pkg_state [Array] array of pkg_state input to test
+      def check_pkg_state_array(pkg_state)
+        if pkg_state.empty? || (pkg_state - ["install", "upgrade"]).any?
+          raise Vanagon::Error, "#{pkg_state} should be an array containing one or more of ['install', 'upgrade']"
+        end
+      end
+
       # Adds action to run during the preinstall phase of packaging
       #
-      # @param pkg_state [Array] the state in which the scripts should execute. Can be
-      #   one or multiple of 'install' and 'upgrade'.
+      # @param pkg_state [String, Array] the package state during which the scripts should execute.
+      #   Accepts either a single string ("install" or "upgrade"), or an Array of Strings (["install", "upgrade"]).
       # @param scripts [Array] the Bourne shell compatible scriptlet(s) to execute
       def add_preinstall_action(pkg_state, scripts)
         pkg_state = Array(pkg_state)
         scripts = Array(scripts)
-
-        if pkg_state.empty? || !(pkg_state - ["install", "upgrade"]).empty?
-          raise Vanagon::Error, "#{pkg_state} should be an array containing one or more of ['install', 'upgrade']"
-        end
+        check_pkg_state_array(pkg_state)
         @component.preinstall_actions << OpenStruct.new(:pkg_state => pkg_state, :scripts => scripts)
+      end
+
+      # Adds trigger for scripts to be run on specified pkg_state.
+      #
+      # @param pkg_state [String, Array] the package state during which the scripts should execute.
+      #   Accepts either a single string ("install" or "upgrade"), or an Array of Strings (["install", "upgrade"]).
+      # @param scripts [Array] the rpm pkg scriptlet(s) to execute
+      # @param pkg [String] the package the trigger will be set in
+      def add_rpm_install_triggers(pkg_state, scripts, pkg)
+        pkg_state = Array(pkg_state)
+        scripts = Array(scripts)
+        check_pkg_state_array(pkg_state)
+        @component.install_triggers << OpenStruct.new(:pkg_state => pkg_state, :scripts => scripts, :pkg => pkg)
+      end
+
+      # Adds interest trigger based on the specified packaging state and interest name.
+      #
+      # @param pkg_state [String, Array] the package state during which the scripts should execute.
+      #   Accepts either a single string ("install" or "upgrade"), or an Array of Strings (["install", "upgrade"]).
+      # @param scripts [Array] the scripts to run for the interest trigger
+      # @param interest_name [String] the name of the interest trigger
+      def add_debian_interest_triggers(pkg_state, scripts, interest_name)
+        pkg_state = Array(pkg_state)
+        scripts = Array(scripts)
+        check_pkg_state_array(pkg_state)
+        @component.interest_triggers << OpenStruct.new(:pkg_state => pkg_state, :scripts => scripts, :interest_name => interest_name)
+      end
+
+      # Adds activate trigger name to be watched
+      #
+      # @param activate_name [String] the activate trigger name
+      def add_debian_activate_triggers(activate_name)
+        @component.activate_triggers << OpenStruct.new(:activate_name => activate_name)
       end
 
       # Adds action to run during the postinstall phase of packaging
@@ -439,10 +479,7 @@ class Vanagon
       def add_postinstall_action(pkg_state, scripts)
         pkg_state = Array(pkg_state)
         scripts = Array(scripts)
-
-        if pkg_state.empty? || !(pkg_state - ["install", "upgrade"]).empty?
-          raise Vanagon::Error, "#{pkg_state} should be an array containing one or more of ['install', 'upgrade']"
-        end
+        check_pkg_state_array(pkg_state)
         @component.postinstall_actions << OpenStruct.new(:pkg_state => pkg_state, :scripts => scripts)
       end
 
--- a/lib/vanagon/platform/deb.rb
+++ b/lib/vanagon/platform/deb.rb
@@ -29,9 +29,14 @@ class Vanagon
       # @param workdir [String] working directory to stage the evaluated templates in
       # @param name [String] name of the project
       # @param binding [Binding] binding to use in evaluating the packaging templates
-      def generate_packaging_artifacts(workdir, name, binding)
+      # @param project [Vanagon::Project] Vanagon::Project we are building for
+      def generate_packaging_artifacts(workdir, name, binding, project) # rubocop:disable Metrics/AbcSize
         deb_dir = File.join(workdir, "debian")
         FileUtils.mkdir_p(deb_dir)
+
+        unless project.get_interest_triggers("install").empty? && project.get_interest_triggers("upgrade").empty? && project.get_activate_triggers.empty?
+          erb_file(File.join(VANAGON_ROOT, "resources/deb/triggers.erb"), File.join(deb_dir, "#{name}.triggers"), false, { :binding => binding })
+        end
 
         # Lots of templates here
         ["changelog", "conffiles", "control", "docs", "dirs", "install", "preinst", "postinst", "postrm", "prerm", "rules"].each do |deb_file|
--- a/lib/vanagon/platform/rpm.rb
+++ b/lib/vanagon/platform/rpm.rb
@@ -33,7 +33,8 @@ class Vanagon
       # @param workdir [String] working directory to stage the evaluated templates in
       # @param name [String] name of the project
       # @param binding [Binding] binding to use in evaluating the packaging templates
-      def generate_packaging_artifacts(workdir, name, binding)
+      # @param project [Vanagon::Project] Vanagon::Project we are building for
+      def generate_packaging_artifacts(workdir, name, binding, project)
         erb_file(File.join(VANAGON_ROOT, "resources/rpm/project.spec.erb"), File.join(workdir, "#{name}.spec"), false, { :binding => binding })
       end
 
--- a/lib/vanagon/project.rb
+++ b/lib/vanagon/project.rb
@@ -168,7 +168,7 @@ class Vanagon
     def get_files
       files = []
       files.push @version_file if @version_file
-      files.push @components.map(&:files).flatten
+      files.push components.flat_map(&:files)
       files.flatten.uniq
     end
 
@@ -210,7 +210,7 @@ class Vanagon
     # @return [Array] array of runtime requirements for the project
     def get_requires
       req = []
-      req << @components.map(&:requires).flatten
+      req << components.flat_map(&:requires)
       req << @requires
       req.flatten.uniq
     end
@@ -221,7 +221,7 @@ class Vanagon
     def get_replaces
       replaces = []
       replaces.push @replaces.flatten
-      replaces.push @components.map(&:replaces).flatten
+      replaces.push components.flat_map(&:replaces)
       replaces.flatten.uniq
     end
 
@@ -231,7 +231,7 @@ class Vanagon
 
     # Collects all of the conflicts for the project and its components
     def get_conflicts
-      conflicts = @components.flat_map(&:conflicts) + @conflicts
+      conflicts = components.flat_map(&:conflicts) + @conflicts
       # Mash the whole thing down into a flat Array
       conflicts.flatten.uniq
     end
@@ -248,7 +248,7 @@ class Vanagon
     # @param [string] name of service to grab
     # @return [@component.service obj] specific service
     def get_service(name)
-      @components.each do |component|
+      components.each do |component|
         if component.name == name
           return component.service
         end
@@ -262,7 +262,7 @@ class Vanagon
     def get_provides
       provides = []
       provides.push @provides.flatten
-      provides.push @components.map(&:provides).flatten
+      provides.push components.flat_map(&:provides)
       provides.flatten.uniq
     end
 
@@ -270,15 +270,26 @@ class Vanagon
       !get_provides.empty?
     end
 
-    # Collects the preinstall packaging actions for the project and it's components
-    # for the specified packaging state
+    # Checks that the string pkg_state is valid (install OR upgrade).
+    # Return vanagon error if invalid
+    #
+    # @param pkg_state [String] package state input
+    def check_pkg_state_string(pkg_state)
+      unless ["install", "upgrade"].include? pkg_state
+        raise Vanagon::Error, "#{pkg_state} should be a string containing one of 'install' or 'upgrade'"
+      end
+    end
+
+    # Collects the preinstall packaging actions for the project and its components
+    #  for the specified packaging state
     #
     # @param pkg_state [String] the package state we want to run the given scripts for.
-    #   Can be one or more of 'install' or 'upgrade'
+    #   Can be one of 'install' or 'upgrade'
     # @return [String] string of Bourne shell compatible scriptlets to execute during the preinstall
     #   phase of packaging during the state of the system defined by pkg_state (either install or upgrade)
     def get_preinstall_actions(pkg_state)
-      scripts = @components.map(&:preinstall_actions).flatten.compact.select { |s| s.pkg_state.include? pkg_state }.map(&:scripts)
+      check_pkg_state_string(pkg_state)
+      scripts = components.flat_map(&:preinstall_actions).compact.select { |s| s.pkg_state.include? pkg_state }.map(&:scripts)
       if scripts.empty?
         return ': no preinstall scripts provided'
       else
@@ -286,16 +297,81 @@ class Vanagon
       end
     end
 
-
-    # Collects the postinstall packaging actions for the project and it's components
-    # for the specified packaging state
+    # Collects the install trigger scripts for the project for the specified packing state
     #
     # @param pkg_state [String] the package state we want to run the given scripts for.
-    #   Can be one or more of 'install' or 'upgrade'
+    #   Can be one of the 'install' or 'upgrade'
+    # @return [Hash] of scriptlets to execute during the pkg_state (install or upgrade)
+    #   there can be more than one script for each package (key)
+    def get_trigger_scripts(pkg_state)
+      triggers = Hash.new { |hsh, key| hsh[key] = [] }
+      check_pkg_state_string(pkg_state)
+      pkgs = components.flat_map(&:install_triggers).compact.select { |s| s.pkg_state.include? pkg_state }
+      pkgs.each do |package|
+        triggers[package.pkg].push package.scripts
+      end
+      triggers
+    end
+
+    # Grabs the install trigger scripts for the specified pkg
+    #
+    # @param pkg [String] the pkg we watch for being installed
+    def get_install_trigger_scripts(pkg)
+      scripts = get_trigger_scripts("install")
+      return scripts[pkg].join("\n")
+    end
+
+    # Grabs the upgrade trigger scripts for the specified pkg
+    #
+    # @param pkg [String] the pkg we watch for being upgraded
+    def get_upgrade_trigger_scripts(pkg)
+      scripts = get_trigger_scripts("upgrade")
+      return scripts[pkg].join("\n")
+    end
+
+    # Grabs all pkgs that have trigger scripts for 'install' and 'upgrade'
+    #
+    # @return [Array] a list of all the pkgs that have trigger scripts
+    def get_all_trigger_pkgs()
+      install_triggers = get_trigger_scripts("install")
+      upgrade_triggers = get_trigger_scripts("upgrade")
+      packages = (install_triggers.keys + upgrade_triggers.keys).uniq
+      return packages
+    end
+
+    # Collects the interest triggers for the project and its scripts for the
+    #  specified packaging state
+    #
+    # @param pkg_state [String] the package state we want to run the given scripts for.
+    #   Can be one of 'install' or 'upgrade'
+    # @return [Array] of OpenStructs of all interest triggers for the pkg_state
+    #   Use array of openstructs because we need both interest_name and the scripts
+    def get_interest_triggers(pkg_state)
+      interest_triggers = []
+      check_pkg_state_string(pkg_state)
+      interests = components.flat_map(&:interest_triggers).compact.select { |s| s.pkg_state.include? pkg_state }
+      interests.each do |interest|
+        interest_triggers.push(interest)
+      end
+      interest_triggers.flatten.compact
+    end
+
+    # Collects activate triggers for the project and its components
+    #
+    # @return [Array] of activate triggers
+    def get_activate_triggers()
+      components.flat_map(&:activate_triggers).compact.map(&:activate_name)
+    end
+
+    # Collects the postinstall packaging actions for the project and it's components
+    #  for the specified packaging state
+    #
+    # @param pkg_state [String] the package state we want to run the given scripts for.
+    #   Can be one of 'install' or 'upgrade'
     # @return [String] string of Bourne shell compatible scriptlets to execute during the postinstall
     #   phase of packaging during the state of the system defined by pkg_state (either install or upgrade)
     def get_postinstall_actions(pkg_state)
-      scripts = @components.map(&:postinstall_actions).flatten.compact.select { |s| s.pkg_state.include? pkg_state }.map(&:scripts)
+      scripts = components.flat_map(&:postinstall_actions).compact.select { |s| s.pkg_state.include? pkg_state }.map(&:scripts)
       if scripts.empty?
         return ': no postinstall scripts provided'
       else
@@ -311,7 +387,7 @@ class Vanagon
     # @return [String] string of Bourne shell compatible scriptlets to execute during the preremove
     #   phase of packaging during the state of the system defined by pkg_state (either removal or upgrade)
     def get_preremove_actions(pkg_state)
-      scripts = @components.map(&:preremove_actions).flatten.compact.select { |s| s.pkg_state.include? pkg_state }.map(&:scripts)
+      scripts = components.flat_map(&:preremove_actions).compact.select { |s| s.pkg_state.include? pkg_state }.map(&:scripts)
       if scripts.empty?
         return ': no preremove scripts provided'
       else
@@ -327,7 +403,7 @@ class Vanagon
     # @return [String] string of Bourne shell compatible scriptlets to execute during the postremove
     #   phase of packaging during the state of the system defined by pkg_state (either removal or upgrade)
     def get_postremove_actions(pkg_state)
-      scripts = @components.map(&:postremove_actions).flatten.compact.select { |s| s.pkg_state.include? pkg_state }.map(&:scripts)
+      scripts = components.flat_map(&:postremove_actions).compact.select { |s| s.pkg_state.include? pkg_state }.map(&:scripts)
       if scripts.empty?
         return ': no postremove scripts provided'
       else
@@ -339,7 +415,7 @@ class Vanagon
     #
     # @return [Array] array of configfiles installed by components of the project
     def get_configfiles
-      @components.map(&:configfiles).flatten.uniq
+      components.flat_map(&:configfiles).uniq
     end
 
     def has_configfiles?
@@ -352,7 +428,7 @@ class Vanagon
     def get_directories
       dirs = []
       dirs.push @directories
-      dirs.push @components.map(&:directories).flatten
+      dirs.push components.flat_map(&:directories)
       dirs.flatten.uniq
     end
 
@@ -384,7 +460,7 @@ class Vanagon
     #
     # @return [Array] the services provided by components in the project
     def get_services
-      @components.map(&:service).flatten.compact
+      components.flat_map(&:service).compact
     end
 
     # Simple utility for determining if the components in the project declare
@@ -415,7 +491,7 @@ class Vanagon
     #
     # @return [Array] a listing of component names and versions
     def generate_bill_of_materials
-      @components.map { |comp| "#{comp.name} #{comp.version}" }.sort
+      components.map { |comp| "#{comp.name} #{comp.version}" }.sort
     end
 
     # Method to generate the command to create a tarball of the project
@@ -451,7 +527,7 @@ class Vanagon
     # @param component [Vanagon::Component] component to check for already satisfied build dependencies
     # @return [Array] a list of the build dependencies for the given component that are satisfied by other components in the project
     def list_component_dependencies(component)
-      component.build_requires.select { |dep| @components.map(&:name).include?(dep) }
+      component.build_requires.select { |dep| components.map(&:name).include?(dep) }
     end
 
     # Get the package name for the project on the current platform
@@ -473,7 +549,7 @@ class Vanagon
     #
     # @param workdir [String] workdir to put the packaging files into
     def generate_packaging_artifacts(workdir)
-      @platform.generate_packaging_artifacts(workdir, @name, binding)
+      @platform.generate_packaging_artifacts(workdir, @name, binding, self)
     end
 
     # Generate a json hash which lists all of the dependant components of the
@@ -482,7 +558,7 @@ class Vanagon
     # @return [Hash] where the top level keys are components and their values
     #   are hashes with additional information on the component.
     def generate_dependencies_info
-      @components.each_with_object({}) do |component, hsh|
+      components.each_with_object({}) do |component, hsh|
         hsh.merge!(component.get_dependency_hash)
       end
     end
--- a/resources/deb/postinst.erb
+++ b/resources/deb/postinst.erb
@@ -44,6 +44,21 @@ fi
   <%- end -%>
 <%- end -%>
 
+# Run trigger scripts on install if defined
+if [ $1 = triggered ] && [ -z $2 ]; then
+<%- get_interest_triggers("install").each do |trigger| -%>
+<%= trigger.scripts.join("\n") %>
+<%- end -%>
+fi
+
+if [ $1 = triggered ]&& [ -n $2 ]; then
+<%- get_interest_triggers("upgrade").each do |trigger| -%>
+# Run trigger scripts on upgrade if defined
+  <%= trigger.scripts.join("\n") %>
+<%- end -%>
+fi
+
+
 # Run postinstall scripts on install if defined
 if [ $1 = configure ] && [ -z $2 ] ; then
   <%= get_postinstall_actions("install") %>
--- a/resources/deb/triggers.erb
+++ b/resources/deb/triggers.erb
@@ -1,0 +1,11 @@
+<%- get_activate_triggers.each do |trigger| -%>
+activate <%= trigger %>
+<% end -%>
+
+<%- get_interest_triggers("install").each do |trigger| -%>
+interest <%= trigger.interest_name %>
+<% end -%>
+
+<%-	get_interest_triggers("upgrade").each do |trigger| -%>
+interest <%= trigger.interest_name %>
+<% end -%>
--- a/resources/rpm/project.spec.erb
+++ b/resources/rpm/project.spec.erb
@@ -185,6 +185,22 @@ if [ "$1" -gt 1 ] ; then
 fi
 
 
+<%- get_all_trigger_pkgs.each do |pkg| %>
+
+%triggerin -- <%= pkg %>
+# Run scripts on install if defined
+[ $2 = 0 ] || exit 0
+if [ "$1" -eq 1 ] ; then
+  <%= get_install_trigger_scripts(pkg) %>
+  : # end line
+fi
+elif ["$1" -gt 1 ] ;
+  <%= get_upgrade_trigger_scripts(pkg) %>
+  : # end line
+fi
+<% end -%>
+
+
 %post
 <%- if @platform.is_aix? || (@platform.is_el? && @platform.os_version.to_i == 4) -%>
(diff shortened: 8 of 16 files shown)


sarameisburger:
[puppet-agent.triggers.txt](https://github.com/puppetlabs/vanagon/files/1278548/puppet-agent.triggers.txt)
[postinst.txt](https://github.com/puppetlabs/vanagon/files/1278551/postinst.txt)
[puppet-agent.spec.txt](https://github.com/puppetlabs/vanagon/files/1278552/puppet-agent.spec.txt)

lib/vanagon/component/dsl.rb
```diff
@@ -424,13 +424,52 @@ def environment(*env) # rubocop:disable Metrics/AbcSize
...
+      # Adds trigger to run during the install phase of packaging
+      #
+      # @param pkg_state [Array] the state in which the scripts should execute. Can be
+      #   one or multiple of 'install' and 'upgrade'.
+      # @param scripts [Array] the Bourne shell compatible scriptlet(s) to execute
+      # @param pkg [String] the package the trigger will be set in
+      def add_install_triggers(pkg_state, scripts, pkg)
+        pkg_state = Array(pkg_state)
+        scripts = Array(scripts)
+        pkg = String(pkg)
```
mwaggett: Are these type coercions necessary?

lib/vanagon/component/dsl.rb
```diff
@@ -424,13 +424,52 @@ def environment(*env) # rubocop:disable Metrics/AbcSize
...
+        if pkg_state.empty? || !(pkg_state - ["install", "upgrade"]).empty?
+          raise Vanagon::Error, "#{pkg_state} should be an array containing one or more of ['install', 'upgrade']"
+        end
+        @component.install_triggers << OpenStruct.new(:pkg_state => pkg_state, :scripts => scripts, :pkg => pkg)
```
mwaggett: I think `{ :pkg_state => pkg_state, :scripts => scripts, :pkg => pkg }` is more idiomatic.

lib/vanagon/component/dsl.rb
```diff
@@ -424,13 +424,52 @@ def environment(*env) # rubocop:disable Metrics/AbcSize
...
+      # @param interest_name String the name of the interest trigger
+      def add_debian_interest_triggers(pkg_state, scripts, interest_name)
+        pkg_state = Array(pkg_state)
+        scripts = Array(scripts)
+        interest_name = String(interest_name)
+        if pkg_state.empty? || !(pkg_state - ["install", "upgrade"]).empty?
+          raise Vanagon::Error, "#{pkg_state} should be an array containing one or more of ['install', 'upgrade']"
+        end
+        @component.interest_triggers << OpenStruct.new(:pkg_state => pkg_state, :scripts => scripts, :interest_name => interest_name)
+      end
```
mwaggett: same comments as for `add_install_triggers`

lib/vanagon/component/dsl.rb
```diff
@@ -424,13 +424,52 @@ def environment(*env) # rubocop:disable Metrics/AbcSize
...
+
+      # Adds activate trigger name to be watched
+      #
+      # @param activate_name String the activate trigger name
+      def add_debian_activate_triggers(activate_name)
+        activate_name = String(activate_name)
+        @component.activate_triggers << OpenStruct.new(:activate_name => activate_name)
+      end
```
mwaggett: same comments as for `add_install_triggers`

lib/vanagon/platform/rpm.rb
```diff
@@ -33,7 +33,8 @@ def generate_package(project) # rubocop:disable Metrics/AbcSize
       # @param workdir [String] working directory to stage the evaluated templates in
       # @param name [String] name of the project
       # @param binding [Binding] binding to use in evaluating the packaging templates
-      def generate_packaging_artifacts(workdir, name, binding)
+      # @param project [String] project used for trigger scripts
+      def generate_packaging_artifacts(workdir, name, binding, project)
         erb_file(File.join(VANAGON_ROOT, "resources/rpm/project.spec.erb"), File.join(workdir, "#{name}.spec"), false, { :binding => binding })
```
mwaggett: Did you mean for `project` here to be the `project` param? If so, should be `#{project}`. If not, what is the `project` param used for?

lib/vanagon/project.rb
```diff
@@ -270,22 +270,75 @@ def has_provides?
       !get_provides.empty?
     end
 
-    # Collects the preinstall packaging actions for the project and it's components
-    # for the specified packaging state
+    # Collects the preinstall packaging actions for the project and its components
```
mwaggett: 💯

lib/vanagon/project.rb
```diff
@@ -270,22 +270,75 @@ def has_provides?
...
+    #   for the specified packaging state
```
mwaggett: *super nitpick, so feel free to ignore, but* this space is not really necessary at all

lib/vanagon/project.rb
```diff
@@ -270,22 +270,75 @@ def has_provides?
...
     #
     # @param pkg_state [String] the package state we want to run the given scripts for.
     #   Can be one or more of 'install' or 'upgrade'
     # @return [String] string of Bourne shell compatible scriptlets to execute during the preinstall
     #   phase of packaging during the state of the system defined by pkg_state (either install or upgrade)
-    def get_preinstall_actions(pkg_state)
-      scripts = @components.map(&:preinstall_actions).flatten.compact.select { |s| s.pkg_state.include? pkg_state }.map(&:scripts)
-      if scripts.empty?
+    def get_preinstall_actions(pkg_state) # rubocop:disable Metrics/AbcSize
```
sarameisburger: Found a better way to do this so this method will be changing to something a lot more simple

lib/vanagon/component/dsl.rb
underscorgan: using OpenStruct is consistent with how we've done this elsewhere in this file

lib/vanagon/component/dsl.rb
underscorgan: We may be able to get rid of the string coercion for pkg, but for pkg_state and scripts we want to coerce input into an array, whether passed an array or a string

lib/vanagon/component/dsl.rb
```diff
@@ -424,13 +424,52 @@ def environment(*env) # rubocop:disable Metrics/AbcSize
...
+      #   one or multiple of 'install' and 'upgrade'
+      # @param scripts [Array] the scripts to run for the interest trigger
+      # @param interest_name String the name of the interest trigger
+      def add_debian_interest_triggers(pkg_state, scripts, interest_name)
+        pkg_state = Array(pkg_state)
+        scripts = Array(scripts)
+        interest_name = String(interest_name)
+        if pkg_state.empty? || !(pkg_state - ["install", "upgrade"]).empty?
+          raise Vanagon::Error, "#{pkg_state} should be an array containing one or more of ['install', 'upgrade']"
+        end
```
mwaggett: Since you do this check in a bunch of places, it may be useful to separate this out into its own function, maybe `valid_pkg_state?`?

lib/vanagon/project.rb
sarameisburger: Jk, I lied.

lib/vanagon/platform/rpm.rb
melissa: This has always been `project.spec.erb`. It's the name of the template before it is parsed. It will be renamed to `#{name}.spec` after the template is parsed.

lib/vanagon/component/dsl.rb
sarameisburger: Yeah good call, I removed the string coercions

mwaggett: 👍

lib/vanagon/component/dsl.rb
sarameisburger: Ohhh yes I like that

lib/vanagon/component/dsl.rb
mwaggett: ooh

lib/vanagon/platform/rpm.rb
mwaggett: My question about what the `project` param is used for still holds.

underscorgan: It's only used for deb packaging but was added to the method signature in all the places since it gets overridden

lib/vanagon/component/dsl.rb
mckern: This is a reasonable workaround but it doesn't correspond to your yardoc for the method signature. If the method is public (will map to the Component DSL) and it accepts Array or String for those first two values, the doc should probably say so.

lib/vanagon/component/dsl.rb
```diff
@@ -424,13 +424,52 @@ def environment(*env) # rubocop:disable Metrics/AbcSize
...
+      #
+      # @param pkg_state [Array] the state in which the scripts should execute. Can be
+      #   one or multiple of 'install' and 'upgrade'.
+      # @param scripts [Array] the Bourne shell compatible scriptlet(s) to execute
+      # @param pkg [String] the package the trigger will be set in
+      def add_install_triggers(pkg_state, scripts, pkg)
+        pkg_state = Array(pkg_state)
+        scripts = Array(scripts)
+        pkg = String(pkg)
+        if pkg_state.empty? || !(pkg_state - ["install", "upgrade"]).empty?
```
mckern: Would `(pkg_state & %w[install upgrade]).any?` satisfy your needs without the boolean inversion/coercion?

lib/vanagon/component/dsl.rb
mckern: Nitpick, don't block on it: I know we do this everywhere else here, but this should probably be using the accessor (`_component`) instead of the instance variable (`@component`).

lib/vanagon/component/dsl.rb
mckern: Yeah, `#add_install_triggers` and `add_debian_interest_triggers` are like 95% identical; that commonality could probably be plucked out into a private method that they both use, and the remaining operation would be what each method brings to the party. Is that what you were getting at, @mwaggett?

lib/vanagon/project.rb
```diff
@@ -270,22 +270,75 @@ def has_provides?
...
+      pkg_state = Array(pkg_state)
+      preinstall_scripts = Array.new
```
mckern: `[]` is preferred to `Array.new` for... _reasons_?

lib/vanagon/project.rb
```diff
@@ -270,22 +270,75 @@ def has_provides?
...
+      if pkg_state.empty? || !(pkg_state - ["install", "upgrade"]).empty?
+        raise Vanagon::Error, "#{pkg_state} should be an array containing one or more of ['install', 'upgrade']"
+      end
+      preinstalls = @components.map(&:preinstall_actions).flatten.compact.select { |s| s.pkg_state == pkg_state }
```
mckern: Will `components.flat_map(&:preinstall_actions).select { |s| s.pkg_state == pkg_state }` work here?

(If you're calling `#compact`, `#flatten`, and `#select`, that's a light code smell. I don't know what your data structure looks like offhand, so I might be wrong. Test it (if you can do it quickly) and see?

sarameisburger: Looks like it works with the change!

lib/vanagon/component/dsl.rb
mwaggett: I was talking about checking that `pkg_state` contains the appropriate values, since that's used in a handful of these methods.
But yeah, I'm in favor of DRYing things up in all the ways. 😊

lib/vanagon/project.rb
mwaggett: +1

underscorgan: but they don't work the same
```
irb(main):001:0> foo = ['a']
=> ["a"]
irb(main):002:0> [foo]
=> [["a"]]
irb(main):003:0> Array(foo)
=> ["a"]
```

underscorgan: oh, woops. ignore me. Thought this was about the type coerscion.

lib/vanagon/component/dsl.rb
mckern: Some variation of `pkg_state - ["install", "upgrade"]).empty?` is used in trigger definitions a lot throughout this PR. That and the coercion of various inputs to these methods indicates that maybe they could be extracted out into their own data structure (maybe even a module or a class, not conceptually dissimilar to Environment variables) that handles validation. Maybe not for this PR if there's a need for this sooner rather than later, but it's something to think about.

lib/vanagon/component/dsl.rb
mckern: Sorry, I meant `#empty?`, not `#any?`; comment updated

lib/vanagon/project.rb
mckern: I'm kind of goofing around here, but does this method work if rewritten as follows? It passes tests but you're closer to this problem than I am.

```ruby
def get_preinstall_actions(pkg_state) # rubocop:disable Metrics/AbcSize
  pkg_state = Array(pkg_state)

  if pkg_state.empty? || (pkg_state & %w[install upgrade]).any?
    raise Vanagon::Error,
          "#{pkg_state} should be an array containing one or more of ['install', 'upgrade']"
  end

  preinstall_scripts = components.each_with_object([]) do |component, arr|
    arr.push(component.preinstall_actions.select { |action| action.pkg_state == pkg_state })
  end

  return ': no preinstall scripts provided' if preinstall_scripts.empty?
  preinstall_scripts.flatten.compact.join("\n")
end
```

I tried to get the ABC size down (I got it to 16.4 by reducing intermediate assignment and 15.33 by removing coercion; was aiming for a straight 15 but c'est la vie), but I may have rewritten it too much. Anyway, I wanted to introduce you to Ruby's [`Enumerable#each_with_object`](https://ruby-doc.org/core-2.2.0/Enumerable.html) method. If I wrote this right, then this will allow you to cut down on passes over the first-level Enumerable object (hash, array, set, etc.) that is `#components`/`@components`.

mckern: Oh. I can get the ABC score down to/below 15 if I yank the `#compact` and `#flatten` off the final return. Comment updated. Still not super-sure this works right -- the test case doesn't seem to completely exercise this method.

lib/vanagon/project.rb
```diff
@@ -270,14 +270,25 @@ def has_provides?
...
 
-    # Collects the preinstall packaging actions for the project and it's components
-    # for the specified packaging state
+    # Checks that the pkg_state is valid, return vanagon error if invalid
     #
-    # @param pkg_state [String] the package state we want to run the given scripts for.
+    # @param String pkg_state String of package state input
+    # @param String function the function that called
+    def check_pkg_state(pkg_state, method_name)
+      if pkg_state.empty? || !(["install", "upgrade"].include? pkg_state)
```
mckern: Is `!(["install", "upgrade"].include? pkg_state)` ultimately just comparing String values? I've probably done it that way in Vanagon before too but I think maybe I regret it if I did? Anyway, if this is just a String comparison, then you could simplify it to

```ruby
# case insensitive, because why not?
pkg_state.match? /^upgrade|install$/i
```

lib/vanagon/project.rb
```diff
@@ -286,9 +297,48 @@ def get_preinstall_actions(pkg_state)
       end
     end
 
+    # Collects the install trigger scripts for the project for the specified packing state
+    #
+    # @param pkg_state String the package state we want to run the given scripts for.
+    #  Can be one or more of the 'install' or 'upgrade'
+    # @return {} of scriptlets to execute during the pkg_state (install or upgrade)
+    #   there can be more than one script for each package (key)
```
mckern: The Yardoc is malformed. Try:

```
# @param [String] pkg_state the package state we want to run the given scripts for.
#   Can be one or more of the 'install' or 'upgrade'
# @return [Hash] a collection of one or more scriptlets to execute during the pkg_state (install or upgrade)
```

lib/vanagon/project.rb
```diff
@@ -286,9 +297,48 @@ def get_preinstall_actions(pkg_state)
...
+    def get_install_trigger_scripts(pkg_state)
+      triggers = Hash.new { |hsh, key| hsh[key] = [] }
+      check_pkg_state(pkg_state, __method__)
+      pkgs = @components.map(&:install_triggers).flatten.compact.select { |s| s.pkg_state.include? pkg_state }
+      pkgs.each do |package|
+        triggers[package.pkg].push package.scripts
+      end
+      return triggers
```
mckern: Explicit return is discouraged

lib/vanagon/project.rb
mwaggett: `pkg_state` is an array (that should contain one or both of 'upgrade' and 'install'), so that won't work.

mckern: If `pkg_state` is an Array, then what's there now has some edge cases.

```ruby
>> ["install", "upgrade"].include? pkg_state
=> true
>> pkg_state
=> "upgrade"
>> pkg_state = %w[upgrade]
=> [
  [0] "upgrade"
]
>> pkg_state
=> [
  [0] "upgrade"
]
>> ["install", "upgrade"].include? pkg_state
=> false
>>
```
Will pkg_state ever contain anything but the Strings "install" and "upgrade"?

lib/vanagon/project.rb
mckern: FYI, [this is a handy cheatsheet reference](https://gist.github.com/chetan/1827484)

mckern:
Very broad, but very _strong_ suggestion: `pkg_state` is used as a method argument for a LOT of these methods, but it's also used as an accessor in the OpenStruct that underlies the trigger storage implementation (it's used, in one way or another, 88 times in this PR). This means that a lot of the methods are carrying a pretty high cognitive load for "which of the two `pkg_states` am I working with right now?" when I try to read them -- I'm assuming this is true for other folks too. Consider using `pkg_state_name` or `pkg_state_stage` or `pollywollydoodleallday` for the method arguments; anything but `pkg_state`.

lib/vanagon/component/dsl.rb
```diff
@@ -416,6 +416,16 @@ def environment(*env) # rubocop:disable Metrics/AbcSize
         eos
       end
 
+      # Checks that the pkg_state is valid, returns vanagon error if invalid
+      #
+      # @param pkg_state [Array] array of pkg_state input to test
+      # @param function [String] the function that called
+      def check_pkg_state(pkg_state, method_name)
+        if pkg_state.empty? || !(pkg_state - ["install", "upgrade"]).empty?
```
mckern: You could probably simplify this to a very, very simple `pkg_state.sort == %w[install upgrade]` and skip the `#empty?` and the `||` -- pkg_state **has** to be defined, and it should only ever be an Array; so sorting it and comparing it to the only allowed value should simplify this.

mckern: Note that this will return `nil` in both the original implementation and my suggested revision -- this means it can't be used for direct comparisons or for any truthy operations, nor can it be included in any method chains. Consider this slight revision of my revision, which instead returns the original object using a [guard clause](http://www.rubydoc.info/github/bbatsov/rubocop/Rubocop/Cop/Style/GuardClause), or raises an error:

```ruby
def check_pkg_state(pkg_state, method_name)
  allowed_values = %w[install upgrade]
  return pkg_state unless pkg_state.sort != allowed_values.sort
  raise Vanagon::Error,
        "#{pkg_state} should be an array containing one or more of #{allowed_values} in ##{method_name}"
end
```

mckern: So, reading the doc for this method, it looks like the argument`pkg_state` *must* be an Array, but that it can contain one or both of the allowed values? I didn't catch that before any of this but I think I did notice it the other night. I'd mentioned that `!(pkg_state - %w[install upgrade]).empty?` can be rewritten as `(pkg_state - %w[install upgrade]).any?`. Taking all of the above into account, that same guidance would make my revised revision look like this:

```ruby
def check_pkg_state(pkg_state, method_name)
  allowed_values = %w[install upgrade]
  return pkg_state if (pkg_state - allowed_values.sort).empty?
  raise Vanagon::Error,
        "#{pkg_state} should be an array containing one or more of #{allowed_values} in ##{method_name}"
end
```

mckern: Tangent: that double `##` in the error message in my suggested revs. is intentional -- it'll print out the method name using the standard Ruby convention for an instance method:

```ruby
>> puts "#{pkg_state} should be an array containing one or more of #{allowed_values} in ##{method_name}"
["upgrade", "install"] should be an array containing one or more of ["install", "upgrade"] in #m_e_t_h_o_d_man
=> nil
```

lib/vanagon/project.rb
```diff
@@ -270,14 +270,25 @@ def has_provides?
...
-    # for the specified packaging state
+    # Checks that the pkg_state is valid, return vanagon error if invalid
+    #
+    # @param pkg_state [String] package state input
+    # @param function [String] the function that called
+    def check_pkg_state(pkg_state, method_name)
+      if pkg_state.empty? || !(["install", "upgrade"].include? pkg_state)
+        raise Vanagon::Error, "#{pkg_state} should be a string containing one of 'install' or 'upgrade' in #{method_name}"
+      end
+    end
```
mckern: Using the suggested guidance, this would look like this:

def check_pkg_state(pkg_state, method_name)
  allowed_values = %w[install upgrade]
  return pkg_state if allowed_values.include? pkg_state
  raise Vanagon::Error,
        "#{pkg_state} should be a string containing one of #{allowed_values.join", "} in ##{method_name}"
end
```

You can skip the `#empty?` check, since the `#include?` check will catch that particular use case:

```ruby
>> %w[a b c d e f g].include? ""
=> false
```

lib/vanagon/component.rb
```diff
@@ -92,6 +92,15 @@ class Component
     # preinstall_actions is a two-dimensional Array, describing scripts that
     # should be executed before a given component is installed.
     attr_accessor :preinstall_actions
+    # install_triggers is a two-dimensional Array, describing scripts that
+    # should be executed when a package is installed or upgraded
+    attr_accessor :install_triggers
```
mckern: Isn't `@install_triggers` actually a one-dimensional array of OpenStructs?

lib/vanagon/component/dsl.rb
```diff
@@ -424,13 +434,43 @@ def environment(*env) # rubocop:disable Metrics/AbcSize
...
-          raise Vanagon::Error, "#{pkg_state} should be an array containing one or more of ['install', 'upgrade']"
-        end
+        check_pkg_state(pkg_state, __method__)
         @component.preinstall_actions << OpenStruct.new(:pkg_state => pkg_state, :scripts => scripts)
       end
 
+      # Adds trigger for scripts to be run on specified pkg_state.
+      #
+      # @param pkg_state [String] the state in which the scripts should execute. Can be
+      #   one or multiple of 'install' and 'upgrade'
```
mckern: Yardoc suggestion: you can specify all of the allowed classes for a method argument in Yard. That'd look like this:

```
# @param pkg_state [String, Array] the package state during which the scripts should execute.
#     Accepts either a single string ("install" or "upgrade"), or an Array of Strings (["install", "upgrade"]).
```

lib/vanagon/project.rb
mckern: Oh, I forgot. You can ditch `method_name` as an argument, and just use `__callee__` to ask Ruby for whatever method called `#check_pkg_state`.

lib/vanagon/component.rb
sarameisburger: Yup, good catch, that was a copypasta error

lib/vanagon/project.rb
sarameisburger: __callee__ seems to be returning 'check_pkg_state' but I've decided to just remove that bit instead

lib/vanagon/platform/deb.rb
```diff
@@ -29,10 +29,15 @@ def generate_package(project) # rubocop:disable Metrics/AbcSize
       # @param workdir [String] working directory to stage the evaluated templates in
       # @param name [String] name of the project
       # @param binding [Binding] binding to use in evaluating the packaging templates
-      def generate_packaging_artifacts(workdir, name, binding)
+      # @param project [String] name of the project
```
underscorgan: `project` isn't the name, it's the `Vanagon::Project` that we're building for

lib/vanagon/platform/rpm.rb
```diff
@@ -33,7 +33,8 @@ def generate_package(project) # rubocop:disable Metrics/AbcSize
       # @param workdir [String] working directory to stage the evaluated templates in
       # @param name [String] name of the project
       # @param binding [Binding] binding to use in evaluating the packaging templates
-      def generate_packaging_artifacts(workdir, name, binding)
+      # @param project [String] project used for trigger scripts
```
underscorgan: same comment as above for `project`

lib/vanagon/project.rb
underscorgan: yeah, you should be able to replace this with just `triggers`

lib/vanagon/project.rb
```diff
@@ -270,14 +270,24 @@ def has_provides?
...
-    # Collects the preinstall packaging actions for the project and it's components
-    # for the specified packaging state
+    # Checks that the pkg_state is valid (install OR upgrade), return vanagon error if invalid
+    #
+    # @param pkg_state [String] package state input
+    def check_pkg_state(pkg_state)
+      unless ["install", "upgrade"].include? pkg_state
+        raise Vanagon::Error, "#{pkg_state} should be a string containing one of 'install' or 'upgrade'"
+      end
+    end
```
underscorgan: Looks like there are two separate implementations of this. Wondering if there's somewhere good we can add this more generically so we can define it once and use it in all the places! Maybe in `Vanagon::Utilities`?

lib/vanagon/platform/deb.rb
sarameisburger: Oh whoops!

lib/vanagon/project.rb
sarameisburger: One checks for an array containing 1 or more of install/upgrade, the other checks that the string is either install or upgrade. Would we still want to combine it?

underscorgan: ah, woops. Didn't read the code closely enough. I'm trying to think of a neat way to get these combined still but nothing is coming to mind, so could the methods have different names? I'm not sure the best way to name it unambiguously, but methods for two different things with the same name is a little confusing :)

sarameisburger: Yeah for sure, I'll make that change!

sarameisburger: Would "check_project_pkg_state" and "check_dsl_pkg_state" be alright?

underscorgan: Hm, no, that still seems confusing.

Maybe having `_string` vs `_array` in the names? I think mostly this is confusing for me because we're only checking for a subset of the possible package states (usually things will either happen on (install, upgrade, or install and upgrade) or (removal, upgrade, or removal and upgrade) I don't have any great suggestions for how to make this better though, might need to think on it more.

sarameisburger: Ok cool cool, I'll change those names and if anything I can always make a ticket to combine those later when we've had more time to think about it

lib/vanagon/platform/deb.rb
```diff
@@ -29,10 +29,15 @@ def generate_package(project) # rubocop:disable Metrics/AbcSize
       # @param workdir [String] working directory to stage the evaluated templates in
       # @param name [String] name of the project
       # @param binding [Binding] binding to use in evaluating the packaging templates
-      def generate_packaging_artifacts(workdir, name, binding)
+      # @param project [String] Vanagon::Project we are building for
+      def generate_packaging_artifacts(workdir, name, binding, project) # rubocop:disable Metrics/AbcSize
```
underscorgan: You'll need to update this in a few more places to make sure all the generate_packaging_artifacts include the project

```
lib/vanagon/platform/osx.rb
67:      def generate_packaging_artifacts(workdir, name, binding) # rubocop:disable Metrics/AbcSize

lib/vanagon/platform/solaris_11.rb
43:      def generate_packaging_artifacts(workdir, name, binding)

lib/vanagon/platform/solaris_10.rb
84:      def generate_packaging_artifacts(workdir, name, binding)

lib/vanagon/platform/windows.rb
39:      def generate_packaging_artifacts(workdir, name, binding)

File: lib/vanagon/project.rb [hunk 24]
```diff
@@ -286,9 +297,48 @@ def get_preinstall_actions(pkg_state)
       end
     end
 
+    # Collects the install trigger scripts for the project for the specified packing state
+    #
+    # @param pkg_state [String] the package state we want to run the given scripts for.
+    #   Can be one or more of the 'install' or 'upgrade'
```
underscorgan: Since this is a string now should probably get rid of the `or more`

sarameisburger: Whoops yeah, forgot to change the comment

lib/vanagon/platform/deb.rb
sarameisburger: Can we make it an optional parameter?

sarameisburger: Not super great with Ruby so not sure if that would be possible/a good idea

underscorgan: So, ruby has optional parameters but I'm not sure it's going to buy us anything here.

There are two different `generate_packaging_artifacts` right now: one defined in `Vanagon::Project`, which is the method that's initially called when vanagon runs. That method calls the `generate_packaging_artifacts` for the `Vanagon::Platform`. There's no generic `generate_packaging_artifacts`, only the `generate_packaging_artifacts` methods defined for each platform. So, if we don't update all of the methods it won't be able to find it for anything other than rpms and debs

sarameisburger: ohhh gotcha gotcha

lib/vanagon/project.rb
```diff
@@ -286,12 +297,51 @@ def get_preinstall_actions(pkg_state)
...
+    # Collects the install trigger scripts for the project for the specified packing state
+    #
+    # @param pkg_state [String] the package state we want to run the given scripts for.
+    #   Can be one of the 'install' or 'upgrade'
+    # @return [Hash] of scriptlets to execute during the pkg_state (install or upgrade)
+    #   there can be more than one script for each package (key)
+    def get_install_trigger_scripts(pkg_state)
+      triggers = Hash.new { |hsh, key| hsh[key] = [] }
+      check_pkg_state_string(pkg_state)
+      pkgs = @components.map(&:install_triggers).flatten.compact.select { |s| s.pkg_state.include? pkg_state }
```
mckern: Suggestion: Use `#flat_map` and the `#components` accessor instead of `#map.flatten` and `@components`:

```ruby
pkgs = components.flat_map(&:install_triggers).compact.select { |s| s.pkg_state.include? pkg_state }
```

lib/vanagon/project.rb
```diff
@@ -286,12 +297,51 @@ def get_preinstall_actions(pkg_state)
...
+    #  specified packaging state
+    #
+    # @param pkg_state [String] the package state we want to run the given scripts for.
+    #   Can be one of 'install' or 'upgrade'
+    # @return [Array] of OpenStructs of all interest triggers for the pkg_state
+    #   Use array of openstructs because we need both interest_name and the scripts
+    def get_interest_triggers(pkg_state)
+      interest_triggers = []
+      check_pkg_state_string(pkg_state)
+      interests = @components.map(&:interest_triggers).flatten.compact.select { |s| s.pkg_state.include? pkg_state }
```
mckern: Same as above: `#flat_map` and use accessor

resources/deb/postinst.erb
```diff
@@ -44,6 +44,27 @@ fi
   <%- end -%>
 <%- end -%>
 
+# adding support for triggers
+<%- install_triggers = get_interest_triggers("install").uniq
+  upgrade_triggers = get_interest_triggers("upgrade").uniq -%>
```
mckern: Can the logic from lines 48 and 49 be moved out of the ERB and into methods in one of the underlying classes? Logic beyond "if/else" in an ERB is a code smell.

resources/deb/triggers.erb
```diff
@@ -0,0 +1,17 @@
...
+<%- end -%>
+
+<%- unless get_interest_triggers("install").empty? -%>
+<%- get_interest_triggers("install").each do |trigger| -%>
+interest <%= trigger.interest_name %>
+<% end -%>
+<%- end -%>
+
+<%- unless get_interest_triggers("upgrade").empty? -%>
+<%-	get_interest_triggers("upgrade").each do |trigger| -%>
```
mckern: If `#get_interest_triggers` returns an empty Iterable object (like an empty array), then you can lose the `get_interest_triggers("upgrade").empty?` test because Ruby (and therefore ERB) will happily skip over doing anything inside the loop if the Iterable object is empty.

resources/rpm/project.spec.erb
```diff
@@ -184,6 +184,26 @@ if [ "$1" -gt 1 ] ; then
   <%= get_preinstall_actions("upgrade") %>
 fi
 
+<%- install_triggers = get_install_trigger_scripts("install")
+ upgrade_triggers = get_install_trigger_scripts("upgrade")
+ -%>
```
mckern: Previous comment about moving the logic out of the ERB if possible (it's usually possible)

resources/rpm/project.spec.erb
```diff
@@ -184,6 +184,26 @@ if [ "$1" -gt 1 ] ; then
...
+<%- unless install_triggers.empty? && upgrade_triggers.empty? -%>
```
mckern: Since this is a compound statement, consider using:

```ruby
if install_triggers.any? && upgrade_triggers.any?
```

mckern:
I left some more comments and suggestions but please don't feel like you should block on them if you need this *now*.

resources/deb/postinst.erb
underscorgan: Is it the `uniq` you want moved out? I think that could be moved into the `get_interest_triggers` method, but mostly this is just setting variables that we can use later on to save calling this method every time, which seems ok to me :woman_shrugging:

mckern: No, I mean that the assignment should move out of the ERB. ERB files are templates, so they shouldn't have assignment in them -- ideally they're just mapping data from an instance of something to a human-readable format. In this case by using memoization, the assignment could be shoved out of the template and into the underlying class with (I believe) the same runtime cost (comments inline):

```ruby
class Vanagon
  class Project
    # Renamed to a bang (!) method, because it's always live and it never caches
    def get_interest_triggers!(pkg_state)
      interest_triggers = []
      check_pkg_state_string(pkg_state)
      components.flat_map(&:interest_triggers).compact.select { |s| s.pkg_state.include? pkg_state }
      interests.each do |interest|
        interest_triggers.push(interest)
      end
      # I am of the opinion that this many transformations chained together
      # (6 as of last count) is a code smell -- maybe flag this for future
      # refactoring in the comments?
      interest_triggers.flatten.uniq.compact
    end

    def get_interest_triggers(pkg_state)
      # Ensure that @interest_triggers is always a Hash
      @interest_triggers ||= {}
      # Now, check the hash to see if we've cached the computed triggers for
      # a given value of pkg_state (hashed for efficiency); if it has been,
      # then just return it because we don't need to keep building that Hash
      return @interest_triggers if @interest_triggers[pkg_state.hash]
      # Otherwise, assign the computed list of triggers with the hashed
      # value of pkg_state as a key; I believe this should return the
      # expected computed interest_triggers
      @interest_triggers[pkg_state.hash] = get_interest_triggers!(pkg_state)
    end

    def get_install_triggers
      get_interest_triggers("install")
    end

    def get_upgrade_triggers
      get_interest_triggers("upgrade")
    end
  end
end
```

Then `#get_install_triggers` and `#get_upgrade_triggers` can be called anywhere in the ERB template without concern for the cost of computing them or the need to worry about state or assignment in the ERB template itself.

mckern: (`#get_install_triggers` and `#get_upgrade_triggers` could be renamed to just `#install_triggers`/`#upgrade_triggers`and that'd be a drop-in for the variables as named now, I think)

resources/deb/triggers.erb
```diff
@@ -0,0 +1,17 @@
+<%- unless get_activate_triggers.empty? -%>
```
demophoon: I don't think we need to test if this is empty since `for-each`'ing over an empty array should have the same effect.

demophoon: oh.. if I had scrolled down I would have seen Ryan's comment.

resources/deb/triggers.erb
mckern: @demophoon ^^^

sarameisburger:
*I need to update the spec tests*

lib/vanagon/component.rb
```diff
@@ -92,6 +92,15 @@ class Component
...
     attr_accessor :preinstall_actions
+    # install_triggers is a one-dimensional Array of OpenStructs, describing scripts that
+    # should be executed when a package is installed or upgraded
+    attr_accessor :install_triggers
+    # interest_triggers is a one-dimensional Array of OpenStructs, describing scripts that
+    # should be executed when a package identifies an interest trigger
+    attr_accessor :interest_triggers
+    # activate_triggers is a one-dimentional Array of Strings, describing scripts that
+    # should be executed when a package identifies an activate trigger
+    attr_accessor :activate_triggers
```
mwaggett: Why is this an Array of Strings while the others are Arrays of OpenStructs?

lib/vanagon/component/dsl.rb
```diff
@@ -416,21 +416,68 @@ def environment(*env) # rubocop:disable Metrics/AbcSize
...
+      # Adds activate trigger name to be watched
+      #
+      # @param activate_name [String] the activate trigger name
+      def add_debian_activate_triggers(activate_name)
+        @component.activate_triggers << OpenStruct.new(:activate_name => activate_name)
+      end
+
+      def get_all_trigger_pkgs()
+        install_triggers = add_debian_interest_triggers("install")
+        upgrade_triggers = add_debian_interest_triggers("upgrade")
```
mwaggett: Doesn't `add_debian_interest_triggers` need three params?

lib/vanagon/component/dsl.rb
```diff
@@ -416,21 +416,68 @@ def environment(*env) # rubocop:disable Metrics/AbcSize
...
+      end
+
+      # Adds activate trigger name to be watched
+      #
+      # @param activate_name [String] the activate trigger name
+      def add_debian_activate_triggers(activate_name)
+        @component.activate_triggers << OpenStruct.new(:activate_name => activate_name)
+      end
+
+      def get_all_trigger_pkgs()
```
mwaggett: If this is only for debian, I feel like the function name should reflect that.

lib/vanagon/component/dsl.rb
```diff
@@ -416,21 +416,68 @@ def environment(*env) # rubocop:disable Metrics/AbcSize
...
         @component.preinstall_actions << OpenStruct.new(:pkg_state => pkg_state, :scripts => scripts)
       end
 
+      # Adds trigger for scripts to be run on specified pkg_state.
+      #
+      # @param pkg_state [String, Array] the package state during which the scripts should execute.
+      #   Accepts either a single string ("install" or "upgrade"), or an Array of Strings (["install", "upgrade"]).
+      # @param scripts [Array] the rpm pkg scriptlet(s) to execute
+      # @param pkg [String] the package the trigger will be set in
+      def add_install_rpm_triggers(pkg_state, scripts, pkg)
```
mwaggett: I would call this `add_rpm_install_triggers` to be consistent with how you've named debian things below.

lib/vanagon/component.rb
sarameisburger: Activate triggers are names used to track packages while interest triggers have names and scripts we want to run. This page explains it better than I do https://manpages.debian.org/jessie/dpkg-dev/deb-triggers.5.en.html

lib/vanagon/component/dsl.rb
sarameisburger: Yes! Thank you! I called the wrong method when I was renaming stuff

sarameisburger: Oh actually I meant to delete that method, I was wondering why tests still passed haha

lib/vanagon/platform/deb.rb
```diff
@@ -29,10 +29,15 @@ def generate_package(project) # rubocop:disable Metrics/AbcSize
       # @param workdir [String] working directory to stage the evaluated templates in
       # @param name [String] name of the project
       # @param binding [Binding] binding to use in evaluating the packaging templates
-      def generate_packaging_artifacts(workdir, name, binding)
+      # @param project [String] Vanagon::Project we are building for
```
mwaggett: I'm unclear on what the difference between `name` and `project` is.

lib/vanagon/component.rb
mwaggett: ookay! 👍

lib/vanagon/component/dsl.rb
mwaggett: lol

lib/vanagon/platform/deb.rb
sarameisburger: Ok yeah so this is my fault. name is the name of the project and is a string so for example "puppet-agent" while project is the entirety of a Vanagon::Project and contains a bunch of stuff which we set and is of type Vanagon::Project. I was thinking we passed it in as a String which is definitely not right.

mwaggett: ahh, gotcha. Yeah, `Vanagon::Project` seems like the right type. 😊

mwaggett:
100th comment woo!

lib/vanagon/component/dsl.rb
shrug: +1

lib/vanagon/platform/deb.rb
```diff
@@ -29,10 +29,22 @@ def generate_package(project) # rubocop:disable Metrics/AbcSize
...
 
+        puts "the type"
+        puts project.class.inspect
+
+        unless project.get_interest_triggers("install").empty? && project.get_interest_triggers("upgrade").empty? && project.get_activate_triggers.empty?
+          erb_file(File.join(VANAGON_ROOT, "resources/deb/triggers.erb"), File.join(deb_dir, "#{name}.triggers"), false, { :binding => binding })
+          puts "the name"
+          puts name.inspect
+          #puts "the project"
+          #puts project.inspect
```
mwaggett: probably don't want these comments

sarameisburger: Whoops! Good catch

lib/vanagon/platform/deb.rb
```diff
@@ -29,10 +29,18 @@ def generate_package(project) # rubocop:disable Metrics/AbcSize
...
       # @param name [String] name of the project
       # @param binding [Binding] binding to use in evaluating the packaging templates
-      def generate_packaging_artifacts(workdir, name, binding)
+      # @param project [Vanagon::Project] Vanagon::Project we are building for
+      def generate_packaging_artifacts(workdir, name, binding, project) # rubocop:disable Metrics/AbcSize
         deb_dir = File.join(workdir, "debian")
         FileUtils.mkdir_p(deb_dir)
 
+        puts "the type"
+        puts project.class.inspect
```
mwaggett: Did you want to leave these in here too or was this just for debugging?

PR #499, opened 2017-09-29
(VANAGON-109) Fix postinst to not break for empty trigger scripts

sarameisburger:
Bash doesn't like empty if statements


sarameisburger:
[postinst.txt](https://github.com/puppetlabs/vanagon/files/1344939/postinst.txt)

underscorgan:
@sarameisburger have you run a test build with these?

sarameisburger:
Ran a build and nothing broke

PR #500, opened 2017-09-29
VANAGON-76 implement more flexible preserve

e-gris:
I'm bummed about the statefulness of the parser module but needed to preserve it for backward compatibility. Needed the full refactor for flexibility.

### DIFF
--- a/lib/vanagon/driver.rb
+++ b/lib/vanagon/driver.rb
@@ -129,11 +129,15 @@ class Vanagon
       @engine.dispatch("(cd #{@engine.remote_workdir}; #{@platform.make})")
       @engine.retrieve_built_artifact
 
-      unless @preserve
+      if %i[never on-failure].include? @preserve
         @engine.teardown
         cleanup_workdir
       end
     rescue => e
+      if [:never].include? @preserve
+        @engine.teardown
+        cleanup_workdir
+      end
       $stderr.puts e
       $stderr.puts e.backtrace.join("\n")
       raise e
--- a/lib/vanagon/optparse.rb
+++ b/lib/vanagon/optparse.rb
@@ -2,40 +2,86 @@ require 'optparse'
 
 class Vanagon
   class OptParse
-    FLAGS = {
-        :workdir => ['-w DIR', '--workdir DIR', "Working directory where build source should be put (defaults to a tmpdir)"],
-        :"remote-workdir" => ['-r DIR', '--remote-workdir DIR', "Working directory where build source should be put on the remote host (defaults to a tmpdir)"],
-        :configdir => ['-c', '--configdir DIR', 'Configs dir (defaults to $pwd/configs)'],
-        :target => ['-t HOST', '--target HOST', 'Configure a target machine for build and packaging (defaults to grabbing one from the pooler)'],
-        :engine => ['-e ENGINE', '--engine ENGINE', "A custom engine to use (defaults to the pooler) [base, local, docker, pooler currently supported]"],
-        :skipcheck => ['--skipcheck', 'Skip the `check` stage when building components'],
-        :preserve => ['-p', '--preserve', 'Whether to tear down the VM on success or not (defaults to false)'],
-        :verbose => ['-v', '--verbose', 'Verbose output (does nothing)'],
-        :only_build => ["--only-build COMPONENT,COMPONENT,...", Array, 'Only build this array of components']
-      }.freeze
+    def initialize(banner, symbols = []) # rubocop:disable Metrics/AbcSize
+      ## symbols array kept for backward compatibility but ignored
 
-    def initialize(banner, options = [])
-      @options = Hash[options.map { |v| [v, nil] }]
-      @optparse = OptionParser.new do |opts|
+      @options = Hash.new
+      @options[:preserve] = :never
+
+      @option_parser = OptionParser.new do |opts| # rubocop:disable Metrics/BlockLength
         opts.banner = banner
+        opts.separator ""
+        opts.separator "Options:"
 
-        FLAGS.each_pair do |name, args|
-          if @options.include?(name)
-            opts.on(*args) do |value|
-              @options[name] = value
-            end
+        opts.on("-h",
+                "--help",
+                "Display help") do
+          $stdout.puts opts
+          exit 1
+        end
+
+        opts.on("-v",
+                "--verbose",
+                "Verbose output (does nothing)") do |verbose|
+          @options[:verbose] = verbose
+        end
+
+        opts.on("-w DIRECTORY",
+                "--workdir DIRECTORY",
+                "Working directory on the local host (defaults to calling mktemp)") do |workdir|
+          @options[:workdir] = workdir
+        end
+
+        opts.on("-r DIRECTORY",
+                "--remote-workdir DIRECTORY",
+                "Working directory on the remote host (defaults to calling mktemp)") do |remote|
+          @options[:"remote-workdir"] = remote
+        end
+
+        opts.on("-c DIRECTORY",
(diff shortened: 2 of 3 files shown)


lib/vanagon/driver.rb
```diff
@@ -129,18 +129,18 @@ def run # rubocop:disable Metrics/AbcSize
...
+        @engine.teardown
+        cleanup_workdir
+      end
       $stderr.puts e
       $stderr.puts e.backtrace.join("\n")
       raise e
-    ensure
-      if ["hardware", "ec2"].include?(@engine.name)
-        @engine.teardown
-      end
```
underscorgan: I think we want hardware/ec2 platforms to never get preserved so we don't end up leaving these up either blocking other people or costing money.
@McdonaldSeanp might know more though

lib/vanagon/optparse.rb
```diff
@@ -2,40 +2,87 @@
...
-        :preserve => ['-p', '--preserve', 'Whether to tear down the VM on success or not (defaults to false)'],
-        :verbose => ['-v', '--verbose', 'Verbose output (does nothing)'],
-        :only_build => ["--only-build COMPONENT,COMPONENT,...", Array, 'Only build this array of components']
-      }.freeze
-
-    def initialize(banner, options = [])
-      @options = Hash[options.map { |v| [v, nil] }]
-      @optparse = OptionParser.new do |opts|
+    def initialize(banner, symbols = []) # rubocop:disable Metrics/AbcSize
+      ## EXG -- don't understand why symbols array is needed.
```
underscorgan: oh no I think I found what the symbols array is doing ... it looks like it's setting the valid options for each of the subcommands and only adding the option if it's listed as one of the supported symbols

e-gris: Ah ha. That explains the confusing original implementation.

I'm tempted then to monkey-patch #on to take the symbols array as an argument and handle it that way.

underscorgan: I've so far avoided monkey patches in ruby (someone scared me early on in my ruby days), are there non-terrible ways to do this?

e-gris: IMO, there's a balance here between simplicity (allowing the parser to parser whatever, even if it doesn't apply for the subcommand) and the cost of accuracy.

The classic, non-monkey-patch, approach is to wrap or subclass the OptionsParser object. My preferences, in descending order.

1. Opt for simplicity. Let the subcommand ignore stuff not relevant to it.
2. Monkey patch #on. This, IMO, is the textbook use-case for it.
3. Wrap OptionsParser. Major advantage is that this pattern is very easy to recognize and contains the least surprise over monkey patching.
4. Subclass OptionsParser. Uggh.

I'm happy to consider alternate opinions.

underscorgan: Cool, thanks for the explanation! I think option 1 makes a lot of sense, it doesn't seem particularly dangerous to be ignoring some of the options, and a lot of that can be clarified in the README in the future.

e-gris: Of course a further thought -- replacing everything with docopt could be valuable.

```ruby
options[subcommand] = <<~ENDDOC
  <options description for subcommand>
ENDDOC ...

options = Docopt::docopt(options[subcommand])
```

However, this would case a necessary change in all the consumers. For the sake of 1 switch this feels like an even bigger rabbit hole than I've found myself in.

e-gris:
@McdonaldSeanp can you comment about the EC2 teardown? Should it *always* be taken down even if the user requests otherwise?

e-gris:
@MikaelSmith Since this appears to have been originally yours, do you have any thoughts about being less tight about worrying about exactly which options are parsed?

lib/vanagon/driver.rb
McdonaldSeanp: I can't speak to the ec2 engine, but yes the hardware should always be torn down.

MikaelSmith:
I don't remember being involved with this before. Maybe you meant another Mike?

e-gris:
Maybe @murdok5 ??

4cee78b9 (Michael Smith    2015-02-23 17:14:03 -0800 19)       @optparse = OptionParser.new do |opts|

lib/vanagon/optparse.rb
e-gris: As of now, the 'opt-for-simplicity' is the leading option.

mwaggett:
squash commits?

e-gris:
@mwaggett I keep forgetting that that's desirable here. I'll do it after @underscorgan has taken a look.

mwaggett:
@e-gris yeah, I still don't tooootally have a handle on when it's good to squash and when it isn't, but for things like typos, rather than separate improvements, I say it's better to squash - that way if we need to revert or track down a past change, everything that should be together *is* together, ya know?

lib/vanagon/driver.rb
```diff
@@ -129,11 +129,15 @@ def run # rubocop:disable Metrics/AbcSize
...
-      unless @preserve
+      if %i[never on-failure].include? @preserve
         @engine.teardown
         cleanup_workdir
       end
     rescue => e
+      if [:never].include? @preserve
+        @engine.teardown
+        cleanup_workdir
+      end
```
mwaggett: I don't understand why this is in the rescue block.

lib/vanagon/optparse.rb
```diff
@@ -2,40 +2,86 @@
...
+          @options[:workdir] = workdir
+        end
+
+        opts.on("-r DIRECTORY",
+                "--remote-workdir DIRECTORY",
+                "Working directory on the remote host (defaults to calling mktemp)") do |remote|
+          @options[:"remote-workdir"] = remote
+        end
+
+        opts.on("-c",
```
mwaggett: should be `"-c DIRECTORY",` ya?

e-gris:
@mwaggett Good catch.

lib/vanagon/driver.rb
mwaggett: jk, figured it out

lib/vanagon/optparse.rb
```diff
@@ -2,40 +2,86 @@
...
+        end
+
+        opts.on("-p [RULE]",
+                "--preserve [RULE]",
+                ["never", "on-failure", "always"],
+                "Rule for VM preservation. [never, on-failure, always]") do |rule|
+          if rule.nil?
+            @options[:preserve] = :always
+          else
+            @options[:preserve] = rule
```
melissa: So, question. Above, we're checking the contents of this option against symbols, but there's nothing here that enforces that this is a symbol. Is this me being overly concerned? Or is there something here about OptionParser that I'm missing that handles that?

Also, on line 9 you have `@options[:preserve] = :never`. It seems like we have conflicting defaults here.

e-gris: First point: agreed, this could be cleaner. It functions as-is but it's playing it too loose with Ruby conveniences.

Second point: The defaulting behavior is as we agreed:

- With no --preserve argument: preserve is set to 'never'
- With --preserve but no argument: preserve is set to 'always'
- With --preserve=\<argument\>: preserve is set to \<argument\>

melissa: Oh, derp, that makes perfect sense. Thank you!

And to the symbol check, there are a lot of places in our code where we play it fast and loose with Ruby conveniences. If this is working currently, then I suppose there's no reason to be too concerned until this starts to fail.

melissa:
I know you're not developing on a fork, but how would you feel about squashing some of these commits?

underscorgan:
there were lots of thumbs before the squash, so I just merged it.

PR #501, opened 2017-10-09
(VANAGON-96) (VANAGON-97) Generate compiled archives

underscorgan:


lib/vanagon/platform.rb
```diff
@@ -435,6 +435,25 @@ def add_build_repository(*args)
       end
     end
 
+    def generate_compiled_archive(project)
+      metadata = project.build_manifest_json(true)
+      metadata.gsub!(/\n/, '\n')
+      target_dir = project.repo ? output_dir(project.repo) : output_dir
+      name_and_version = "#{project.name}-#{project.version}"
+      archive_dir = "#{project.name}-archive"
```
e-gris: Minor (feel free to ignore): 

* I like spelling out names, archive_directory. I think it makes code easier to read. 
* Unless there are portability concerns, using the --long-options to commands makes them clearer

underscorgan: Unfortunately for this there are portability concerns (as this will run on all the platforms we support) so I've got to go with the short options.

I can fix at least some of the `dir` -> `directory`, but I think that shortening is pretty pervasive throughout the codebase so some of that might not make sense to do here.

e-gris: I'd be happy with consistency inside of new methods.

lib/vanagon/platform.rb
```diff
@@ -435,6 +435,23 @@ def add_build_repository(*args)
       end
     end
 
+    def generate_compiled_archive(project)
```
mwaggett: probably want yardoc for this method, since we have 'em for the rest.

lib/vanagon/project/dsl.rb
```diff
@@ -157,6 +157,20 @@ def generate_source_artifacts(source_artifacts)
...
+      def generate_packages(pkg)
+        @project.generate_packages = pkg
+      end
+
+      # Output os-specific archives containing the binary output
+      #
+      # @param archive [Boolean] whether or not to output archives
+      def generate_archives(archive)
+        @project.compiled_archive = archive
+      end
```
mwaggett: Are these two used anywhere?

underscorgan: These are needed to be able to set those in vanagon projects.

lib/vanagon/platform.rb
underscorgan: ok, will add

lib/vanagon/project/dsl.rb
mwaggett: ahh okay. 👍

lib/vanagon/platform/solaris_10.rb
```diff
@@ -48,7 +48,6 @@ def generate_package(project) # rubocop:disable Metrics/AbcSize
 
           bom_install,
 
-          "rm #{name_and_version}.tar.gz",
```
haus: You may run into issues here. IIRC when building some of the bigger/more complex packages (gcc), either the pooler or hardware ran out of space. Shouldn't be hard to address on the hardware/vm side, but just a heads up.

underscorgan: ok, cool, thanks @haus

underscorgan: I tested a gcc build on sol10 which seemed fine (failed with perm issues shipping the artifacts, but no failures before that). Do you know if there was anything else that might've triggered this?

haus: might also depend on the tmpdir used (which i believe has changed over time). might not be worth worrying over yet

lib/vanagon/platform.rb
```diff
@@ -435,6 +435,29 @@ def add_build_repository(*args)
       end
     end
 
+    # Save the generic compiled archive and relevant metadata as packaging
+    # output. This will include a json file will all of the components/versions
```
sarameisburger: I think you meant the second 'will' to be 'with'

underscorgan: woops, yup. fixing

underscorgan:
Woops, found a bit of a bad rebase in some more manual testing. Fixing up now.

underscorgan:
fix for that issue pushed.

lib/vanagon/platform.rb
```diff
@@ -435,6 +435,31 @@ def add_build_repository(*args)
       end
     end
 
+    # Save the generic compiled archive and relevant metadata as packaging
+    # output. This will include a json file with all of the components/versions
+    # that were built and a bill of materials when relevant. The archive will be
+    # a gzipped tarball.
+    #
+    # @param project The Vanagon::Project to run this on
```
mwaggett: probably want a `@return` saying that this returns an array of commands, rather than executing the commands themselves.

underscorgan: added

PR #502, opened 2017-10-11
(VANAGON-99) Add the ability to consume precompiled tarballs

underscorgan:
This commit adds the `install_only` option for the component, which,
when true will skip makefile rules that are not needed with precompiled
tarballs (unpack, patch, configure, build, check). This is only intended
to work with tarballs generated using
Vanagon::Project.generate_archives.


lib/vanagon/component/rules.rb
```diff
@@ -48,17 +48,26 @@ def initialize(component, project, platform)
...
-          patch_rule,
-          configure_rule,
-          build_rule,
-          check_rule,
-          install_rule,
-          clean_rule,
-          clobber_rule,
-        ]
+        if component.install_only
+          list = [
```
e-gris: Using something less generic than `list` seems appropriate. Maybe something `list_of_rules`?

Minor: I would also change the pattern to initialize & override rather than if/else:

```ruby
list_of_rules = [
  default_rule1,
  default_rule2,
  ...
]

if component.install_only
  list_of_rules  = [
    override_rule1,
    override_rule2,
    ...
  ]
end
```

At this point, it doesn't matter all that much but it could increase maintainability/refactorability in the future.

underscorgan: Good call, thanks @e-gris !

e-gris: Any time I see an 'else', it triggers scrutiny. There are times where they make perfect sense but they are surprisingly few.

lib/vanagon/component.rb
```diff
@@ -115,6 +115,12 @@ class Component
     # usually a String, but not required to be.
     attr_accessor :cleanup_source
 
+    # When dealing with compiled artifacts generated by setting
+    # `project.generate archives true`, you only need to install the component.
```
mwaggett: (nitpick) should there be an underscore in `generate archives`?

PR #506, opened 2017-10-25
(VANAGON-111) Update to standardized archive names

underscorgan:
Updated to the standard documented in VANAGON-98.

### DIFF
--- a/lib/vanagon/platform.rb
+++ b/lib/vanagon/platform.rb
@@ -443,21 +443,21 @@ class Vanagon
     # @param project The Vanagon::Project to run this on
     # @return array of commands to be run
     def generate_compiled_archive(project)
-      target_directory = project.repo ? output_dir(project.repo) : output_dir
       name_and_version = "#{project.name}-#{project.version}"
+      name_and_version_and_platform = "#{name_and_version}.#{name}"
       archive_directory = "#{project.name}-archive"
       metadata = project.build_manifest_json(true)
       metadata.gsub!(/\n/, '\n')
       [
-        "mkdir -p output/#{target_directory}",
+        "mkdir output",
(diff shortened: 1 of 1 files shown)


lib/vanagon/platform.rb
```diff
@@ -443,21 +443,21 @@ def add_build_repository(*args)
     # @param project The Vanagon::Project to run this on
     # @return array of commands to be run
     def generate_compiled_archive(project)
-      target_directory = project.repo ? output_dir(project.repo) : output_dir
       name_and_version = "#{project.name}-#{project.version}"
+      name_and_version_and_platform = "#{name_and_version}.#{name}"
```
mwaggett: what is `name`?

mwaggett: `platform.name`?

sarameisburger: @mwaggett it's "the name we call it" based on the info on line 7

mwaggett: ooh. I don't see a `name` method though? Should it be `@name`?

underscorgan: the attr_accessor let's us use `name` https://github.com/underscorgan/vanagon/blob/acbe95afd215fb7901c90c732c1507c540e2c0cb/lib/vanagon/platform.rb#L8

mwaggett: oookay! That stuff still confuses me, so I'd love to go over it sometime.
"That stuff" being what accessors do, when to use the @ vars, etc.

sarameisburger:
@underscorgan does this need to wait on anything to be merged in?

underscorgan:
@sarameisburger it's good to go!

PR #507, opened 2017-10-25
(VANAGON-111) Generate a sha1 sum at build time

underscorgan:
In order to validate that the archive has the contents we expect it to, we generate the sha1 sum at build time for verification at build time of the parent project.

### DIFF
--- a/lib/vanagon/platform.rb
+++ b/lib/vanagon/platform.rb
@@ -37,6 +37,7 @@ class Vanagon
     attr_accessor :rpmbuild # This is RedHat/EL/Fedora/SLES specific
     attr_accessor :sort
     attr_accessor :tar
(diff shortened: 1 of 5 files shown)


underscorgan:
Will be rebased after #506 is merged

underscorgan:
This is ready now

mwaggett:
What happens for platforms where you didn't set `@shasum`?

underscorgan:
@mwaggett defaults to 'sha1sum' in the platform's PATH https://github.com/puppetlabs/vanagon/pull/507/files#diff-1d87445e7829950b3efbe54d777322a9R219

mwaggett:
ooh, I guess I misunderstood the fact that `sha1sum` is the actual command

PR #508, opened 2017-10-25
(maint) Simplify compiled archive generation

underscorgan:
The way we were populating the compiled archives in the past caused failures on Solaris. The version of `gtar` on Solaris 10 and 11 claims to support the `-C` flag to change directories prior to adding files, however this ends up giving errors about the directories not being found. Since each line of the makefile executes in a separate shell anyways, let's just `cd` prior to generating the tarball to simplify the commands.

### DIFF
--- a/lib/vanagon/platform.rb
+++ b/lib/vanagon/platform.rb
@@ -456,7 +456,7 @@ class Vanagon
         "mkdir #{archive_directory}",
         "gunzip -c #{name_and_version}.tar.gz | '#{tar}' -C #{archive_directory} -xf -",
         "rm #{name_and_version}.tar.gz",
-        "#{tar} cf #{name_and_version_and_platform}.tar -C #{archive_directory}/#{name_and_version} `#{find} #{archive_directory}/#{name_and_version} -maxdepth 1 -mindepth 1 -type d | sed -e 's|#{archive_directory}/#{name_and_version}/||'`",
+        "cd #{archive_directory}/#{name_and_version}; #{tar} cf ../../#{name_and_version_and_platform}.tar `#{find} . -type d`",
         "gzip -9c #{name_and_version_and_platform}.tar > #{name_and_version_and_platform}.tar.gz",
         "echo -e \"#{metadata}\" > output/#{name_and_version_and_platform}.json",
         "cp bill-of-materials output/#{name_and_version_and_platform}-bill-of-materials ||:",
--- a/lib/vanagon/platform.rb
+++ b/lib/vanagon/platform.rb
@@ -456,7 +456,7 @@ class Vanagon
         "mkdir #{archive_directory}",
         "gunzip -c #{name_and_version}.tar.gz | '#{tar}' -C #{archive_directory} -xf -",
         "rm #{name_and_version}.tar.gz",
-        "cd #{archive_directory}/#{name_and_version}; #{tar} cf ../../#{name_and_version_and_platform}.tar *",
+        "cd #{archive_directory}/#{name_and_version}; rm -f bill-of-materials; #{tar} cf ../../#{name_and_version_and_platform}.tar *",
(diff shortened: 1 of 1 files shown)


underscorgan:
Should be rebased on #506 and #507

lib/vanagon/platform.rb
```diff
@@ -456,7 +456,7 @@ def generate_compiled_archive(project)
         "mkdir #{archive_directory}",
         "gunzip -c #{name_and_version}.tar.gz | '#{tar}' -C #{archive_directory} -xf -",
         "rm #{name_and_version}.tar.gz",
-        "#{tar} cf #{name_and_version_and_platform}.tar -C #{archive_directory}/#{name_and_version} `#{find} #{archive_directory}/#{name_and_version} -maxdepth 1 -mindepth 1 -type d | sed -e 's|#{archive_directory}/#{name_and_version}/||'`",
+        "cd #{archive_directory}/#{name_and_version}; #{tar} cf ../../#{name_and_version_and_platform}.tar `#{find} . -type d`",
```
mwaggett: Can you expand the tar flags? I'm guessing `cf` is `--copy --file`? but I don't know for sure. I'm adopting @e-gris's preference for explicitness. 😊

underscorgan: No, we can't, we need to use the short flags for compatibility. `c` is `--create` and `f` is `--file` in gnu tar but there are no long options for bsd tar

mwaggett: oooh

melissa: I have a question of the find command changes. The first you're only grabbing the top level set of directories, and the second, you're grabbing all of the directories. When I run this locally, I get very different results. Is this intentional?

```
[0] melissa@mrt:~$ find ci-job-configs -maxdepth 1 -mindepth 1 -type d
ci-job-configs/.git
ci-job-configs/.github
ci-job-configs/doc
ci-job-configs/jenkii
ci-job-configs/puppet
ci-job-configs/register-repo
ci-job-configs/resources
ci-job-configs/utils
[0] melissa@mrt:~$ cd ci-job-configs ; find . -type d
.
./.git
./.git/refs
./.git/refs/heads
./.git/refs/tags
./.git/refs/remotes
./.git/refs/remotes/origin
./.git/branches
./.git/hooks
./.git/info
./.git/objects
./.git/objects/pack
./.git/objects/info
./.git/logs
./.git/logs/refs
./.git/logs/refs/remotes
./.git/logs/refs/remotes/origin
./.git/logs/refs/heads
./.github
```

melissa: The directory prefix and getting rid of sed I'm totally on board with. I'm worried about the content disparities. Also, the last ci-job-configs example wasn't a full grab, I just pasted the top few lines

e-gris: Reinforcing what @underscorgan said: explicit is nice but portability can put the kibosh on it. If the script is expected to work on non-GNU systems (Mac and BSD, for instance), care must be taken to stick with common forms.

underscorgan: I had portability issues with `mindepth` and `maxdepth`. `tar` will add things recursively, so I thought that should end up with the same output, though with a quick test I ended up with duplicates when I was adding all the directories, might try to '*'

```
morgan@river.corp.puppetlabs.net % tar cf test2.tar foo foo/bar foo/bar/baz
10:29:19: /tmp
morgan@river.corp.puppetlabs.net % tar tf test2.tar
foo/
foo/bar/
foo/bar/baz/
foo/bar/
foo/bar/baz/
foo/bar/baz/
```

underscorgan:
I'm trying to get rid of the find altogether. The directory we're tarring up should only contain other directories, but running a test now to validate.

e-gris:
find is just a fancy glob :)

PR #510, opened 2017-10-26
(maint) Print backtrace if retry-on-fail fails

underscorgan:
Getting an error message is cool, but without the backtrace of what
actually went wrong it's really hard to debug.


lib/vanagon/utilities.rb
```diff
@@ -147,6 +147,7 @@ def retry_with_timeout(tries = 5, timeout = 1, &blk)
 
       message = "Block failed maximum number of #{tries} tries"
       message += "\n with error #{error.message}" unless error.nil?
```
e-gris: Nitpick. Code Repetition. Refactor with

```ruby
unless error.nil?
  message +=  "\n with error #{error.message}" +
                        "\n#{error.backtrace.join("\n")}
end
```

mwaggett: maybe this will make Rubocop happier too 😊

PR #511, opened 2017-10-27
(VANAGON-111) enable consumption of remote checksums

underscorgan:
This adds the ability to set the sum_url for a component instead of
setting the sum directly.


underscorgan:
gonna rework things a bit

PR #512, opened 2017-10-27
(VANAGON-111) enable consumption of remote checksums

underscorgan:
This adds the ability to set the `sum` for a component to the url of a
file containing the checksum rather than the actual checksum.

### DIFF
--- a/lib/vanagon/component/source/http.rb
+++ b/lib/vanagon/component/source/http.rb
@@ -48,7 +48,7 @@ class Vanagon
         # @param workdir [String] working directory to download into
         # @param sum_type [String] type of sum we are verifying
         # @raise [RuntimeError] an exception is raised is sum is nil
-        def initialize(url, sum:, workdir:, sum_type:, **options)
+        def initialize(url, sum:, workdir:, sum_type:, **options) # rubocop:disable Metrics/AbcSize
           unless sum
             fail "sum is required to validate the http source"
           end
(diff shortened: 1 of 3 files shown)


underscorgan:
I initially implemented this with the addition of a new `sum_url` variable, but when rubocop complained about the number of parameters I reconsidered the issue and decided to let `sum` be either the checksum or a url to a file containing a checksum. Initial implementation is available at https://github.com/underscorgan/vanagon/commit/5bd80460d922803d1485fa7600777f86c329ac58, if we decide that's the better option I can always just add an exception for number of parameters

underscorgan:
Doing some more testing

lib/vanagon/component/source/http.rb
```diff
@@ -62,6 +63,16 @@ def initialize(url, sum:, workdir:, sum_type:, **options)
...
           @workdir = workdir
           @sum_type = sum_type
+
+          if Vanagon::Component::Source::Http.valid_url?(@sum)
+            sum_file = download(@sum)
+            File.open(File.join(@workdir, sum_file)) do |file|
+              # the sha1 files generated during archive creation  are formatted
+              # "<sha1sum> <filename>". This will also work for sources that
+              # only contain the checksum.
+              @sum = file.read.split.first
```
haus: What happens if `@sum` is set to nil?

underscorgan: earlier up in the initialize method we fail unless sum, and nil evaluates to falsey, so it won't get to this

haus: if the fetched file is empty, then read will return `""`, split will return `[]` and first will return `nil`.

underscorgan: ah, good call. will rework this bit

shrug:
Artifactory keeps the various checksum formats as metadata that can be queried via the REST API. We'll probably want to switch to being able to use that at some point. Just food for thought.

underscorgan:
@shrug yeah, I definitely think that's something that we'll want to add. I think this will be useful both for archive stuff in the interim and also generally useful for third party things you might be building that publishes their sums online, like openssl.

PR #513, opened 2017-10-31
(RE-9300) Enforce ship to artifactory

melissa:


bin/ship
```diff
@@ -12,4 +12,17 @@ else
   require 'packaging'
   Pkg::Util::RakeUtils.load_packaging_tasks
   Pkg::Util::RakeUtils.invoke_task('pl:jenkins:ship', 'artifacts', 'output')
+  begin
+    Pkg::Util::RakeUtils.invoke_task('pl:jenkins:ship_to_artifactory', 'output')
+  rescue
+    warn <<-DOC
```
melissa: I don't want to fail hard here yet because of the different requirements this update has. Each project that calls this needs to update their individual Gemfiles to include both the newer packaging repo and the artifactory gem. We also need to ensure where ever this ship is happening has the correct authorization.

mwaggett: I agree and this seems like a good (temporary) solution, but how to we plan to make sure teams see this and update? because won't this just end up in the console logs, which people won't look at unless something fails hard?

Gemfile
```diff
@@ -23,7 +23,7 @@ end
 group(:development, :test) do
   gem 'json'
   gem 'lock_manager', *lock_manager_location_for(ENV['LOCK_MANAGER_LOCATION'] || '>= 0')
-  gem 'packaging', github: 'puppetlabs/packaging', branch: 'master'
+  gem 'packaging', github: 'puppetlabs/packaging', branch: '1.0.x'
```
melissa: I don't know what the far-reaching impacts of this change will be....

bin/ship
shrug: As we cut over we'll change this to be a hard fail, but this is better for testing/troubleshooting as we phase it in

melissa: Yup, they won't see it unless they're looking for it. We can always change it to fail hard when we're fairly sure everyone has the setup they need, and then deal with any failures that surface then.

PR #515, opened 2017-11-13
(maint) Do not place source URIs into the mirror list

branan:
When populating the mirror list, vanagon attempts to include rewritten
URIs as a fallback for when the user has not specified any mirrors
explicitly.

Unfortunately, the URI rewriter returns the original URI if there is
no matching rule. This means that the original source URI ends up in
the mirror list, and can potentially be accidentally used even when
the user has specified a dedicated mirror to target.


branan:
@mckern I blame you (but also myself because I think I reviewed your mirror PR)

lib/vanagon/component.rb
```diff
@@ -265,7 +265,8 @@ def fetch_url(options)
     # a rewritten URL if there's any value for #url.
     def deprecated_rewrite_url
       return nil unless url
-      Vanagon::Component::Source::Rewrite.parse_and_rewrite(url)
+      rewritten_url = Vanagon::Component::Source::Rewrite.parse_and_rewrite(url)
+      url == rewritten_url ? nil : rewritten_url
```
mwaggett: Is there a non-deprecated method to fix this in?

branan: This is the method that's called directly when `mirrors` is trying to populate its default value. The other option is to do the check in the `mirrors` accessor if this just returned `url`

lib/vanagon/component.rb
```diff
@@ -265,7 +265,8 @@ def fetch_url(options)
     # a rewritten URL if there's any value for #url.
     def deprecated_rewrite_url
       return nil unless url
-      Vanagon::Component::Source::Rewrite.parse_and_rewrite(url)
+      rewritten_url = Vanagon::Component::Source::Rewrite.parse_and_rewrite(url)
```
shrug: Maybe add a comment here explaining what's up?

mckern:
@branan 

kristen-bell-laughing-to-crying

lib/vanagon/component.rb
branan: I updated both the comment for this method and the docstring for the `mirrors` accessor

lib/vanagon/component.rb
mwaggett: Maybe this isn't the place for this work, but I feel like `mirrors` shouldn't be calling a deprecated method. Like, why not address this issue in `Vanagon::Component::Source::Rewrite.parse_and_rewrite` and have `mirrors` use that?

branan: The entire rewrite system is deprecated. This method could perhaps better be called `call_deprecated_rewrite_for_compatibility`, but that's a very long method name

branan:
Another option would be to just pull out the rewrite support in the next 0.Y release (SemVer says we 0.Ys are essentially Xs). That would also solve this problem, as we'd no longer be trying to compute the fallback mirror list from rewrites.

I'm not sure if we have internal vanagon projects that still rely on rewrites, though

PR #516, opened 2017-11-16
(maint) Add nspooler support in vanagon for checking out from nspooler

McdonaldSeanp:
This commit adds the ability to pull hosts from the nspooler as well as the
vmpooler. The way this works is simple: the pooler engine now checks both the
vmpooler and nspooler for hosts, and if the response is 'ok'='false' vanagon
just moves on to try the second pooler


ScottGarman:
:+1:

PR #517, opened 2017-11-16
(VANAGON-110) Inherit upstream project settings

underscorgan:


underscorgan:
To do:

- [ ] moar docs
- [ ] tests
- [ ] make sure we're being reasonable with the loading/merging order
- [ ] make sure we have nice error handling

PR #522, opened 2017-12-06
(maint) RPMs are not being properly stripped

underscorgan:


resources/rpm/project.spec.erb
```diff
@@ -14,6 +14,16 @@
     /usr/lib64/rpm/brp-strip-static-archive %{__strip} \
     /usr/lib64/rpm/brp-strip-comment-note %{__strip} %{__objdump} \
 %{nil}
+<%- elsif !@platform.is_aix? -%>
```
mwaggett: Do you need to check for cisco-wrlinux here also?

underscorgan: nope, see the start of the if statement https://github.com/puppetlabs/vanagon/pull/522/files#diff-a92f582d860af1e32c4e5b9e8f9bb177R3

mwaggett: ahh 👍

mwaggett:
🙈 looks fine to me, but I don't know enough about this to give an official ✅

sarameisburger:
Is this description accurate as to why we are doing this? https://imvoid.wordpress.com/2013/06/05/stripping-and-stopping-stripping-of-binaries-in-rpm-build/

underscorgan:
@sarameisburger basically, we aren't currently providing debug packages, and we also want to remove debug symbols from the regular packages for size considerations. Right now rpm packages are fairly significantly larger than deb packages, with this change I was able to cut ~5MB off of the puppet-agent package for el7.

sarameisburger:
Ah ok, that makes sense. Thanks for the detailed explanation!

PR #529, opened 2017-12-27
Update Rubocop to 0.52.1

mckern:
I'm sick, so you know what that means: I'm touching your code again. Before I finish the Docker work I started back in, like June, I thought I'd update Rubocop and let it catch any egregious mess-ups I might make. Here's the needful to make that happen.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -23,10 +23,10 @@ end
 group(:development, :test) do
   gem 'json'
   gem 'lock_manager', *lock_manager_location_for(ENV['LOCK_MANAGER_LOCATION'] || '>= 0')
-  gem 'packaging', github: 'puppetlabs/packaging', branch: '1.0.x'
+  gem 'packaging', :git => 'https://github.com/puppetlabs/packaging', branch: '1.0.x'
   gem 'rake', require: false
   gem 'rspec', '~> 3.0', require: false
-  gem 'rubocop', "~> 0.49.1", require: false
+  gem 'rubocop', "~> 0.52.1", require: false
   gem 'simplecov', require: false
   gem 'yard', require: false
 end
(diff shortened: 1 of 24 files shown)


Gemfile
```diff
@@ -23,10 +23,10 @@ end
 group(:development, :test) do
   gem 'json'
   gem 'lock_manager', *lock_manager_location_for(ENV['LOCK_MANAGER_LOCATION'] || '>= 0')
-  gem 'packaging', github: 'puppetlabs/packaging', branch: '1.0.x'
+  gem 'packaging', :git => 'https://github.com/puppetlabs/packaging', branch: '1.0.x'
```
mckern: Oh right, I fixed this while I was in here. I got tired of Bundler shouting at me about the insecure `git://` protocol.

mckern:
Since Rubocop rewrote all uses of `$stderr.puts` to `warn`, I am thinking about aliasing `Kernel#warn` to `Kernel#info`. Does that sound like the cold medicine talking?

shrug:
Just for visual consistency in the logger call vs the stderr output?

mckern:
@shrug yeah

shrug:
I think I'm down with that.

mckern:
Gonna wait and think about whether or not the alias is a good idea. I have *a lot* of thoughts about it. Don't let it hold this up.

underscorgan:
Do `info` calls get printed to stdout? Would that break builds that are dependent on specific output which is why things ended up in stderr in the first place?

mckern:
@underscorgan there is no `info` yet; it'd just be me aliasing `Kernel.warn` (which is all `warn` really is) to `Kernel.info`. But I'm thinking that I'd rather build meaningful output and logging into Vanagon rather than mucking around with aliases in `Kernel`.

PR #531, opened 2018-01-03
(VANAGON-108) Allow for customizable mktemp

underscorgan:
OSX doesn't support the mktemp flags we use on other platforms, so
specify the correct flags for OSX, and allow mktemp to be overwritten in
platform definitions.


underscorgan:
tested on el7/xenial/osx/sol10sparc/aix6.1

PR #532, opened 2018-01-08
(VANAGON-34) Allow passing version operators for `conflicts`, `replaces`, and `provides`

underscorgan:


lib/vanagon/platform.rb
```diff
@@ -486,5 +491,27 @@ def provision_with(command)
       provisioning << command
       provisioning.flatten!
     end
+
+    # version strings for dependencies, conflicts, replaces, etc need some munging
+    # based on platform.
+    #
+    # @param version_string operator(<,>,=,<=,>=) and version to be munged
```
mwaggett: I'm a little confused about what this param should look like. Do you think you could include an example?

underscorgan: @mwaggett done.

PR #534, opened 2018-01-10
(VANAGON-116) Retry source fetches individually

branan:
Previously, Vanagon would retry fetching all sources as a single
block. This could confuse debugging of failed fetches, since it's
possible for a differnt source to fail in the retry than in the
original run. It also (paradoxically) increased the chances for an
overall job failure, since successful downloads were given another
change to break.

This change causes Vanagon to retry fetching each component
individually, giving each an opportunity to succeed, and making
debugging cleaner when fetches fail.

Since on most platforms build dependencies are install in a single
package manager transaction, that is still treated monolithically in
the driver.

### DIFF
--- a/lib/vanagon/driver.rb
+++ b/lib/vanagon/driver.rb
@@ -11,7 +11,14 @@ class Vanagon
   class Driver
     include Vanagon::Utilities
     attr_accessor :platform, :project, :target, :workdir, :remote_workdir, :verbose, :preserve
-    attr_accessor :timeout, :retry_count
+
+    def timeout
+      @timeout ||= @project.timeout || ENV["VANAGON_TIMEOUT"] || 7200
+    end
+
+    def retry_count
+      @retry_count ||= @project.retry_count || ENV["VANAGON_RETRY_COUNT"] || 1
+    end
 
     def initialize(platform, project, options = { workdir: nil, configdir: nil, target: nil, engine: nil, components: nil, skipcheck: false, verbose: false, preserve: false, only_build: nil, remote_workdir: nil }) # rubocop:disable Metrics/AbcSize
       @verbose = options[:verbose]
@@ -118,8 +125,10 @@ class Vanagon
       @engine.startup(workdir)
 
       warn "Target is #{@engine.target}"
-      retry_task { install_build_dependencies }
-      retry_task { @project.fetch_sources(workdir) }
+      Vanagon::Utilities.retry_with_timeout(retry_count, timeout) do
+        install_build_dependencies
+      end
+      @project.fetch_sources(workdir, retry_count, timeout)
 
       @project.make_makefile(workdir)
       @project.make_bill_of_materials(workdir)
@@ -154,21 +163,11 @@ class Vanagon
       end
 
       warn "rendering Makefile"
-      retry_task { @project.fetch_sources(workdir) }
+      @project.fetch_sources(workdir, retry_count, timeout)
       @project.make_bill_of_materials(workdir)
       @project.generate_packaging_artifacts(workdir)
       @project.make_makefile(workdir)
     end
-
-    # Retry the provided block, use the retry count and timeout
-    # values from the project, if available, otherwise use some
-    # sane defaults.
-    def retry_task(&block)
-      @timeout = @project.timeout || ENV["VANAGON_TIMEOUT"] || 7200
-      @retry_count = @project.retry_count || ENV["VANAGON_RETRY_COUNT"] || 1
-      Vanagon::Utilities.retry_with_timeout(@retry_count, @timeout) { yield }
-    end
-    private :retry_task
 
     # Initialize the logging instance
     def loginit(logfile)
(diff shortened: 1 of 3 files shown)


branan:
`Surrounding space missing in default value assignment.` -_-

Is this a default cop, or did we make a conscious decision to encourage unnecessary whitespace in our code style? *sigh* I guess I'll fix it

mwaggett:
@branan @mckern bumped rubocop recently

underscorgan:
I don't think that's from the new change, we've always had rubocop yell if you didn't have spaces around things.

lib/vanagon/driver.rb
```diff
@@ -32,6 +32,8 @@ def initialize(platform, project, options = { workdir: nil, configdir: nil, targ
       loginit('vanagon_hosts.log')
 
       @remote_workdir = options[:"remote-workdir"]
+      @timeout = @project.timeout || ENV["VANAGON_TIMEOUT"] || 7200
+      @retry_count = @project.retry_count || ENV["VANAGON_RETRY_COUNT"] || 1
```
mwaggett: Is 1 the ideal retry default?

branan: That's what it was before (shrug)

underscorgan: I think so

ScottGarman:
Nit: seems like a couple of grammar/typos in the commit message, "change/(chance)" and "install(ed)"

lib/vanagon/driver.rb
mwaggett: oh, I see, shoulda kept scrolling..

branan:
OK, I'm not gonna refactor initialize to try to make that cop happy in this PR. Any suggestions here? I can add the comment to disable that check on that method

lib/vanagon/project.rb
```diff
@@ -162,12 +162,20 @@ def merged_environment
...
       @components.each do |component|
-        component.get_source(workdir)
+        Vanagon::Utilities.retry_with_timeout(retry_count, timeout) do
+          component.get_source(workdir)
+        end
         # Fetch secondary sources
-        component.get_sources(workdir)
-        component.get_patches(workdir)
+        Vanagon::Utilities.retry_with_timeout(retry_count, timeout) do
+          component.get_sources(workdir)
```
mwaggett: was this meant to be `get_patches`? you've got `get_sources` twice.

mwaggett: oh wait, that's below..

branan: Not the best-named methods in the world, alas

mwaggett: `get_source` vs `get_sources` ugh

mwaggett: ignore me

branan:
It looks like rubocop might be treating those `@var = a || b || c` as pretty high complexity (each || counts as 1 point). That seems overly-aggressive to me, since it's a pretty common ruby initialization idiom.

I can move initialization of `@timeout` and `@retry_count` to be lazy in their accessors, to see if that's enough to make rubocop happy. If that doesn't do the trick I'm giving up and adding a "let this method be complicated" flag to the method.

ScottGarman:
:+1:

branan:
Looks like that did it. I still don't like rubocop.

lib/vanagon/project.rb
```diff
@@ -162,12 +162,20 @@ def merged_environment
     # Collects all sources and patches into the provided workdir
     #
     # @param workdir [String] directory to stage sources into
-    def fetch_sources(workdir)
+    # @param retry_count [Number] number of times to retry each fetch
+    # @param timeout [Number] How long to wait for each fetch before aborting
```
mwaggett: maybe worth adding that time is in seconds?

also, does yardoc like [Number] over [Integer]?

branan: Apparently 'Integer'. I honestly just guessed at the base class of number types in ruby (turns out its `Numeric`, which is an adjective not a noun -_- )

mwaggett: of course

lib/vanagon/project.rb
mckern: ignore me -grand galactic inquisitor

For future use

haus:
differnt => different in your commit message.

haus:
git commit | aspell

branan:
@haus I'm not bad at spllenig, YOU'RE bad at spllenig!

lib/vanagon/driver.rb
```diff
@@ -11,7 +11,14 @@ class Vanagon
   class Driver
     include Vanagon::Utilities
     attr_accessor :platform, :project, :target, :workdir, :remote_workdir, :verbose, :preserve
-    attr_accessor :timeout, :retry_count
+
+    def timeout()
+      @timeout ||= @project.timeout || ENV["VANAGON_TIMEOUT"] || 7200
+    end
+
+    def retry_count()
```
underscorgan: [nitpick] fairly consistently throughout the codebase when we have a def that takes no argument we use `def name` instead of `def name()`

branan:
@underscorgan parens removed.

PR #535, opened 2018-01-11
(maint) A few enhancements for Vanagon::Component::Source

underscorgan:


underscorgan:
@MikaelSmith fyi, this branch should work around the issue you were seeing. Still need to get tests/cleanup before this is mergeable

lib/vanagon/component/source.rb
```diff
@@ -16,16 +16,21 @@ class << self
         # @param workdir [String] working directory to fetch the source into
         # @return [Vanagon::Component::Source] the correct subtype for the given source
         def source(uri, **options)
+          source_type = determine_source_type(uri)
```
haus: Seems like you probably shouldn't call `determine_source_type` if `source_type` was specified?

lib/vanagon/component/source.rb
```diff
@@ -15,34 +15,66 @@ class << self
         # @param options [Hash] hash of the options needed for the subtype
         # @param workdir [String] working directory to fetch the source into
         # @return [Vanagon::Component::Source] the correct subtype for the given source
-        def source(uri, **options)
-          # First we try git
-          if Vanagon::Component::Source::Git.valid_remote?(uri)
+        def source(uri, **options) # rubocop:disable Metrics/AbcSize
+          if uri =~ /^git/
```
e-gris: Would favor

```ruby
if uri.start_with?("git")
```
just to keep unneeded regexs out of the picture.

underscorgan:
@demophoon @shrug @e-gris I guess one point for this PR is do we prefer encoding `git` in the uri (917f2f1aeee8abd67d5fa51f7cc04da7a49fff2a), or adding a `source_type` parameter to the component (480b0a518c8f7582e1399758cae3585c06c073ad) ? I could go either way I think, but I think we should pick one and be consistent about it.

demophoon:
We have been using the automagic way without the `source_type` parameter so far, I'm okay with out it and pushing that out to some other release of vanagon if that is what we do want to move towards so we are not introducing potentially breaking changes now.

underscorgan:
@e-gris hmm, I feel like I like setting source type with an if/else more than duplicating the return for the Vanagon::Componenet::Source::Git, but :woman_shrugging:

e-gris:
I was looking only at pure mechanics rather than overall design.

Regarding the latter, I'm not clear which is correct.

PR #542, opened 2018-01-30
(maint) Add packaging as a runtime dependency

underscorgan:


underscorgan:
My thinking here is you do in fact need the packaging gem to run `ship` and `repo` commands. Also, having version bounds in the vanagon gemspec would allow us to just need to update vanagon in each project, rather than updating packaging and vanagon if we need to bump the packaging bounds. This also might be a terrible idea.

PR #543, opened 2018-02-05
(PA-1245) Add functionality to help build cfacter gem

McdonaldSeanp:
This PR adds functionality to allow work for PA-1245 (cfacter gem) to succeed.

it adds functionality for the following:

* performing an erb translation on a source file
* adds an only_fetch function to projects to allow them to #1 fetch specific files from the build and #2 not perform packaging steps (if only_fetch is specified vanagon stops before the packaging steps)
* adds release_from_git to project to calculate release based on number of commits since last tag.


lib/vanagon/driver.rb
```diff
@@ -116,14 +116,21 @@ def install_build_dependencies # rubocop:disable Metrics/AbcSize
...
+    def run # rubocop:disable Metrics/AbcSize, Metrics/PerceivedComplexity
       # Simple sanity check for the project
       if @project.version.nil? or @project.version.empty?
         raise Vanagon::Error, "Project requires a version set, all is lost."
       end
 
-      @engine.startup(workdir)
+      # if only_fetch has been set in the project, don't execute the
+      # whole makefile. Instead just perform the installation.
+      if @project.only_fetch.empty?
```
e-gris: Minor. Would rather see this as a guardian clause:

```ruby
make_target = ''
if ! @project.only_fetch.empty?
  make_target = @project.name + '-project'
end
```

lib/vanagon/project/dsl.rb
```diff
@@ -171,6 +171,21 @@ def generate_archives(archive)
         @project.compiled_archive = archive
       end
 
+      # Sets the release for the project to the number of commits since the
+      # last tag. Requires that a git tag be present
+      # and reachable from the current commit in that repository.
+      #
+      def release_from_git
+        opengit = Git.open(File.expand_path("..", Vanagon::Driver.configdir))
```
e-gris: Minor. Use 'open_git' rather than 'opengit'

McdonaldSeanp:
@e-gris updated with your suggestions

lib/vanagon/component.rb
```diff
@@ -326,6 +327,9 @@ def get_sources(workdir) # rubocop:disable Metrics/AbcSize
         )
         src.fetch
         src.verify
+        if source.erb
```
underscorgan: You're adding a new option that can be passed with `add_source` but you didn't document it

lib/vanagon/project/dsl.rb
```diff
@@ -317,6 +317,10 @@ def package_override(var)
         platform = @project.platform
         platform.package_override(self._project, var)
       end
+
+      def only_fetch(path)
```
underscorgan: Please add heredoc for the new method

underscorgan: I'm also torn here because I find calling this `only_fetch` confusing (mostly, it sounds to me like something that's going to be a binary setting), but it's not clear to me what a better name might be. Might be better to have some boolean parameter to disable the extra make targets, along with more generalized functionality to customize which artifact(s) you want to retrieve?

lib/vanagon/project/dsl.rb
```diff
@@ -171,6 +171,21 @@ def generate_archives(archive)
...
+      # Sets the release for the project to the number of commits since the
+      # last tag. Requires that a git tag be present
+      # and reachable from the current commit in that repository.
+      #
+      def release_from_git
+        open_git = Git.open(File.expand_path("..", Vanagon::Driver.configdir))
+        last_tag = open_git.describe('HEAD', { :abbrev => 0 })
+        # log has to be given an arbitrarily large parameter so that
+        # it will fetch more than 30 commits (which is the default)
+        git_release = open_git.log(10_000).between(last_tag, 'HEAD').count
```
underscorgan: I'm wondering if you need to be counting log entries, or if just parsing the second part of the git describe would be ok? If `git describe` returns `0.15.2-5-g332dbcc`, can you just split on '-' and take the second to last part as the number of commits? (I'm looking at the `version_from_git` method below)

lib/vanagon/project/dsl.rb
mwaggett: I was thinking the same thing re: confusing name.

McdonaldSeanp: @underscorgan I had the exact same thoughts. I also thought about the two parameter approach and was going to wait and see what y'all thought of this approach first.

I'll update the PR with a second bool param

lib/vanagon/engine/base.rb
```diff
@@ -73,9 +73,15 @@ def ship_workdir(workdir)
...
+      def retrieve_built_artifact(only_fetch)
         FileUtils.mkdir_p("output")
-        Vanagon::Utilities.rsync_from("#{@remote_workdir}/output/*", "#{@target_user}@#{@target}", "output/", @platform.ssh_port)
+        if only_fetch.empty?
+          Vanagon::Utilities.rsync_from("#{@remote_workdir}/output/*", "#{@target_user}@#{@target}", "output/", @platform.ssh_port)
+        else
+          only_fetch.each do |path|
+            Vanagon::Utilities.rsync_from(path.to_s, "#{@target_user}@#{@target}", "output/", @platform.ssh_port)
+          end
+        end
```
mwaggett: I'm wondering if there's a way to DRY this up.

mwaggett: also I feel like `output` should be a var, rather than hard-coded a bunch.

lib/vanagon/component.rb
McdonaldSeanp: Does the doc at https://github.com/puppetlabs/vanagon/pull/543/files#diff-b302efb59b1963d379e36b80fb9890a5R371 suffice? or are there other place I should document it?

lib/vanagon/driver.rb
mckern: I know this has already been updated, but consider `unless` instead if `if !` -- super minor stylistic choice but also very in-line with the overall code throughout Vanagon

lib/vanagon/project/dsl.rb
mckern: (Morgan doesn't need my validation, but I was going to leave a similar comment)

mckern: (Yes, I stubbed out every single option to `git rev-list`, and followed the "support dash-case and snake_case" examples of the upstream Git library. For **reasons**.)

McdonaldSeanp:
@underscorgan @mwaggett updated with your suggestions

lib/vanagon/project/dsl.rb
```diff
@@ -171,6 +171,21 @@ def generate_archives(archive)
...
+      # Sets the release for the project to the number of commits since the
+      # last tag. Requires that a git tag be present
+      # and reachable from the current commit in that repository.
+      #
+      def release_from_git
+        git_version = Git.open(File.expand_path("..", Vanagon::Driver.configdir)).describe('HEAD', tags: true)
+        # git describe actually contains the information we need here.
+        # for example: 0.15.2-5-g332dbcc, the -5- indicates there are 5
+        # commits since the last tag, so just split on '-' and take the second
+        # output
```
underscorgan: [nitpick] you're taking the second from the last output, which is probably the correct behavior (in case the tag has a `-` in it), but the comment doesn't match actual behavior.

lib/vanagon/driver.rb
```diff
@@ -132,11 +138,12 @@ def run # rubocop:disable Metrics/AbcSize
 
       @project.make_makefile(workdir)
       @project.make_bill_of_materials(workdir)
-      @project.generate_packaging_artifacts(workdir)
+      # Don't generate packaging artifacts if no_packaging is set
+      @project.generate_packaging_artifacts(workdir) if @project.no_packaging
```
mwaggett: This seems.. wrong.

McdonaldSeanp: Oh, yeah, that's backwards

McdonaldSeanp: Fixed

lib/vanagon/engine/base.rb
```diff
@@ -73,9 +73,15 @@ def ship_workdir(workdir)
...
-        FileUtils.mkdir_p("output")
-        Vanagon::Utilities.rsync_from("#{@remote_workdir}/output/*", "#{@target_user}@#{@target}", "output/", @platform.ssh_port)
+      def retrieve_built_artifact(artifacts_to_fetch, no_packaging)
+        output_path = 'output/'
+        FileUtils.mkdir_p(output_path)
+        unless no_packaging
+          artifacts_to_fetch << "#{@remote_workdir}/output/*"
+        end
+        artifacts_to_fetch.each do |path|
+          Vanagon::Utilities.rsync_from(path, "#{@target_user}@#{@target}", output_path, @platform.ssh_port)
```
underscorgan: Are you expecting the `artifacts_to_fetch` list to include build metadata/etc files or should we automagically be pulling these in?

McdonaldSeanp: I didn't modify what happens by default in this function, so I _think_ the build metadata/etc stuff was just included in the output/ dir on the builder right?

underscorgan: I think those only get fetched due to the `output/*` in L80 though, so I think if `no_packaging` is true you won't get them

McdonaldSeanp: Ah, I see. I'll have this function produce metadata files if no_packaging is true

McdonaldSeanp: @underscorgan so, I took a look at what we do with the metadata file. And it turns out we only fetch a build_metadata file from output/ when we are generating an archive as the output: https://github.com/puppetlabs/vanagon/blob/0.15.1/lib/vanagon/platform.rb#L459L485

And even that is a duplicate. The save_manifest_json function here: https://github.com/puppetlabs/vanagon/blob/0.15.1/lib/vanagon/driver.rb#L136 saves the same information in ext (which it will still do during a no_packaging build)

Since no_packaging should identify that no packaging steps are going to be executed, it wouldn't make sense have generate_compiled_archive and no_packaging set to true.

With that information, should I still figure out making and grabbing the metadata file from output/ ?

lib/vanagon/project/dsl.rb
```diff
@@ -171,6 +171,21 @@ def generate_archives(archive)
...
+      def release_from_git
+        git_version = Git.open(File.expand_path("..", Vanagon::Driver.configdir)).describe('HEAD', tags: true)
+        # git describe actually contains the information we need here.
+        # for example: 0.15.2-5-g332dbcc, the -5- indicates there are 5
+        # commits since the last tag, so just split on '-' and take the
+        # second to last output
+        release(git_version.to_s.split('-')[-2])
+      rescue Git::GitExecuteError
+        warn "Directory '#{dirname}' cannot be versioned by git. Maybe it hasn't been tagged yet?"
+      end
```
mckern: I dumped a longish comment and some suggested example code in https://github.com/puppetlabs/vanagon/pull/543#discussion_r166536715, and totally overlooked that you might not be getting notices or notice something buried in a comment attached to an outdated commit. Take a look at that for an alternate suggestion along the lines that led to the switch to `split`.

McdonaldSeanp: @underscorgan do you have a particular preference on which approach to take?

I wouldn't mind adding the new git rev-list lib from Ryan's example: https://gist.github.com/mckern/89d2da3ad1ef7e1417e88424b3ac933f but I defer to you as to whether you want to add the additional complexity there.

underscorgan: @McdonaldSeanp I like the approach Ryan had, can you give that a shot?

McdonaldSeanp: Sure thing!

mckern: (I know it's a lot more code than doing the `split` operation -- it won't break my heart if y'all opt not to extend the upstream Git library. I understand the tradeoffs on abstraction vs. maintainability vs. solving your specific problem that you actually have right now.)

mckern: (I tossed a quick license on the example that basically gives you permission to use the code however you want to do whatever you want. Just dotting T's and crossing I's.)

crosseyedemoji

McdonaldSeanp:
@underscorgan is there anything else you need on this PR?

underscorgan:
@McdonaldSeanp I'd like to see a test build with this in use. I tried building cfacter off the branch in the PR you listed above but I haven't been able to build that successfully.

McdonaldSeanp:
Okay, there's a passing test build here:

lib/git/rev_list.rb
```diff
@@ -0,0 +1,79 @@
...
+  end
+end
+
+module Git
+  class Base
+    def rev_list(committish = nil, opts = {})
+      lib.rev_list(committish, opts)
+    end
+  end
+end
```
underscorgan: I want to see if I can simplify/future-proof this a bit. I'll give it a shot and get something pushed up for you to take a look at.

underscorgan: I have a simplification up at: https://github.com/McdonaldSeanp/vanagon/pull/1

McdonaldSeanp:
@underscorgan your pr was merged in to mine, thx for the simplification :)

lib/git/rev_list.rb
```diff
@@ -0,0 +1,36 @@
+require 'git'
+
+module LibRevList
+  def rev_list(committish = nil, opts = {})
```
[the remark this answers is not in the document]
McdonaldSeanp: @mwaggett I added tests 👍

lib/git/rev_list.rb
```diff
@@ -0,0 +1,36 @@
...
+    arr_opts = []
+
+    opts.each do |k, v|
+      # allow for passing, say, :max-count or :max_count
+      k = k.to_s
+      k.tr!('_', '-')
+      if v && v.to_s.downcase == 'true' # rubocop:disable Performance/Casecmp
```
McdonaldSeanp: So, here's the deal with `Performance/Casecmp` : We have three options.

1. We do what I did above and disable the cop here
2. We use `casecmp` instead of downcase, _but_ that means the new line would look like:
`if v && v.to_s.casecmp('true') == 0`
which does not look better to me. There is a casecmp? function available that returns a bool and not an integer, but only in ruby > 2.4
3. We disable that cop because it's asking us to do something that makes things look strange.

mwaggett: I'm fine with 3.

lib/vanagon/engine/base.rb
underscorgan: I was worried we wouldn't end up with those files, but since we do, I'm fine with this.

McdonaldSeanp:
@underscorgan @mwaggett kk, updated with the disabled cop, is this ready to merge?

mwaggett:
@McdonaldSeanp There's an empty fake erb file that's getting included in your commit.

McdonaldSeanp:
@mwaggett yep! that's part of the test setup for the new erb source functionality

mwaggett:
ooh didn't realize it was intentional. alrighty.

PR #546, opened 2018-02-28
(VANAGON-120) Run postinstall scripts in the postinstall phase

branan:
Previously, Vanagon ran postinstall scripts for upgrades in the
postuninstall phase. This meant that any postinstall scripts would
always be sourced from the /previous/ version of a package,
which makes it rather difficult to fix bugs in the script or to
change the contract between a preinstall and postinstall phase.

This runs those scripts in the expected phase of the RPM
transaction.

### DIFF
--- a/resources/rpm/project.spec.erb
+++ b/resources/rpm/project.spec.erb
@@ -218,6 +218,11 @@ fi
 if [ "$1" -eq 1 ] ; then
   <%= get_postinstall_actions("install") %>
 fi
+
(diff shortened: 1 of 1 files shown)


branan:
For anyone concerned about horrible fallout of moving things around: AFAICT this is an AIX-specific path through the spec file at this point. It used to also be EL4, but we dropped that a while ago.

haus:
@branan i'm not sure what posinstall is

PR #548, opened 2018-03-14
 (VANAGON-125) Move AIX postinstall scripts to 'triggerpostun'

branan:
In VANAGON-120, we moved postinstall upgrade scripts from the postun
step to the post step, so that they would always come from the new
version of the package, instead of the old. Unfortunately, vanagon
historically has had an unconditional service stop in the preun for
AIX. This meant that when upgrading from older packages, the services
we restart in post would be stopped again in preun.

We need to ensure we can still run the version of the script that
comes with the new package, but do so after the old package's
aggressive preun has run. To do that, the new package registers a
triggerpostun handler for its own name. This handler will be run from
the new package near the end of the transaction - after the old
package's preun and postun scripts.

### DIFF
--- a/resources/rpm/project.spec.erb
+++ b/resources/rpm/project.spec.erb
@@ -194,6 +194,14 @@ if [ "$1" -gt 1 ] ; then
   <%= get_preinstall_actions("upgrade") %>
 fi
 
+<%- if @platform.is_aix? -%>
+    <%- get_services.each do |service| -%>
+        # Stop service on upgrade so that it can be restarted at the end
+        if [ "$1" -gt 1 ] ; then
+            /usr/bin/stopsrc -s <%= service.name -%> > /dev/null 2>&1 || :
+        fi
+    <%- end -%>
+<%- end -%>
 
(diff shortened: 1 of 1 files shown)


branan:
I've confirmed all the scripts are behaving as we want now - postinstall runs after all the pre/post-uninstall steps, but is sourced from the newer version of the package on upgrades.

resources/rpm/project.spec.erb
```diff
@@ -248,6 +251,18 @@ fi
...
+%triggerpostun -- <%= @name %>
+# old versions of vanagon generated an unconditional service stop in the preun. This
+# means that postinstall actions that mess with the service need to be run AFTER that
+# step. To do so, we can run them in the "triggerpostun" when an upgrade is happening.
+
+# Run postinstall scripts on upgrade if defined
+if [ "$2" -gt 0 ] ; then
+  <%= get_postinstall_actions("upgrade") %>
+fi
+<%- end -%>
```
mwaggett: (nitpick) indentation in this block is off

branan: Updated

PR #551, opened 2018-03-21
(VANAGON-75) Fail build explicitly when patch filenames conflict

scotje:
Going to work on adding a couple tests in the morning.

### DIFF
--- a/lib/vanagon/component.rb
+++ b/lib/vanagon/component.rb
@@ -349,7 +349,12 @@ class Vanagon
       unless @patches.empty?
         patchdir = File.join(workdir, "patches")
         FileUtils.mkdir_p(patchdir)
-        FileUtils.cp(@patches.map(&:path), patchdir)
+        @patches.each do |p|
+          target = File.join(patchdir, File.basename(p.path))
+          raise Vanagon::Error, "Duplicate patch files detected, '#{p.path}' would have overwritten '#{target}'! Please ensure patch file names are unique across all components." if File.exist?(target)
+
+          FileUtils.cp(p.path, patchdir)
+        end
       end
     end
 
--- a/spec/lib/vanagon/component_spec.rb
+++ b/spec/lib/vanagon/component_spec.rb
@@ -193,4 +193,35 @@ describe "Vanagon::Component" do
       subject.get_sources(@workdir)
     end
   end
+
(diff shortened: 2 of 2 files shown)


lib/vanagon/component.rb
```diff
@@ -349,7 +349,12 @@ def get_patches(workdir)
       unless @patches.empty?
         patchdir = File.join(workdir, "patches")
         FileUtils.mkdir_p(patchdir)
-        FileUtils.cp(@patches.map(&:path), patchdir)
+        @patches.each do |p|
+          target = File.join(patchdir, File.basename(p.path))
+          raise Vanagon::Error, "Duplicate patch files detected, '#{p.path}' would have overwritten '#{target}'! Please ensure patch file names are unique across all components." if File.exist?(target)
+
+          FileUtils.cp(p.path, patchdir)
+        end
```
mckern: Can the test for patch uniqueness be moved into a method of its own, outside of `get_patches`?

mckern: Oh, alternatively, should patch copying be namespaced so that patch files only need to be unique per component? Just spit-balling here. It's late and I'm awake.

scotje: I could definitely move the check out to a helper method.

I think the long-term fix would be something like you suggest in terms of namespacing patch files or maybe preserving more of the path components from where they are being copied from, but I was just trying to add something quick to make it really obvious when you hit this instead of just potentially getting patch failures (or, more terrifyingly, maybe *not* getting patch failures).

mckern: That's not unreasonable -- when I was a more active maintainer, I'd push for fixing it the first time because there's a slim chance anyone will circle back around once the immediate shortcoming is fixed but it's not my baby anymore. Is there a corresponding JIRA ticket for this problem, where some of the conversation here could be captured?

scotje:
Added some unit tests to cover the current implementation.

lib/vanagon/component.rb
mwaggett: ++ to what mckern said

underscorgan: This is a pretty nasty bug if you hit it. I'm fine with this fix going in as is and ticketing more reworking for patches.

PR #552, opened 2018-03-21
(PA-1909) Add build requirement overrides

McdonaldSeanp:
This commit adds a new feature to the platform object:
platform.override_build_requirement specifies that the requirement given as a
paramter should be overriden and instead use the override when a component
specifies the original requirement.

For example, if a component specifies:

component.build_requires 'gcc'

and the platform specifies

platform.override_build_requirement 'gcc', 'pl-gcc'

the component will prefer the override and specify 'pl-gcc' as the actual
requirement.

This functionality can be ignored in the component by setting
ignore_overrides: true like so:

component.build_requires 'gcc', ignore_overrides: true


ScottGarman:
Having had some discussion context around this already, I'm :+1: on this. The goal here is to enable refactoring agent builds in the future to be able to set whether or not various pl-build-tools packages are used at a high abstraction level, within the platform definition.

stahnma:
I think this is more complicated than using logic in the agent project....

Overrides are more magic than if statements. 

Also, pl-gcc is not the requirement, gcc over a certain version is, which should make pl-gcc the override over system gcc, IMHO. 

What is a world where this method makes sense outside of puppet-agent? It still feels extremely specific to that use case and like we're tucking complexity into vanagon vs in the agent (where we do  have a very tough problem).

McdonaldSeanp:
> Also, pl-gcc is not the requirement, gcc over a certain version is, which should make pl-gcc the override over system gcc, IMHO.

Yeah, what you are describing was the intention. gcc was the requirement and we specified that we were overriding gcc with pl-gcc.

> I think this is more complicated than using logic in the agent project....
> Overrides are more magic than if statements.

It's possible that this idea might make more sense if you use the word provides instead of overrides. In my ideal world what I want is to specify that openssl (and every other component that needs it) needs a gcc to build:

`pkg.build_requires 'gcc'`

and then in the platform configuration for old things I want to specify that we need to use pl-gcc to satisfy any requirements on gcc:

`platform.satisfy_build_requirement_with 'gcc', 'pl-gcc'`

In my opinion that's not really 'magic' it's exactly what we are doing with the 'if' statements, simply moved in to the platform object.

We want to move that information in to the platform objects because 1. it's platform specific information, not component specific and 2. the widespread branching in the components isn't sustainable or scaling well.

McdonaldSeanp:
IMO this information is exactly what the platform object should be for, read as a sentence:

`platform.satisfy_build_requirement_with 'gcc', 'pl-gcc'`

"this platform satisfies gcc requirements with pl-gcc"

stahnma:
Ok, I think it's starting to take shape in my brain.

So, let me ask this -- would we have system tools be the default and special case pl-build-tools? If that's the case, I feel lot better about it. I suppose it would also allow things like

requires 'openssl' 'libressl'
requires 'gcc', 'clang'

That does have some use outside just PA. 👍 if I am understanding and we default to not pl-specifics.

stahnma:
One minor comment, could we do something with keywords vs ordered?

Like

satisfy_requirements "preferred => gcc", allowed => "pl-gcc"

That's not great, but maybe is easier to grok than a comma separated list. I'm not sure. Thoughts?

McdonaldSeanp:
> So, let me ask this -- would we have system tools be the default and special case pl-build-tools? If that's the case, I feel lot better about it. I suppose it would also allow things like

That was exactly my intention here. We default to system tools and then insert pl-build-tools where we need to.

specifically I'm imagining something similar to:
```
if ENV['INTERNAL_BUILD']
  satisfy_requirements preferred => "gcc", allowed => "pl-gcc"
  satisfy_requirements preferred => "boost", allowed => "pl-boost"
  satisfy_requirements preferred => "libyaml", allowed => "pl-libyaml"
...
end
```
in the platform files so we internally could build with build tools using INTERNAL_BUILD=true. This won't end up exactly how we would implement it , but I think the general structure might be the same.

I like the idea of keywords, what about:

`platform.satisfy_requirements_with requirement => "gcc", use => "pl-gcc"`

mckern:
> What is a world where this method makes sense outside of puppet-agent? It still feels extremely specific to that use case and like we're tucking complexity into vanagon vs in the agent (where we do have a very tough problem).

and 

> What is a world where this method makes sense outside of puppet-agent? It still feels extremely specific to that use case and like we're tucking complexity into vanagon vs in the agent (where we do have a very tough problem).

swedish_chef

lib/vanagon/platform/dsl.rb
```diff
@@ -405,6 +405,20 @@ def zypper_repo(definition)
       def add_build_repository(*args)
         @platform.add_build_repository(*args)
       end
+
+      # Adds an override to a build requirement. When component.build_requires is
+      # called, the component will check to see if the requirement matches one of
+      # these overrides, and use the override if there's a match.
+      #
+      # @param build_requirement [String] build requirement to overrides
```
mwaggett: (nitpick) 'overrides' -> 'override'

lib/vanagon/platform/dsl.rb
```diff
@@ -405,6 +405,20 @@ def zypper_repo(definition)
...
+      # @param override [String] value of the override
+      # @raise [Vanagon::Error] an arror is raised if an attempt is made to override a requirement that already has one
```
mwaggett: 'arror' -> 'error'

lib/vanagon/platform/dsl.rb
```diff
@@ -405,6 +405,20 @@ def zypper_repo(definition)
...
+      def override_build_requirement(build_requirement, override)
+        if @platform.build_requirement_overrides.key?(build_requirement)
+          raise Vanagon::Error, "build requirement already overridden by #{@platform.build_requirement_overrides[build_requirement]}"
```
mwaggett: not entirely sure when/where this error would pop up, but may also be useful to include the original build requirement in the message.

spec/lib/vanagon/component/dsl_spec.rb
```diff
@@ -298,13 +298,36 @@
...
+      allow(fake_platform).to receive(:build_requirement_overrides).and_return({ 'library2' => 'override' })
+      comp = Vanagon::Component::DSL.new('buildreq-test', {}, fake_platform)
+      comp.build_requires('library1')
+      comp.build_requires('library2')
+      expect(comp._component.build_requires).to include('library1')
+      expect(comp._component.build_requires).to_not include('library2')
+      expect(comp._component.build_requires).to include('override')
+    end
+
+    it 'Does not the override specified by the platform object when ignore_overrides is true' do
```
mwaggett: not sure what this was meant to say buuuuut..

mwaggett: maybe just 'override' between 'Does' and 'not'?

mwaggett:
This seems reasonable to me. It seems not-that-uncommon to want to satisfy a requirement differently on different platforms, so I'm +1.

mwaggett:
@McdonaldSeanp Did you do any test builds?

McdonaldSeanp:
@mwaggett not yet, I can get you some tomorrow

lib/vanagon/platform/dsl.rb
mckern: (If `Overrides` is being used as a plurality, I can live with that)

shrug:
@McdonaldSeanp Any time for with test builds?

shrug:
ping!

shrug:
Looks like it needs a rebase

mwaggett:
and test builds @McdonaldSeanp 😝

shrug:
Should we close this and reopen if we need to?

PR #553, opened 2018-03-28
(maint) Fixes for custom bill-of-materials paths

underscorgan:
When using a custom project bill-of-materials path for precompiled archives, the bill-of-materials file was not correctly removed from the compiled archive, and thus conflicts with projects using the archive with the same custom bill-of-materials path.

The files added to the tarball should include the path to the bill-of-materials, not just the default bill-of-materials location.

### DIFF
--- a/lib/vanagon/platform.rb
+++ b/lib/vanagon/platform.rb
@@ -469,13 +469,27 @@ class Vanagon
       final_archive = "output/#{name_and_version_and_platform}.tar.gz"
       archive_directory = "#{project.name}-archive"
       metadata = project.build_manifest_json(true)
+
+      # previously, we weren't properly handling the case of custom BOM paths.
+      # If we have a custom BOM path, during Makefile execution, the top-level
+      # BOM is moved to the custom path. So, when cleaning up BOMs for non-custom
+      # paths we just want to remove the BOM at the top level of the tarball.
+      # But, if we have a custom BOM path we want to move it back to where it
+      # was prior to the Makefile execution so we can preserve it as an artifact
+      # but not leave it to conflict if it's installed in the same custom path
+      # as a project using this archive.
+      bill_of_materials_command = 'rm -f bill-of-materials'
+      if project.bill_of_materials
+        bill_of_materials_command = "mv .#{project.bill_of_materials.path}/bill-of-materials ../.."
+      end
+
       metadata.gsub!(/\n/, '\n')
       [
         "mkdir output",
         "mkdir #{archive_directory}",
         "gunzip -c #{name_and_version}.tar.gz | '#{tar}' -C #{archive_directory} -xf -",
         "rm #{name_and_version}.tar.gz",
-        "cd #{archive_directory}/#{name_and_version}; rm -f bill-of-materials; #{tar} cf ../../#{name_and_version_and_platform}.tar *",
+        "cd #{archive_directory}/#{name_and_version}; #{bill_of_materials_command}; #{tar} cf ../../#{name_and_version_and_platform}.tar *",
         "gzip -9c #{name_and_version_and_platform}.tar > #{name_and_version_and_platform}.tar.gz",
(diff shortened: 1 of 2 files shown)


lib/vanagon/platform.rb
```diff
@@ -469,13 +469,27 @@ def generate_compiled_archive(project)
...
+      # BOM is moved to the custom path. So, when cleaning up BOMs for non-custom
+      # paths we just want to remove the BOM at the top level of the tarball.
+      # But, if we have a custom BOM path we want to move it back to where it
+      # was prior to the Makefile execution so we can preserve it as an artifact
+      # but not leave it to conflict if it's installed in the same custom path
+      # as a project using this archive.
+      bill_of_materials_command = 'rm -f bill-of-materials'
+      if project.bill_of_materials
+        bill_of_materials_command = "mv .#{project.bill_of_materials.path}/bill-of-materials ../.."
+      end
```
mckern: twocents_1914


Consider this as a more idiomatic Ruby variable assignment:

```ruby
bill_of_materials_command = if project.bill_of_materials
                              "mv .#{project.bill_of_materials.path}/bill-of-materials ../.."
                            else
                              'rm -f bill-of-materials'
                            end
```

(though I think Rubocop might want it aligned under `bill_of_materials_command`, not under `if`)

lib/vanagon/project.rb
```diff
@@ -525,7 +525,13 @@ def has_services?
     #
     # @return [Array] all the files and directories that should be included in the tarball
     def get_tarball_files
-      files = ['file-list', 'bill-of-materials']
+      files = ['bill-of-materials']
+
+      if bill_of_materials
+        files = ["#{bill_of_materials.path}/bill-of-materials"]
+      end
```
mckern: Samesies ^^

lib/vanagon/platform.rb
mwaggett: `bill_of_materials_command = project.bill_of_materials ? "mv .#{project.bill_of_materials.path}/bill-of-materials ../.." : 'rm -f bill-of-materials'` seems more idiomatic to me?
but also, I'm totally fine with how it is rn.

mwaggett:
I'm honestly a little confused about what exactly this is fixing, but I trust you so

underscorgan:
@mwaggett there were a few places in the code where we weren't paying attention to the custom directory the bill-of-materials file is stored under when we were operating on the bill-of-materials. This updates those places to use the directory if it's set.

This issue cropped up when both a runtime and parent project have the same custom path. Without these fixes, in that case the runtime bill-of-materials would attempt to overwrite the parent project bill-of-materials and cause build failures. This makes custom bills of materials handled consistently with the standard bill of materials.

underscorgan:
And for the assignment style comments. I can update to use @mckern 's suggestion if there are strong feelings, but I don't have any for that 🤷‍♀️ . I'm generally 👎 on using selectors, they tend to make code more confusing / harder to read so I don't want to use that @mwaggett.

underscorgan:
@mwaggett The method where it was being removed and is now being moved out of the way is a compiled archive (like runtime) specific method. For compiled archives, they are formatted such that they're intended to be unpacked at the root level of the filesystem. For those jobs, the bill of materials is shipped alongside the archive rather than inside of it so it doesn't litter the filesystem.

We're moving it so it ends up in the right place to get shipped alongside the archive. This relates to https://github.com/puppetlabs/vanagon/blob/master/resources/Makefile.erb#L42-L44. With custom bill of materials paths the file is moved from its default location during the makefile execution, which means there is no bill of materials at the top level build directory if there's a custom path.


tl;dr - for compiled archive generation we were incorrectly shipping a bill-of-materials in the archive where we shouldn't have been. Moving it removes it from the archive while maintaining it as an artifact to ship alongside the compiled archive.

mwaggett:
ahhhh gotcha, thank you for explaining!

PR #554, opened 2018-03-28
(PA-1958) Add redhatfips as an RPM platform

ScottGarman:
redhatfips is a platform based on redhat with FIPS features enabled. We
treat it as a separate platform due to its unique build requirements.


ScottGarman:
Thanks @mwaggett, I've updated the PR accordingly.

PR #555, opened 2018-03-30
(VANAGON-117) Remove unused '--target' parameter

underscorgan:


underscorgan:
@e-gris nope. Target is still used for `build` and `render`, it's just passed as a positional parameter rather than a flagged parameter. The driver ends up initialized with
```
Vanagon::Driver.new(platform, project, options.merge({ :target => target }))
```

PR #556, opened 2018-04-06
(VANAGON-123) Add support for multiple service types

underscorgan:
Debian in particular supports running modern systems under either
systemd or sysv. This adds support for passing multiple servicetypes in
your platform definition and also installing multiple services.

### DIFF
--- a/lib/vanagon/component/dsl.rb
+++ b/lib/vanagon/component/dsl.rb
@@ -163,36 +163,40 @@ class Vanagon
       #
       # @param service_file [String] path to the service file relative to the source
       # @param default_file [String] path to the default file relative to the source
-      # @param service_name [String] name of the service
-      # @param service_type [String] type of the service (network, application, system, etc)
-      # @param link_target [String] executable service file should be linked to
-      def install_service(service_file, default_file = nil, service_name = @component.name, service_type: nil, link_target: nil) # rubocop:disable Metrics/AbcSize
-        case @component.platform.servicetype
+      # @param options optional extra parameters
+      #        service_name [String] name of the service
+      #        service_type [String] type of the service (network, application, system, etc)
+      #        link_target [String] executable service file should be linked to
+      def install_service(service_file, default_file = nil, service_name = @component.name, **options) # rubocop:disable Metrics/AbcSize
+        servicetype = options[:init_system] || @component.platform.servicetype
+        servicedir = @component.platform.get_service_dir(servicetype)
+
+        case servicetype
         when "sysv"
-          target_service_file = File.join(@component.platform.servicedir, service_name)
+          target_service_file = File.join(servicedir, service_name)
           target_default_file = File.join(@component.platform.defaultdir, service_name)
           target_mode = '0755'
           default_mode = '0644'
         when "systemd"
-          target_service_file = File.join(@component.platform.servicedir, "#{service_name}.service")
+          target_service_file = File.join(servicedir, "#{service_name}.service")
           target_default_file = File.join(@component.platform.defaultdir, service_name)
           target_mode = '0644'
           default_mode = '0644'
         when "launchd"
-          target_service_file = File.join(@component.platform.servicedir, "#{service_name}.plist")
+          target_service_file = File.join(servicedir, "#{service_name}.plist")
           target_mode = '0644'
           default_mode = '0644'
         when "smf"
-          target_service_file = File.join(@component.platform.servicedir, service_type.to_s, "#{service_name}.xml")
+          target_service_file = File.join(servicedir, options[:service_type].to_s, "#{service_name}.xml")
           target_default_file = File.join(@component.platform.defaultdir, service_name)
           target_mode = '0644'
           default_mode = '0755'
         when "aix"
-          @component.service = OpenStruct.new(:name => service_name, :service_command => File.read(service_file).chomp)
+          @component.service << OpenStruct.new(:name => service_name, :service_command => File.read(service_file).chomp)
           # Return here because there is no file to install, just a string read in
           return
         when "windows"
-          @component.service = OpenStruct.new(\
+          @component.service << OpenStruct.new(\
             :bindir_id => "#{service_name.gsub(/[^A-Za-z0-9]/, '').upcase}BINDIR", \
             :service_file => service_file, \
             :component_group_id => "#{service_name.gsub(/[^A-Za-z0-9]/, '')}Component"\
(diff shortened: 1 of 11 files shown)


underscorgan:
see also: https://github.com/puppetlabs/puppet-agent/pull/1393

underscorgan:
testing at http://jenkins-release.delivery.puppetlabs.net/job/vanagon_generic_job/1477/

lib/vanagon/component/dsl.rb
```diff
@@ -163,36 +163,40 @@ def conflicts(pkgname, version = nil)
...
           target_mode = '0644'
           default_mode = '0644'
         when "launchd"
-          target_service_file = File.join(@component.platform.servicedir, "#{service_name}.plist")
+          target_service_file = File.join(servicedir, "#{service_name}.plist")
           target_mode = '0644'
           default_mode = '0644'
         when "smf"
-          target_service_file = File.join(@component.platform.servicedir, service_type.to_s, "#{service_name}.xml")
+          target_service_file = File.join(servicedir, options[:service_type].to_s, "#{service_name}.xml")
```
mwaggett: Should this be the `servicetype` you defined above?

lib/vanagon/component/dsl.rb
```diff
@@ -217,7 +221,7 @@ def install_service(service_file, default_file = nil, service_name = @component.
         end
 
         # Register the service for use in packaging
-        @component.service = OpenStruct.new(:name => service_name, :service_file => target_service_file, :type => service_type)
+        @component.service << OpenStruct.new(:name => service_name, :service_file => target_service_file, :type => options[:service_type])
```
mwaggett: this too

lib/vanagon/platform.rb
```diff
@@ -239,6 +242,7 @@ def initialize(name) # rubocop:disable Metrics/AbcSize
       # Our first attempt at defining metadata about a platform
       @cross_compiled ||= false
       @valid_operators ||= ['<', '>', '<=', '>=', '=']
+      @servicetypes = []
```
mwaggett: Is there a reason this isn't an `||=`?

lib/vanagon/component/dsl.rb
underscorgan: Nope, that was a pre-existing option used for some packaging types.

lib/vanagon/component/dsl.rb
underscorgan: Nope, that was a pre-existing option used for some packaging types.

lib/vanagon/platform.rb
underscorgan: `@servicetypes` is not something you set directly, it's just an underlying datastructure, so we want to initialize it to an empty array, then other methods will set it.

resources/deb/postinst.erb
```diff
@@ -2,21 +2,21 @@
 <%- get_services.each do |service| -%>
   # switch based on systemd vs systemv
   #
-  <%- if @platform.servicetype == "systemd" -%>
-if [ -z "$2" ]; then
-  systemctl enable <%= service.name %>.service >/dev/null || :
+if [ -d '/run/systemd/system' ] ; then
```
mwaggett: Why not use `@platform.get_service_types.include?("systemd")` like you did for rpm?

lib/vanagon/project.rb
```diff
@@ -289,10 +289,14 @@ def has_conflicts?
     # will return nil
     #
     # @param [string] name of service to grab
-    # @return [@component.service obj] specific service
+    # @return [@component.service obj] specific service, or array of services
+    #         if there's more than one
     def get_service(name)
```
mwaggett: I might be missing something, but I don't see any changes to how we're using `get_service` in this PR, but I would expect there to be some adjusting based on whether there's more than one or not.

underscorgan: get_service isn't used in vanagon anywhere. It has minimal use in projects (for example, in agent it's only used in windows packaging), and will only need change if you're using it and setting multiple services.

resources/deb/postinst.erb
underscorgan: This is adding support for installing both sysv and systemd scripts for debian packages, so we need to switch based on what's actually running on the system

lib/vanagon/platform.rb
```diff
@@ -516,5 +520,33 @@ def version_munger(version_string, default: '=')
...
+      if @servicetypes.empty?
+        return @servicedir
+      end
+      servicedir = @servicetypes.select { |s| s.servicetype.include?(servicetype) }.flat_map(&:servicedir).compact
+
+      if servicedir.size > 1
+        raise Vanagon::Error, "You can only have one service dir for each service type. Found '#{servicedir.join(',')}' for service type #{servicetype}"
+      end
+
+      servicedir.join('')
```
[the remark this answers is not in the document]
underscorgan: Updated.

lib/vanagon/project.rb
mwaggett: ooh okay

lib/vanagon/component/dsl.rb
```diff
@@ -163,36 +163,40 @@ def conflicts(pkgname, version = nil)
...
-      # @param service_type [String] type of the service (network, application, system, etc)
-      # @param link_target [String] executable service file should be linked to
-      def install_service(service_file, default_file = nil, service_name = @component.name, service_type: nil, link_target: nil) # rubocop:disable Metrics/AbcSize
-        case @component.platform.servicetype
+      # @param options optional extra parameters
+      #        service_name [String] name of the service
+      #        service_type [String] type of the service (network, application, system, etc)
+      #        link_target [String] executable service file should be linked to
+      def install_service(service_file, default_file = nil, service_name = @component.name, **options) # rubocop:disable Metrics/AbcSize
+        servicetype = options[:init_system] || @component.platform.servicetype
```
mwaggett: I find it a bit confusing to have this `servicetype` and the `service_type` option. Could one of them be called something else?

underscorgan: The local variable could be renamed, but `servicetype` and `service_type` are already distinct fields in platform and component DSLs / definitely in use so those can't be renamed.

underscorgan: updated.

lib/vanagon/component/dsl.rb
```diff
@@ -163,50 +163,54 @@ def conflicts(pkgname, version = nil)
...
+          @component.service << OpenStruct.new(\
             :bindir_id => "#{service_name.gsub(/[^A-Za-z0-9]/, '').upcase}BINDIR", \
             :service_file => service_file, \
             :component_group_id => "#{service_name.gsub(/[^A-Za-z0-9]/, '')}Component"\
           )
           # return here as we are just collecting the name of the service file to put into the harvest filter list.
           return
         else
-          fail "Don't know how to install the #{@component.platform.servicetype}. Please teach #install_service how to do this."
+          fail "Don't know how to install the #{@component.platform.init_system}. Please teach #install_service how to do this."
```
mwaggett: This should be just `init_system` right?

underscorgan: updated.

PR #563, opened 2018-04-25
(PA-1991) Allow inheriting settings from a local project directory

caseywilliams:
Adds an `inherit_local_settings` function to the DSL that allows vanagon
projects to inherit settings from a vanagon project in a separate local
directory. Accepts a vanagon project name and a path to the project
directory. Does not offer the ability to specify a git revision - the files
are used as they are found.

This is intended to support https://github.com/puppetlabs/puppet-agent/pull/1414.

### DIFF
--- a/lib/vanagon/project/dsl.rb
+++ b/lib/vanagon/project/dsl.rb
@@ -322,6 +322,14 @@ class Vanagon
         @project.load_upstream_settings(upstream_project_name, upstream_git_url, upstream_git_branch)
       end
 
+      # Inherit the settings hash from a locally cloned git repository
+      #
+      # @param local_project_name [String] The vanagon project to load settings from
+      # @param local_project_path [String] The path to the local project directory to load settings from
+      def inherit_local_settings(local_project_name, local_project_path)
+        @project.load_local_settings(local_project_name, local_project_path)
+      end
+
       # Set a package override. Will call the platform-specific implementation
       # This will get set in the spec file, deb rules, etc.
       #
--- a/lib/vanagon/project.rb
+++ b/lib/vanagon/project.rb
@@ -688,5 +688,19 @@ class Vanagon
         upstream_project.cleanup
       end
     end
+
+    # Load the settings hash from a local vanagon project.
+    # This will load the specified vanagon project (with no components) and
+    # merge its settings hash with local settings, overriding any duplicates at
+    # the time of calling. No attempt will be made to check out a specific
+    # version of the local repo; Settings will be used as they are.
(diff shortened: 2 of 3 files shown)


lib/vanagon/project/dsl.rb
```diff
@@ -322,6 +322,14 @@ def inherit_settings(upstream_project_name, upstream_git_url, upstream_git_branc
         @project.load_upstream_settings(upstream_project_name, upstream_git_url, upstream_git_branch)
       end
 
+      # Inherit the settings hash from a locally cloned git repository
+      #
+      # @param local_project_name [String] The vanagon project to load settings from
+      # @param local_project_path [String] The path to the local project directory to load settings from
+      def inherit_local_settings(local_project_name, local_project_path)
+        @project.load_local_settings(local_project_name, local_project_path)
+      end
```
underscorgan: This seems potentially useful from a testing standpoint, but I'm having a hard time with the build reproducibility concerns. More detail in the associated puppet-agent PR, but largely I'm not sure how to actually use this in a project without breaking build reproducibility.

caseywilliams: That makes sense - What if I added a third parameter specifying a component, and used that component's json file to enforce a certain git revision? Or a third parameter just specifying the git revision (which we could read from a json file in the project definition separately)?

underscorgan: I would prefer a parameter specifying the git revision. One thing to consider here is that storing that information in the json file is optional, those parameters could also be set in the component's .rb file, so I'd prefer to load that information either from the json or the ref specified in the component.rb file.

But, if we're loading it locally but requiring it be at a specific SHA, what benefit are we getting from loading it locally?

underscorgan: ah, ok. I hadn't considered that for testing the SHA would be updated in the component config without it being pushed, that seems reasonable.

caseywilliams: @underscorgan @McdonaldSeanp brought up an alternate way we could make this process easier - What do you think about publishing settings in json format while building (alongside other artifacts in the output directory), and using those as the source of settings to inherit in downstream projects? (we could still require that the git revisions matched in the runtime and the downstream component) Does that sound any more palatable than full local inheritance?

underscorgan: I think that would be fine? I don't have a strong preference either way.

caseywilliams:
We're going to rework this approach around loading settings from json - tickets and new PRs to come

PR #565, opened 2018-05-08
(maint) Prepare for 0.15.9 release

demophoon:


CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ This changelog adheres to [Keep a CHANGELOG](http://keepachangelog.com/).
 
 ## [Unreleased]
 
+## [0.15.9] - released on 2018-05-08
+### Added
+- Added support for multiple service types
```
underscorgan: It would probably be good to have a little more context here. I like to link these to ticket numbers also when available.

So, maybe something like:

- (VANAGON-123) Add support for multiple service types for a single platform. This is intended primarily for Debian packages supporting both systemd and sysv init systems.

demophoon:
This should be ready to go now

PR #566, opened 2018-05-17
(VANAGON-130) (VANAGON-131) Publish or inherit settings in YAML

caseywilliams:
I still need to write some tests, but I'm hoping to get some feedback on whether this looks acceptable generally. The project DSL has two new methods here:

- `publish_yaml_settings`, which will write a yaml file containing the project's settings for the current platform, plus a sha1sum for this file, to the output directory alongside anything else that was built. 
- `inherit_yaml_settings`, which accepts URIs to a yaml settings file and its sha1sum file and applies those settings to the current project.

We intend to use these methods in puppet-runtime and puppet-agent, so that we don't need to clone the puppet-runtime git repository to inherit its settings during puppet-agent builds. [VANAGON-130](https://tickets.puppetlabs.com/browse/VANAGON-130) and [VANAGON-131](https://tickets.puppetlabs.com/browse/VANAGON-131) have more details, and [PA-1995](https://tickets.puppetlabs.com/browse/PA-1995) describes how we'd like to use this in puppet-agent.

I've tested these changes locally (using file:// URLs) an in an experimental Jenkins pipeline (using http:// URLs) -- all seems fine, but I'll run through it again once I've opened corresponding PRs in other repos.


lib/vanagon/project.rb
```diff
@@ -665,6 +669,27 @@ def save_manifest_json
...
     end
 
+    # Writes a yaml file at `output/<name>-<version>.<platform>.settings.yaml`
+    # containing settings used to build the current project on the platform
+    # provided (and a corresponding sha1sum file) if `yaml_settings` has been
+    # set in the projcet definition.
+    #
+    # @param [Vanagon::Platform] the platform to publish settings for
+    def publish_yaml_settings(platform)
+      return unless @yaml_settings
```
underscorgan: In (most) new code in vanagon we use the attr_accessors when available, so this should be `return unless yaml_settings`

lib/vanagon/project.rb
```diff
@@ -665,6 +669,27 @@ def save_manifest_json
...
+
+      filename = "#{@name}-#{@version}.#{platform.name}.settings.yaml"
```
underscorgan: More accessors, `filename = "#{name}-#{version}.#{platform.name}.settings.yaml"`

lib/vanagon/project.rb
```diff
@@ -665,6 +669,27 @@ def save_manifest_json
...
+      filepath = File.join('output', filename)
+
+      File.open(filepath, 'w') do |f|
+        f.write(@settings.to_yaml)
+      end
+
+      File.open("#{filepath}.sha1", 'w') do |f|
+        f.write(%x(sha1sum #{filepath}))
```
underscorgan: you're going to need to use `platform.shasum` here to make sure you're calling the right command

caseywilliams:
Ok, thanks @underscorgan -- I've made some updates!

spec/lib/vanagon/project_spec.rb
```diff
@@ -308,6 +408,31 @@ class Project
...
+      sha1_file = double('sha1_file')
+
+      expect(File).to receive(:open).with("output/test-project-version.#{platform_name}.settings.yaml", "w").and_yield(yaml_file)
+      expect(File).to receive(:open).with("output/test-project-version.#{platform_name}.settings.yaml.sha1", "w").and_yield(sha1_file)
+      expect(yaml_file).to receive(:write).with({key: 'value'}.to_yaml)
+      expect(sha1_file).to receive(:write)
+      project._project.publish_yaml_settings(platform)
+    end
+  end
+
```
speedofdark: Happy paths testing look pretty solid. Consider more unhappy path options, such as an existing YAML file with missing SHA1 (vice-versa is possible, but not really likely). Are more merge scenarios needed?

lib/vanagon/project.rb
```diff
@@ -692,5 +717,32 @@ def load_upstream_settings(upstream_project_name, upstream_git_url, upstream_git
...
+        raise Vanagon::Error, "YAML settings can only be inherited from a file or http source (got #{yaml_settings_uri})"
+      end
+
+      Dir.mktmpdir do |working_directory|
+        source = Vanagon::Component::Source.source(yaml_settings_uri,
+                                                   workdir: working_directory,
+                                                   sum: yaml_settings_sha1_uri,
+                                                   sum_type: 'sha1')
+        source.fetch
+        @settings.merge!(YAML.load(File.read(File.join(working_directory, source.file)))) # rubocop:disable Security/YAMLLoad
```
underscorgan: Is there a reason we need to use `YAML.load` instead of `YAML.safe_load`?

caseywilliams: I had thought that Psych flat-out disallowed symbol keys in when using `safe_load`, but it looks like you can explicitly allow them -- I'll do that!

lib/vanagon/project.rb
```diff
@@ -692,5 +718,36 @@ def load_upstream_settings(upstream_project_name, upstream_git_url, upstream_git
...
     end
+
+    # Load the settings hash for the current project/platform combination from a
+    # yaml file as produced by `publish_yaml_settings`. file:// and http:// URIs
+    # are accepted. If the URI uses http://, a sha1 URI is also required.
+    #
+    # @param settings_uri [String] A URI to a yaml settings file
+    # @param settings_sha1_uri [String] A URI to a sha1sum file for the yaml settings file
+    # @raise [Vanagon::Error] when the settings file can't be found
+    def load_yaml_settings(settings_uri, settings_sha1_uri = nil) # rubocop:disable Metrics/AbcSize
```
caseywilliams: This method was getting pretty long -- I've disabled the `AbcSize` metric here, since nothing jumped out at me as an obvious thing to extract into its own method, but I'm open to suggestions.

spec/lib/vanagon/project_spec.rb
```diff
@@ -126,6 +127,179 @@
     end
   end
 
+  describe "#load_yaml_settings" do
```
ekinanp: Looking at the code for `load_yaml_settings`, it seems like it would be enough to init. a single Project object here and then test each case accordingly. Something like:

```
describe '#load_yaml_settings' do
  subject(:project) { Project.new("project", "mock_platform" }
  describe 'with a git source' do
    it 'fails to load anything' do
      expect(project.load_yaml_settings(git_uri).to raise_error(Vanagon::Error)
    end
  end
  // etc for the other cases, mocking only when you need to mock a specific method
end
```

I don't understand the use of all these blocks here when you're only testing that this method works for a given project instance. The test for whether
```
project 'upstream-test' do |proj|
end
```

returns the right project object after all of that set-up should be handled by the corresponding DSL test.

caseywilliams: Hmm. I was following the existing conventions in this file, but it would be nice to get rid of a lot of this mocking. I'll do some experiments tomorrow morning.

ekinanp: Yeah I figured. It just seems like the existing conventions are a little overkill here.

caseywilliams: Thanks @ekinanp, I've updated this to remove all of the project definition blocks

ekinanp: Thanks @caseywilliams, the tests are a lot clearer now. One question -- is there a reason why mocking https://github.com/caseywilliams/vanagon/blob/VANAGON-130/yaml-settings/lib/vanagon/project.rb#L744-L747 requires so many steps? It seems like it should be enough to expect the Component::Source.source to receive the right params and then return a "mock" source that, when its fetch method is invoked, sets its file field to the YAML file we want (simulating the download). You should be able to mock all of that in a top-level before block.

spec/lib/vanagon/project_spec.rb
```diff
@@ -126,6 +127,143 @@
...
+      end
+    end
+
+
+    describe 'with http sources' do
+      let(:http_yaml_uri) { "http:/#{yaml_path}" }
+      let(:http_sha1_uri) { "http:/#{sha1_path}" }
+      let(:http_source) { class_double(Vanagon::Component::Source::Http) }
+
+      let(:valid_http_component_source) do
```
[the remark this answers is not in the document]
ekinanp: @McdonaldSeanp You can still mock https://github.com/caseywilliams/vanagon/blob/VANAGON-130/yaml-settings/lib/vanagon/project.rb#L730 and https://github.com/caseywilliams/vanagon/blob/VANAGON-130/yaml-settings/lib/vanagon/project.rb#L744-L747 -- the latter can expect the http source URI and return the mocked double where the double would be similar to the one used for the local source.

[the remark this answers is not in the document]
ekinanp: It's also weird that the method's implementation doesn't care about Http source (b/c knowledge of that is encapsulated in the Component::Source.source method) but the unit tests do.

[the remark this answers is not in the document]
McdonaldSeanp: Ah, that last comment is likely the correct way to look at it.

I agree that we don't need to specifically identify that load_yaml_settings works with the underlying structure of sources. It should be enough just to test that it correctly works with the top-level source object.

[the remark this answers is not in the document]
ekinanp: Also in the vein of that comment, I think the tests could be simplified by structuring them as:
```
(1) Fails for a local source if the settings file doesn't exist
(2) Fails if the source type is anything other than HTTP or local (e.g. git)
(3) Fails if the source type is HTTP but the settings_sha1_uri is not passed-in
(4) Merges the settings with what's in the YAML file correctly for a valid source
```

Idea being is we only care about certain sources in the failure cases, but the happy path only cares about a generic source object.

So you'd have (at least) four unit tests for this routine.

lib/vanagon/project.rb
```diff
@@ -692,5 +718,36 @@ def load_upstream_settings(upstream_project_name, upstream_git_url, upstream_git
...
+      source_type = Vanagon::Component::Source.determine_source_type(settings_uri)
+      if settings_uri =~ /^file/ && source_type == :unknown
```
ekinanp: I think it would be better to have this be:
```
if source_type == :unknown
```

instead. You can add an additional note to the exception message if the settings_uri is a local source indicating that the file doesn't exist, so something like:

```
if source_type == :unknown
  message = if settings_ur =~ /^file/ "Tried to load YAML settings from settings uri but file doesn't exist!" else "Unknown source type #{settings_uri}"
  raise Vanagon::Error, message
end
```

spec/lib/vanagon/project_spec.rb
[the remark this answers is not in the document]
caseywilliams: @ekinanp I've added a new commit reorganizing the tests -- is that what you had in mind?

lib/vanagon/project.rb
caseywilliams: @ekinanp I've updated the logic as you suggest

lib/vanagon/project.rb
```diff
@@ -692,5 +718,38 @@ def load_upstream_settings(upstream_project_name, upstream_git_url, upstream_git
...
+    # yaml file as produced by `publish_yaml_settings`. file:// and http:// URIs
+    # are accepted. If the URI uses http://, a sha1 URI is also required.
+    #
+    # @param settings_uri [String] A URI to a yaml settings file
+    # @param settings_sha1_uri [String] A URI to a sha1sum file for the yaml settings file
+    # @raise [Vanagon::Error] when the settings file can't be found
+    def load_yaml_settings(settings_uri, settings_sha1_uri = nil) # rubocop:disable Metrics/AbcSize
+      source_type = Vanagon::Component::Source.determine_source_type(settings_uri)
+
+      if [:unknown, :git].include?(source_type)
```
ekinanp: Oh cool, this is cleaner.

spec/lib/vanagon/project_spec.rb
```diff
@@ -126,6 +127,74 @@
...
+
+    before(:each) do
+      allow(Dir).to receive(:mktmpdir) { |&block| block.yield '' }
+    end
+
+    it "fails for a local source if the settings file doesn't exist" do
+      expect { project.load_yaml_settings(local_yaml_uri) }.to raise_error(Vanagon::Error)
+    end
+
+    it "fails if given a git source" do
```
ekinanp: @caseywilliams Yeah, this is pretty much what I was thinking. My only other suggestion is maybe adding another test to ensure it fails for an `:unknown` source since the routine also includes a check for that.

caseywilliams: @ekinanp updated, and squashed the tests updates into the original commit

lib/vanagon/project.rb
```diff
@@ -665,6 +669,28 @@ def save_manifest_json
       end
     end
 
+    # Writes a yaml file at `output/<name>-<version>.<platform>.settings.yaml`
+    # containing settings used to build the current project on the platform
+    # provided (and a corresponding sha1sum file) if `yaml_settings` has been
+    # set in the projcet definition.
```
sarameisburger: super small nitpick but project is misspelled here

lib/vanagon/project.rb
```diff
@@ -665,6 +669,28 @@ def save_manifest_json
...
+      raise(Vanagon::Error, "You must specify a project version") unless version
+
+      filename = "#{name}-#{version}.#{platform.name}.settings.yaml"
+      filepath = File.join('output', filename)
+
+      File.open(filepath, 'w') do |f|
+        f.write(@settings.to_yaml)
+      end
+
+      File.open("#{filepath}.sha1", 'w') do |f|
```
stahnma: Any particular reason for sha1? I think we should be defaulting to sha256 these days. ( I realize the collision changes here are low, but just for uniformity)

caseywilliams: I was working with the assumption that `platform.shasum` used sha1sum by default -- do you (or does anyone) know if there's a way to confidently find a sha256sum executable for any platform?

lib/vanagon/project.rb
```diff
@@ -692,5 +718,38 @@ def load_upstream_settings(upstream_project_name, upstream_git_url, upstream_git
...
+    # are accepted. If the URI uses http://, a sha1 URI is also required.
+    #
+    # @param settings_uri [String] A URI to a yaml settings file
+    # @param settings_sha1_uri [String] A URI to a sha1sum file for the yaml settings file
+    # @raise [Vanagon::Error] when the settings file can't be found
+    def load_yaml_settings(settings_uri, settings_sha1_uri = nil) # rubocop:disable Metrics/AbcSize
+      source_type = Vanagon::Component::Source.determine_source_type(settings_uri)
+
+      if %i[unknown git].include?(source_type)
+        message = if settings_uri =~ /^file/
```
underscorgan: [nitpick] I find this way of assigning variables hard to read, would prefer either
```
message = "Can't inherit settings from '#{settings_uri}'. Only http and file URIs are valid."
if settings_uri =~ /^file/
  message = "Tried to load YAML settings from '#{settings_uri}', but the file doesn't exist."
end
```

or

```
if settings_uri =~ /^file/
  message = "Tried to load YAML settings from '#{settings_uri}', but the file doesn't exist."
else
  message = "Can't inherit settings from '#{settings_uri}'. Only http and file URIs are valid."
end
```

caseywilliams: :+1: I like those styles better, too -- updated

lib/vanagon/project.rb
underscorgan: I can't recall if there was a reason I went with sha1 instead of sha256 when I was adding this functionality for runtime stuff, but it may have been related to platform compatibility? That seems like something to look into, but probably out of the scope of this work

PR #569, opened 2018-06-08
(VANAGON-136) Publish yaml settings during render

caseywilliams:
Writes a yaml representation of project settings to the workdir during
`render` (in addition to during `build`) when `publish_yaml_settings`
has been called in the project configuration.

We're hoping to use this to easily determine (without building everything completely) which settings have been updated for project/platform combinations in puppet-runtime PRs. This also removes the hard-coded `output` destination, which should have been how this method worked originally.

### DIFF
--- a/lib/vanagon/driver.rb
+++ b/lib/vanagon/driver.rb
@@ -144,7 +144,7 @@ class Vanagon
       @engine.ship_workdir(workdir)
       @engine.dispatch("(cd #{@engine.remote_workdir}; #{@platform.make} #{make_target})")
       @engine.retrieve_built_artifact(@project.artifacts_to_fetch, @project.no_packaging)
-      @project.publish_yaml_settings(@platform)
+      @project.publish_yaml_settings(workdir, @platform)
 
       if %i[never on-failure].include? @preserve
         @engine.teardown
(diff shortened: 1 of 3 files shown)


underscorgan:
Can you split out the `workdir` changes into a separate commit?

lib/vanagon/driver.rb
```diff
@@ -144,7 +144,7 @@ def run # rubocop:disable Metrics/AbcSize, Metrics/PerceivedComplexity
       @engine.ship_workdir(workdir)
       @engine.dispatch("(cd #{@engine.remote_workdir}; #{@platform.make} #{make_target})")
       @engine.retrieve_built_artifact(@project.artifacts_to_fetch, @project.no_packaging)
-      @project.publish_yaml_settings(@platform)
+      @project.publish_yaml_settings(workdir, @platform)
```
underscorgan: During the build is the workdir 'output' or a directory that gets synced back to output?

caseywilliams: Hmm, it looks like workdir is a temporary directory that's synced back to output, but now that I'm looking more at it, it seems like it'd be better to just use the `platform.output_dir` or something -- update coming soon.

shrug:
@caseywilliams is this still on your radar?

caseywilliams:
I doubt I'll have time to revisit this for a while -- I'll close it for now and reopen once I have something working and tested

PR #572, opened 2018-06-19
(VANAGON-138) Verify yaml settings SHA

caseywilliams:
When I wrote the yaml settings functionality, I made two mistakes:
- I didn't realize it was necessary to explicitly call `verify` to make use of an http source's sha1 file, and
- I didn't correctly collect the `system` output for the sha1 file

The net result was that we were getting sha1 files for settings.yaml that were empty, but not catching that. I've built some puppet-runtime projects using this branch [here](http://builds.puppetlabs.lan/puppet-runtime/8f9b7f88d2b6e877a9342c825dc1cdf9a50b1778/artifacts/), and sha1 files look  valid now.

### DIFF
--- a/lib/vanagon/project.rb
+++ b/lib/vanagon/project.rb
@@ -743,6 +743,7 @@ class Vanagon
                                                    sum: settings_sha1_uri,
                                                    sum_type: 'sha1')
         source.fetch
+        source.verify
         yaml_path = source.file
         if source_type == :http
           yaml_path = File.join(working_directory, source.file)
--- a/spec/lib/vanagon/project_spec.rb
+++ b/spec/lib/vanagon/project_spec.rb
@@ -165,6 +165,7 @@ describe 'Vanagon::Project' do
       allow(Vanagon::Component::Source::Http).to receive(:valid_url?).with(http_yaml_uri).and_return(true)
       http_source = instance_double(Vanagon::Component::Source::Http)
       allow(Vanagon::Component::Source).to receive(:source).and_return(http_source)
+      allow(http_source).to receive(:verify).and_return(true)
 
       expect { project.load_yaml_settings(http_yaml_uri) }.to raise_error(Vanagon::Error)
     end
@@ -173,6 +174,7 @@ describe 'Vanagon::Project' do
       before(:each) do
         local_source = instance_double(Vanagon::Component::Source::Local)
         allow(local_source).to receive(:fetch)
+        allow(local_source).to receive(:verify).and_return(true)
         allow(local_source).to receive(:file).and_return(yaml_path)
 
         allow(Vanagon::Component::Source).to receive(:determine_source_type).and_return(:local)


spec/lib/vanagon/project_spec.rb
```diff
@@ -165,6 +165,7 @@
       allow(Vanagon::Component::Source::Http).to receive(:valid_url?).with(http_yaml_uri).and_return(true)
       http_source = instance_double(Vanagon::Component::Source::Http)
       allow(Vanagon::Component::Source).to receive(:source).and_return(http_source)
+      allow(http_source).to receive(:verify)
```
mwaggett: Do these need to explicitly return anything or no?

mwaggett: if so, seems like it would be worth testing both good and bad responses.

caseywilliams: Ah yeah, if the values don't match, [it will just fail](https://github.com/puppetlabs/vanagon/blob/master/lib/vanagon/component/source/http.rb#L100) but it does return true if they do -- I'll update

lib/vanagon/project.rb
```diff
@@ -683,7 +683,7 @@ def publish_yaml_settings(platform)
       end
 
       File.open("#{filepath}.sha1", 'w') do |f|
-        f.write(system("#{platform.shasum} #{filepath}", err: File::NULL))
+        f.write(%x(#{platform.shasum} #{filepath}))
```
caseywilliams: I misunderstood where this was being executed - This happens locally. I don't want to rely on a `shasum` executable being available this way, though - does vanagon have some idiomatic way of determining the right shasum executable for the local host? If not I could try a few possibilities and error if they all fail...

lib/vanagon/project.rb
```diff
@@ -683,7 +683,7 @@ def publish_yaml_settings(platform)
       end
 
       File.open("#{filepath}.sha1", 'w') do |f|
-        f.write(system("#{platform.shasum} #{filepath}", err: File::NULL))
+        f.write(%x(shasum #{filepath}))
```
caseywilliams: Ah, I could just do this with `Digest::SHA1`, since this is local - update coming.

caseywilliams:
Marking this as don't merge until I verify it again with puppet-runtime

caseywilliams:
This seems to work now -- output excerpt from from a puppet-agent build using the sha1 file produced during [these runtime builds](http://builds.puppetlabs.lan/puppet-runtime/51a114e6a4ebc419acc2ccced9df3f14bb0a0005/artifacts/):
```
Downloading file 'agent-runtime-5.5.x-201806130.7.g51a114e.debian-9-amd64.settings.yaml.sha1' from url 'http://builds.delivery.puppetlabs.net/puppet-runtime/51a114e6a4ebc419acc2ccced9df3f14bb0a0005/artifacts/agent-runtime-5.5.x-201806130.7.g51a114e.debian-9-amd64.settings.yaml.sha1'
Downloading file 'agent-runtime-5.5.x-201806130.7.g51a114e.debian-9-amd64.settings.yaml' from url 'http://builds.delivery.puppetlabs.net/puppet-runtime/51a114e6a4ebc419acc2ccced9df3f14bb0a0005/artifacts/agent-runtime-5.5.x-201806130.7.g51a114e.debian-9-amd64.settings.yaml'
Verifying file: agent-runtime-5.5.x-201806130.7.g51a114e.debian-9-amd64.settings.yaml against sum: 'e9dfb29af6283e908ba2f0375a98a8a93ab525e6'
```

PR #577, opened 2018-07-23
(RE-11270) Create `ext` directory before writing build_metadata

mwaggett:
As we move towards using the new metadata service, files that previously lived
in the `ext` directory may be removed, removing the whole directory as well.
This commit ensures that the `ext` directory is created before trying to write
a new file to it.


mwaggett:
@e-gris Can you show me where you're seeing that? In my local experimentation, I didn't encounter any errors when trying to create a directory that already existed.

e-gris:
My error. I wasn't using ```mkdir_p```. ```mkdir``` will throw the error but ```mkdir_p``` does not.

PR #578, opened 2018-07-31
(maint) Add build_requirements command

McdonaldSeanp:
Sometimes it's useful to list the external build requirements (those that
aren't other components) of a project. This commit adds the build_requirements
command to vanagon to list out a sorted array of those requirements

### DIFF
--- a/bin/build_requirements
+++ b/bin/build_requirements
@@ -1,0 +1,35 @@
+#!/usr/bin/env ruby
+require 'json'
+require 'vanagon/extensions/ostruct/json'
+require 'vanagon/extensions/set/json'
+require 'vanagon/extensions/hashable'
+
+load File.expand_path(File.join(File.dirname(__FILE__), "..", "lib", "vanagon.rb"))
+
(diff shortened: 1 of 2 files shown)


bin/build_requirements
```diff
@@ -0,0 +1,35 @@
...
+
+load File.expand_path(File.join(File.dirname(__FILE__), "..", "lib", "vanagon.rb"))
+
+optparse = Vanagon::OptParse.new("#{File.basename(__FILE__)} <project-name> <platform-name> [options]", %i[workdir configdir engine])
+options = optparse.parse! ARGV
+
+project = ARGV[0]
+platform = ARGV[1]
+
+unless project or platform
```
mwaggett: shouldn't this be `unless (project and platform)`?  like, you want this warning to come up if either are unset, yeah?

mwaggett: (parens may not be necessary, but I put 'em in for clarity)

McdonaldSeanp: oh, yeah, I think I copied that from another bin and it was wrong.

PR #583, opened 2018-10-14
(VANAGON-122) Add VANAGON_USE_MIRRORS env variable

Sharpie:
This commit adds a VANAGON_USE_MIRRORS environment variable that can be used
to skip fetches from internal Puppet Inc. mirrors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,6 +176,13 @@ when using the hardware engine. It defaults to *redis*, with no domain.
 ##### `LOCK_MANAGER_PORT`
 Port of the system where redis is running. Defaults to *6379*.
 
+##### `VANAGON_USE_MIRRORS`
+Controls whether component sources are downloaded directly from upstream URLs
+or from configured mirrors. Most Puppet projects using Vanagon default to
+fetching components from internal mirrors. Set this variable to `n` when
+building outside of the Puppet private network to download directly from
+upstream sources.
+
 ##### `VANAGON_RETRY_COUNT`
 Some phases of compilation support retries. The default value is *1* but
 setting to any integer value greater than 1 will causes these components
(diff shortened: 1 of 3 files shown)


lib/vanagon/component.rb
```diff
@@ -284,7 +284,11 @@ def deprecated_rewrite_url
     def get_source(workdir) # rubocop:disable Metrics/AbcSize, Metrics/PerceivedComplexity
       opts = options.merge({ workdir: workdir })
       if url || !mirrors.empty?
-        fetch_mirrors(opts) || fetch_url(opts)
+        if ENV['VANAGON_USE_MIRRORS'] == 'n'
```
stahnma: I'd probably make this n, N, no, No, 0, false, False.

README.md
```diff
@@ -176,6 +176,11 @@ when using the hardware engine. It defaults to *redis*, with no domain.
 ##### `LOCK_MANAGER_PORT`
 Port of the system where redis is running. Defaults to *6379*.
 
+##### `VANAGON_USE_MIRRORS`
+Controls whether component sources are downloaded directly from upstream URLs
+or from configured mirrors. Valid values are `y` or `n`. Set to `n` when
+building outside of the Puppet private network to skip unreachable mirrors.
```
stahnma: What's the default value?

stahnma:
Also, thanks for doing this. That's awesome.

lib/vanagon/component.rb
Sharpie: Doesn't look like Vanagon has any existing logic for doing this sort of conversion from multiple values to a "true/false".

Is it really worth adding the additional complexity and conversion logic vs. clearly documenting "y" and "n" as the effective values? If someone sets it to "True" when the documented value is "n", is there really a reasonable expectation for that working?

README.md
Sharpie: Dropped `y` and clarified that most Puppet Vanagon projects fetch from internal mirrors by default and that setting `VANAGON_USE_MIRRORS` to `n` will disable that behavior.

PR #585, opened 2018-10-26
VANAGON-75 namespace patches

e-gris:
Here's my Friday Rampage attempt at VANAGON-75. Still likely has some roughness.

The strategy was to put the assembly target directory inside the patch class rather than calculating it in a couple of different places and use the component name as the default name space.


e-gris:
Rubocop is complaining the the parameter list on patches is now too long.

Would it be better to make a rubocop exception or to change the interface on the method to ```apply_patch(patch, options = {})```?

Should rubocop continue to be a pass/fail gate? I know we've had some discussions about its merits.

underscorgan:
moving to `apply_patch(patch, options = {})` seems good. While rubocop can be a pain, I think for vanagon we're getting around a lot of the annoyances by having tuned our rubocop.yml, and using a relatively tight binding for our rubocop dependency.

Also, have you built anything with this branch of vanagon? Would be good to build a puppet-agent runtime (e.g. https://github.com/puppetlabs/puppet-runtime/blob/master/configs/projects/agent-runtime-master.rb) for a few platforms to make sure nothing 💥

e-gris:
Thanks.

Actually building something was going to be my next task.

lib/vanagon/patch.rb
```diff
@@ -18,21 +31,28 @@ class Patch
...
 
-    # @!attribute [r] destination
-    #   @return [String] The working directory where this patch will be applied. Only used for post-installation patches.
-    attr_reader :destination
+    def initialize(origin_path, options = {}) # rubocop:disable Metrics/AbcSize
+      valid_keys = %i[name_space destination strip fuzz after]
+      bad_keys = options.each_key.reject { |k| valid_keys.include? k }
+
+      unless bad_keys.empty?
+        raise Vanagon::Error, "Bad options in patch initialization: #{bad_keys.to_s}."
```
mwaggett: I think `to_s` is redundant here.

e-gris:
I've been able to build pe-java, pe-nginx, and pe-postgresql locally. Are  there instructions somewhere on how to verify the builds?

lib/vanagon/component.rb
```diff
@@ -350,17 +350,18 @@ def extract_with # rubocop:disable Lint/DuplicateMethods
...
+    def get_patches(patch_root)
+      return if @patches.empty?
+      @patches.each do |patch|
+        patch_assembly_path = File.join(patch_root, patch.assembly_path)
+        if File.exist?(patch_assembly_path)
+          raise Vanagon::Error,
+                "Duplicate patch files detected, '#{patch.origin_path}' would have overwritten '#{patch.assembly_path}'"
         end
+        patch_target_directory = File.dirname(patch_assembly_path)
+        FileUtils.mkdir_p(patch_target_directory) unless Dir.exist?(patch_target_directory)
```
mwaggett: Don't really need to check if `patch_target_directory` exists, since `mkdir_p` will just no-op if the directory already exists.

lib/vanagon/component.rb
```diff
@@ -350,17 +350,18 @@ def extract_with # rubocop:disable Lint/DuplicateMethods
...
-
-          FileUtils.cp(p.path, patchdir)
+    # @param patch_root [String] working directory to put the patches into
+    def get_patches(patch_root)
+      return if @patches.empty?
+      @patches.each do |patch|
+        patch_assembly_path = File.join(patch_root, patch.assembly_path)
+        if File.exist?(patch_assembly_path)
+          raise Vanagon::Error,
+                "Duplicate patch files detected, '#{patch.origin_path}' would have overwritten '#{patch.assembly_path}'"
```
mwaggett: I like the "Please ensure patch file names are unique across all components." part of the previously-existing error message, since it gives you a push in the right direction for what needs to be changed. Could that be added back here?

lib/vanagon/patch.rb
```diff
@@ -18,21 +31,28 @@ class Patch
...
-      @fuzz = fuzz
-      @after = after
-      @destination = destination
+      @origin_path = origin_path
+      @namespace = options[:namespace]
+      @assembly_path = "patches/#{@namespace}/#{File.basename(@origin_path)}"
+      @strip = options[:strip]
+      @fuzz = options[:fuzz]
+      @after = options[:after]
+      @destination = options[:destination]
```
mwaggett: Would it be better for defaults to get set here vs. in the `apply_patch` method? In my head, I think it makes more sense to set these on initialization, but I'm not sure if making that change would have other consequences.

lib/vanagon/patch.rb
```diff
@@ -18,21 +31,28 @@ class Patch
...
-    def initialize(path, strip, fuzz, after, destination)
-      raise Vanagon::Error, "We can only apply patches after the source is unpacked or after installation" unless ['unpack', 'install'].include?(after)
-      @path = path
-      @strip = strip
-      @fuzz = fuzz
-      @after = after
-      @destination = destination
+      @origin_path = origin_path
+      @namespace = options[:namespace]
+      @assembly_path = "patches/#{@namespace}/#{File.basename(@origin_path)}"
```
mwaggett: What if `@namespace` is nil?

lib/vanagon/component.rb
e-gris: Good idea but I'm struggling with the wording. Maybe:

"Ensure all patch file names within a component are unique"

(Also, I have a bugaboo that computers are not allowed 'please' and 'thank you'. That is reserved strictly for human interaction. However, I can keep it here if that sort of consistency is desirable).

mwaggett: I think that works! Or perhaps "Ensure unique file names for all patches of a given component."?

lib/vanagon/patch.rb
e-gris: From dsl.rb:

```ruby
options[:namespace] = requested_options[:namespace] || @component.name
```

My thinking is that by the time it gets here, options[:namespace] has already been vetted for nil, unless \@component is nil, in which case that should throw an exception.

Do you think further checks would be helpful?

e-gris: However, a spec test that forces nil here will be worthwhile.

mwaggett: I think if you use `File.join` instead of constructing the path yourself, any `nil` pieces will just get skipped, so the path would just turn into `"patches/#{File.basename(@origin_path)}"`.

mwaggett: actually jk that does not work

lib/vanagon/patch.rb
e-gris: Tried it that way, passes rspec. I like it.

mwaggett:
Looks good to me (pending test builds)!

Can you squash this into fewer (logical) commits?

e-gris:
Using puppet-enterprise-vanagon as a guinea-pig, this all worked as expected.

```
$ export VANAGON_LOCATION=file:///home/eric_griswold/git/forks/vanagon
$ bundle install
$ bundle exec build pe-nginx ubuntu-16.04-amd64
$ bundle exec build pe-java sles-12-x86_64
```

PR #588, opened 2018-11-10
(maint) Add the option to build a Vanagon project with Docker

ekinanp:


ekinanp:
I've managed to build a CentOS 7 agent image with the changes here -- see https://github.com/ekinanp/puppet-agent/commit/80d5d93c0c315df3358d9ec2220b95bfa881f273 for the changes that needed to be made in puppet-agent (gist was to convert make-syntax => bash-syntax -- I have a comment about that in the second commit message).

Unfortunately, it looks like we will _not_ be able to get Docker's caching here as things currently stand because the working directory's files, specifically those of the sources, _do_ change each time you fetch them. I'm guessing this might be because e.g. Git has some hidden timestamp files/logs/what-not that trick Docker into thinking stuff's changed. We _could_ add extra files to .dockerignore, but I'd rather not assume anything about the sources' file structure. I think this should be handled in Vanagon instead. Some ideas off the top of my head are:
* Configure the sources to utilize lazy fetching, i.e. if the files are already there, then the sources should not do anything. This means that we still wouldn't get the caching behavior in the case when Vanagon creates a temporary working directory, but we would get the caching behavior if we re-use a previous working directory.
* Extend the sources to return paths to the relevant files. For example, this would mean that the Git source would return a path to the git directory. This way, we can restructure the Dockerfile to lazily copy stuff over, i.e. the top-level COPY instruction would move to the component's Dockerfiles, where we'd have a single COPY instruction for copying over all of the relevant source files to the remote working directory for each component.
* Add an extra option to Vanagon build that, when passed-in, would not fetch any sources. This would be useful if someone has a working directory created from a previous run. This is probably the easiest (and fastest) solution.

In any case, this stuff can likely be done later. For now, this PR does correctly build a Vanagon project from a Docker image, which was its original goal.

ekinanp:
I'm not sure if this stuff will be merged to master, so I haven't added any unit tests. However the code does work as-is, so the PR is ready for review. I think it may be useful to file a ticket here for the commits.

I can add the unit tests on Monday if need be. It might be best to add them once all of the Puppercon stuff's here (including the "generate the package" part).

ekinanp:
Here's the Dockerfile for https://github.com/ekinanp/puppet-agent/commit/puppercon

lib/vanagon/component/dockerfile.rb
```diff
@@ -0,0 +1,188 @@
...
+      def initialize(component, project, platform)
+        @component = component
+        @project = project
+        @platform = platform
+      end
+
+      # This override of andand_multiline is purely for aesthetic purposes, i.e. it makes
+      # the generated Dockerfile more readable.
+      def andand_multiline(*commands)
+        cmdjoin(commands, " && \\\n    ")
```
ekinanp: I have this separator so that something like
```
RUN command1 && \
         command2
```

can be printed, but for some reason the newline is only indented by a single space, so
```
RUN command1 && \
   command2
```

is printed instead. I'll look into this later, but wanted to point that out here in case anyone knew why/how to fix it.

lib/vanagon/component/dockerfile.rb
```diff
@@ -0,0 +1,188 @@
...
+      # @see [Vanagon::Component::Source]
+      step("unpack") do
+        [
+          andand_multiline(component.environment_variables, component.extract_with)
+        ]
+      end
+
+      # Apply any patches for this component.
+      step("patch") do
+        commands = []
```
ekinanp: These are pretty much copy-pasted from their corresponding analogues in Component::Rules.

lib/vanagon/docker.rb
```diff
@@ -0,0 +1,18 @@
+require 'vanagon/utilities/shell_utilities'
```
ekinanp: The packaging work will likely add `exec` and `run` here.

lib/vanagon/optparse.rb
```diff
@@ -71,6 +71,11 @@ def initialize(banner, symbols = []) # rubocop:disable Metrics/AbcSize
                 "Only build listed COMPONENTs") do |components|
           @options[:only_build] = components
         end
+
+        opts.on("--build_with_docker",
```
ekinanp: I wonder if this could be a `--build_with <tool>` option instead? Maybe a future thing.

lib/vanagon/patch.rb
```diff
@@ -32,7 +32,7 @@ def initialize(path, strip, fuzz, after, destination)
     end
 
     def cmd(platform)
-      "#{platform.patch} --strip=#{@strip} --fuzz=#{@fuzz} --ignore-whitespace --no-backup-if-mismatch < $(workdir)/patches/#{File.basename(@path)}"
+      "#{platform.patch} --strip=#{@strip} --fuzz=#{@fuzz} --ignore-whitespace --no-backup-if-mismatch < ${workdir}/patches/#{File.basename(@path)}"
```
ekinanp: I'm going to run an ad-hoc build of the agent with this change to make sure it doesn't break anything. Note that if it does, then dmidecode will fail to build.

lib/vanagon/project.rb
```diff
@@ -576,6 +577,52 @@ def make_makefile(workdir)
...
+
+    # Evaluates the Dockerfile template and writes the contents to the workdir
+    # for use in building the project
+    #
+    # @param workdir [String] full path to the workdir to send the evaluated template
+    # @param remote_workdir [String] full path to the remote working directory in the image
+    # @return [String] full path to the generated Dockerfile
+    def make_dockerfile(workdir, remote_workdir)
+      # This variable will be passed into the Dockerfile template.
+      create_directories_cmd = dirnames.map do |dir|
```
ekinanp: Is there a way to compute this value in the ERB file (without using the `{}` syntax for blocks)? Would prefer that if so

lib/vanagon/patch.rb
ekinanp: It didn't break anything -- https://jenkins-master-prod-1.delivery.puppetlabs.net/view/puppet-agent/view/ad-hoc/job/platform_puppet-agent-extra_puppet-agent-packaging_adhoc-ad_hoc/452/

lib/vanagon/optparse.rb
branan: [nit] Would be nice to switch this to dashes instead of underbars, because I'm lazy

lib/vanagon/component/dockerfile.rb
branan: If we want to go with this long term, it probably makes sense to factor some of the shared docker/make logic into a base class for output generation.

lib/vanagon/patch.rb
branan: Context for those not super familliar with make vs. shell: This is changing the variable from being expanded by make to being expanded by the shell interpreter later.

lib/vanagon/component.rb
```diff
@@ -391,5 +392,13 @@ def environment_variables
     def rules(project, platform)
       Vanagon::Component::Rules.new(self, project, platform)
     end
+
+    # Converts the component to a Dockerfile that contains the
+    # build instructions.
+    #
+    # @return [String] the contents of the component's Dockerfile
+    def to_dockerfile(project, platform)
```
branan: Should this just be `dockerfile`, to match the name above?

mwaggett: +1

lib/vanagon/component/dockerfile.rb
```diff
@@ -0,0 +1,188 @@
...
+      # @param platform [Vanagon::Platform] The platform where this component will be built.
+      def initialize(component, project, platform)
+        @component = component
+        @project = project
+        @platform = platform
+      end
+
+      # This override of andand_multiline is purely for aesthetic purposes, i.e. it makes
+      # the generated Dockerfile more readable.
+      def andand_multiline(*commands)
```
branan: Should this maybe be called something else?

lib/vanagon/component/dockerfile.rb
```diff
@@ -0,0 +1,188 @@
...
+        cmdjoin(commands, " && \\\n    ")
+      end
+
+      # Generate the component's Dockerfile
+      #
+      # @return [String] the contents of the component's Dockerfile.
+      def generate
+        steps = [
+          unpack_step,
```
branan: Can we add parens here so that these are clearly method calls?

lib/vanagon/docker.rb
```diff
@@ -0,0 +1,18 @@
...
+  # idiomatic way.
+  module Docker
+    extend self
+
+    def docker_cmd
+      Vanagon::Utilities.find_program_on_path('docker')
+    end
+
+    def build(workdir)
+      Vanagon::Utilities.local_command("#{docker_cmd} build #{workdir}")
```
branan: We should add an image name that is based on the project name and current git commit (either SHA or describe output)

lib/vanagon/driver.rb
```diff
@@ -116,7 +118,23 @@ def install_build_dependencies # rubocop:disable Metrics/AbcSize
       end
     end
 
-    def run # rubocop:disable Metrics/AbcSize, Metrics/PerceivedComplexity
+    def build_with_docker
+      if @platform.base_docker_image.nil?
+        raise Vanagon::Error, "The platform must specify the base Docker image that the project will be building from"
+      end
```
branan: We should also validated the presence of docker early.

lib/vanagon/driver.rb
```diff
@@ -116,7 +118,23 @@ def install_build_dependencies # rubocop:disable Metrics/AbcSize
...
+
+      # Fetching our sources in <workdir>/sources ensures that Docker will not
+      # restart the build from the `COPY <workdir> <remote_workdir>` instruction
+      # everytime the Dockerfile or the .dockerignore files change.
+      sources_dir = File.join(workdir, "sources")
+      FileUtils.mkdir_p(sources_dir)
+
+      @project.fetch_sources(sources_dir, retry_count, timeout)
+      @project.make_dockerfile(workdir, @remote_workdir || "/root")
```
branan: Maybe "/build"?

lib/vanagon/driver.rb
```diff
@@ -164,6 +182,16 @@ def run # rubocop:disable Metrics/AbcSize, Metrics/PerceivedComplexity
       end
     end
 
+    def run
+      if @build_with_docker
+        warn "Building the project with docker ..."
```
branan: info?

resources/Dockerfile.erb
```diff
@@ -0,0 +1,42 @@
...
+
+<%- unless install_build_dependencies_cmd.empty? -%>
+# Install any external build dependencies
+RUN <%= install_build_dependencies_cmd %>
+<%- end -%>
+
+# Create the remote working directory, then copy
+# over the fetched sources
+RUN mkdir -p ${workdir}
+COPY sources ${workdir}
```
branan: We need to copy sources per-component, otherwise changes to any source will invalidate the cache for all layers below this one.

resources/Dockerfile.erb
```diff
@@ -0,0 +1,42 @@
...
+WORKDIR ${workdir}
+
+# Build the components
+<%- sorted_components.each do |comp| -%>
+
+<%= comp.to_dockerfile(self, @platform) %>
+<%- end -%>
+
+# Mark Bash as the entrypoint
+ENTRYPOINT ["/bin/bash"]
```
branan: Might make sense for this to eventually be the "tar up output and copy to /output" command, but for now being interactive so we can poke the build is good :+1:

lib/vanagon/component/dockerfile.rb
```diff
@@ -0,0 +1,188 @@
...
+      # the generated Dockerfile more readable.
+      def andand_multiline(*commands)
+        cmdjoin(commands, " && \\\n    ")
+      end
+
+      # Generate the component's Dockerfile
+      #
+      # @return [String] the contents of the component's Dockerfile.
+      def generate
+        steps = [
```
branan: This is where we should inject source copying, I think

lib/vanagon/component/dockerfile.rb
ekinanp: I think I'll keep it as-is for now because it is consistent with the [andand_multiline in the shell_utilities file](https://github.com/ekinanp/vanagon/blob/puppercon/lib/vanagon/utilities/shell_utilities.rb#L20)

ekinanp:
Quick reminder: For anyone wanting to use the changes in this PR, please branch it off of the second commit -- https://github.com/puppetlabs/vanagon/pull/588/commits/0367f8cda43e8d388361e72913a90ad8d6d7805c. I will be addressing review comments + adding improvements in separate commits.

lib/vanagon/driver.rb
ekinanp: Unfortunately there is no `info` method defined.

lib/vanagon/component/dockerfile.rb
```diff
@@ -66,21 +66,26 @@ def andand_multiline(*commands)
...
         if component.install_only
           steps = [
-            install_step,
+            install_step(),
           ]
         end
 
-        steps.join("\n")
+        copy_sources = [
+          "COPY #{component.name}_sources ${workdir}"
```
mwaggett: missing separating commas

ekinanp: Oh sorry, this isn't yet ready for review. I just like pushing things up out of habit. I'll comment on here when the new changes are ready for review.

ekinanp:
The third commit isn't yet ready for review -- I just like pushing things up iteratively out of personal preference. I'll comment on here when it is.

mwaggett:
ahh, I just have VANAGON_LOCATION pointing to your branch, so I'm getting all your updates 😄

lib/vanagon/driver.rb
mwaggett: `puts`?

ekinanp:
Ok, added another commit addressing the review comments that should now be ready for review. I'll work on the "build a package with Docker" part tomorrow.

lib/vanagon/driver.rb
```diff
@@ -116,7 +118,60 @@ def install_build_dependencies # rubocop:disable Metrics/AbcSize
...
+      # they don't.
+      #
+      # TODO: We should probably add a has_docker? method to Vanagon::Docker,
+      # but only when we refactor Vanagon::Utilities.find_program_on_path to
+      # _not_ fail (i.e. throw a RuntimeError), but have it raise a Vanagon::Error
+      # instead.
+      Vanagon::Docker.docker("--version")
+
+      remote_workdir = @remote_workdir || "/build"
+      docker_output_dir = "/OUTPUT"
```
mwaggett: lowercase?

lib/vanagon/driver.rb
```diff
@@ -116,7 +118,60 @@ def install_build_dependencies # rubocop:disable Metrics/AbcSize
...
+
+      # TODO: We should add the version_from_git helper to the project class
+      # instead. For now, just create a Project::DSL object and call that.
+      project_version = Vanagon::Project::DSL.new(@project.name, @platform).version_from_git
+      docker_name_prefix = "#{@project.name}-#{project_version}"
+      image_tag = "#{docker_name_prefix}:latest"
+
+      Vanagon::Docker.build(workdir, tag: image_tag)
+
+      output_path = File.absolute_path(File.join(VANAGON_ROOT, "output/"))
```
mwaggett: Should we use `docker_output_dir` here instead of hard-coding `"output/"`?

ekinanp: No -- `docker_output_dir` is the volume mount on the Docker container. This one is the `output/` directory on localhost, which is [hard-coded for the make builds too](https://github.com/puppetlabs/vanagon/blob/master/lib/vanagon/engine/base.rb#L77).

mwaggett: yeah, but is there any reason for them to be different?

ekinanp:
Ok, added two new commits to this PR that should be ready for review. I'll continue iterating on it while fixing the other platform/ files with the Make vs. Docker stuff.

lib/vanagon/driver.rb
mwaggett: fair.

ekinanp:
Updated the PR with a commit to fix Debian packaging to work with Docker + Make.

lib/vanagon/optparse.rb
caseywilliams: I think switching this to `--build-with=docker` or `--build-with=make` (or whatever that ends up being called) sooner rather than later is important, so that command line options change as little as possible from release to release

caseywilliams:
This all seems to work well, but we need documentation

caseywilliams:
After playing with this a bit and talking to other teams (RE, PDK, Bolt, Installer), here are a few more comments and requests for features/changes from everyone:

## Tagging

Docker image tags should include the platform name in addition to the project name and version, since the same project may be built at the same version for many platforms.

## Build artifacts

Build artifacts from the docker build process should be the same as those output during a non-docker build. This means including the following in the output directory for package builds:

- the package itself, in its usual subpath,
- `<version>.build_metadata.json`
- `<version>.yaml`

(There are significant issues with the accuracy of build_metadata that have been ticketed in VANAGON-132 and VANAGON-144, but we should keep things consistent until we can fix those separately)

And for tarball builds, as with puppet-runtime, we should see the following in the output directory:

- `<project>-<version>.<platform>.tar.gz`
- `<project>-<version>.<platform>.tar.gz.sha1`
- `<project>-<version>.<platform>-bill-of-materials`
- `<project>-<version>.<platform>.json`
- `<project>-<version>.<platform>.settings.yaml`
- `<project>-<version>.<platform>.settings.yaml.sha1`

We'd also like to see:

- a copy of the Dockerfile used to build the image added to the output directory. These Dockerfiles should be named to include the project name, platform name, and version.
- `LABEL`s should be added to the Dockerfile to record versions of components included in the image, so that they can be read via `docker inspect`

## Other notes

Below are my own thoughts:

### DSL

- I'd like methods added to the DSL to detect the type of build that's currently happening (that is, docker or make), like `is_docker_build` or and `is_make_build` or something along those lines.
- Would it make sense to add a `default_build_tool` setting to the project DSL, so that projects that are only intended to build with one tool (like a toolchain docker image) would default to that tool on each build?

### Environment variables

Right now, it seems that:

- calling `proj.environment` in a project file generates an `ENV` statement in the Dockerfile, but
- calling `pkg.environment` in a component file exports a bash environment variable, as with Makefiles.

It seems counterintuitive to have `environment` behave differently in components vs. projects -- I think I would prefer to have a different DSL method to generate docker `ENV` statements (since it's possible that we'd want to do this in components, too).

### Generated Dockerfile

The generated cleanup/unpack steps in the Dockerfile (and a few others) can sometimes produce weird lines; For example:

```
# shellpath-unpack
RUN : no source, so nothing to extract && \
 : nothing to extract && \
 : nothing to extract
```

These strings are inconsequential in a Makefile and don't seem to affect the functionality of the Dockerifle, but it'd be ideal just not include a run statement in this case.

### Other Dockerfile features

It seems like dockerfile features like `EXPOSE` could become useful from within vanagon at some point, but it seems preemptive to spend a lot of time on them in this initial work. Would it make sense to have some method in the DSL that allows users to set arbitrary Dockerfile rules in the meantime?

ekinanp:
@caseywilliams Thanks for gathering all of these requirements/requests. So it looks like something like
```
<project>-<version>.<platform>
```

would be a good name for the image? For the other thoughts,

* Maybe a DSL method like `project.build_tool` would work? That would return `:docker` or `:make`.
* Yes, I think `default_build_tool` would make sense
* I think the separation between project + component environment variables is due to the fact that the former's meant to be used everywhere (and is accessible by all the components) while the latter is on a per-component basis -- this is what https://github.com/puppetlabs/vanagon/blob/master/resources/Makefile.erb#L4 is suggesting. Don't know if the environment set-up is worth tinkering with, but it could be something to discuss later.
* Agreed on removing the weird lines, that's one less instruction to worry about.

@branan mentioned that we may need to implement our own topological sort since `Tsort#tsort` isn't stable, but I'm not sure where he found that out.

There's also some discussion on refactoring the Dockerfile + Makefile rule generation into a common class (since the commands are the same), but that's something to worry about later.

caseywilliams:
That all sounds good, except I don't think we should leave a decision about `envrionment` until later, since updating the behavior of that function in some subsequent release would be a breaking change. I'm worried that making the docker tool's behavior match the makefile behavior too closely for the sake of DRY may introduce some non-ideal behaviors. 

For example, it seems like the behavior of `proj.environment` could be particularly difficult because it includes an `ENV` statement before statements for other components. For example, if you wanted to use `ENV` to customize the `PATH` in your finished container, but your components depend on setting some other PATH during the build process, working around the `ENV` path in each one could get repetitive.

mwaggett:
@ekinanp @caseywilliams I think `<project>.<platform>:<version>` (i.e. with version as the tag) might be better, so that we don't create a new repository for each new version of the container.

ekinanp:
On second thought, I think I'll start addressing the PR comments once https://tickets.puppetlabs.com/browse/VANAGON-145 is more fleshed out. Would be better that way.

sarameisburger:
@ekinanp do you want to keep this PR open or close until work is done on VANAGON-145?

ekinanp:
@sarameisburger Nah, I'll close it. Thanks for the heads-up.

PR #594, opened 2018-12-20
(maint) 0.15.19 prep

underscorgan:
@scotje let us know if your build works, and we can get this released!


scotje:
Yes! My brp customizations are much more effective now!

PR #595, opened 2019-01-08
VANAGON-132

e-gris:
Not feeling terribly confident but from the _best I can tell_, this does what the ticket asked for. Not certain if there are any other aspects that need to be modified.

### DIFF
--- a/spec/lib/vanagon/project_spec.rb
+++ b/spec/lib/vanagon/project_spec.rb
@@ -1,6 +1,7 @@
 require 'vanagon/project'
 require 'vanagon/driver'
 require 'vanagon/errors'
+require 'fakefs/spec_helpers'
 
 describe 'Vanagon::Project' do
   let(:component) { double(Vanagon::Component) }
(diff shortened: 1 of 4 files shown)


caseywilliams:
This looks great to me at first glance! I'll try running this through the ad-hoc agent pipeline and checking out the results today.

spec/lib/vanagon/project_spec.rb
```diff
@@ -413,6 +414,50 @@ class Project
...
+        'build_time' => '2018-07-10T13:34:25-07:00',
+      }
+      bad_sample_metadata = {
+        'BAD KEY' => 'BAD VALUE'
+      }
+      comp1 = Vanagon::Component.new('test-component-10', {}, {})
+      comp1.version = '1.2.3'
+      @proj.components << comp1
+      @proj.version = '123abcde'
+      FakeFS do
```
mwaggett: What is `FakeFS` / what's it for?

mwaggett:
yeah, I'd like to see a build of puppet-agent and cfacter from the same sha so that we can verify that both metadata files get generated in the way we expect.

But this looks good to me!

spec/lib/vanagon/project_spec.rb
e-gris: FakeFS is an easy way to keep tests from leaving random files on the filesystem. It creates a temporary filesystem that only exists while it is in context.

https://rubygems.org/gems/fakefs

mwaggett:
also, before this goes in, I'd like to see the commits squashed and some more detail in the commit message (e.g. why this change was necessary, what was happening before this change, etc.).

spec/lib/vanagon/project_spec.rb
mwaggett: ooh handy!

caseywilliams:
Hmm, I'm not seeing multiple build_metadata json files created when I run this -- here's an example with the generic vanagon jenkins pipeline: https://jenkins-master-prod-1.delivery.puppetlabs.net/job/platform_vanagon-generic-builder_vanagon-packaging_generic-builder/245/
The artifacts are here: http://builds.puppetlabs.lan/puppet-agent/b7ae939f50c3e1430fd5cfeed2062aee3904102e/artifacts/ but I don't see the platform or project name in the json filename

mwaggett:
@caseywilliams I'm wondering if that's actually using this branch of vanagon. I think VANAGON_LOCATION is not quite set correctly in the job you linked - should be `git@github.com/e-gris:vanagon#vanagon-132` (`:` after username, not `/`). I'm doing a local build now to verify.

mwaggett:
okay, yeah, my local build produced the expected file, but something weird is happening when I try to run the pipeline: https://jenkins-master-prod-1.delivery.puppetlabs.net/job/platform_vanagon-generic-builder_vanagon-prep-project_generic-builder/256/console
I think I've seen that error before and it's a red herring, but I forget what it actually means. Perhaps tomorrow-me will figure it out..

e-gris:
Hmm, that seems scarily close to the ongoing rubygems issue on Artifactory but if you've seen it before. I would be interested in understanding it better.

mwaggett:
It could be that.. I'm not really sure. But you could do your rubygems workaround and we could try again?

mwaggett:
okay, we confirmed that the issue I was seeing _was_ the recurring Artifactory issue.

but I know why we're not seeing the new files on builds: https://github.com/puppetlabs/packaging/blob/1.0.x/tasks/ship.rake#L574-L580
We'll need to update ^that to ship the newly-named metadata.

PR #596, opened 2019-01-22
(PE-24814) Add handling for ghost files in rpm specs

jpartlow:
Making use of the alternates service in a Redhat platform requires
tracking target files as 'ghost' files in the rpm spec.

Per https://fedoraproject.org/wiki/Packaging:Alternatives#How_to_use_alternatives
we touch the generic links so that rpm will be aware of them when
building the package, but mark them now as %ghost in the %files list
so that the dummy files we touched are not actually copied in when the
package is installed.

This is intended to allow commands like `rpm -qf
/opt/puppetlabs/server/bin/psql` to correctly report that the symlink
(created in the post trigger by update-alternatives) is managed by
the pe-postgresql10 package, for example...

This patch adds some getters and setters to allow adding these ghost
file settings through the Component::DSL and then having the Project add
them to the final rpm spec.


jpartlow:
This is needed for puppetlabs/puppet-enterprise-vanagon#157

lib/vanagon/component.rb
```diff
@@ -213,6 +224,14 @@ def configfiles
       @files.select(&:configfile?)
     end
 
+    # Retrieve all the files intended as %ghost entries for an rpm spec
+    # %files section.
+    #
+    # @return [Array] of all the rpm %ghost files.
+    def rpm_ghost_files
+      @ghost_files.to_a
```
mwaggett: Why return as an Array instead of a Set?

lib/vanagon/component/dsl.rb
```diff
@@ -281,6 +281,16 @@ def link(source, target)
         @component.add_file Vanagon::Common::Pathname.file(target)
       end
 
+      # Add a %ghost entry to the rpm %files section.
+      # This indicates a file that is tracked in the rpm database (the package
+      # manages it), but not installed. Used, for example, when setting up
+      # alternative packages for use with update-alternatives.
+      #
+      # @param file [String] installed path of the file.
```
mwaggett: you've called the param `file` here, but `path` below

jpartlow: Oops, thanks; I'll switch that to path in the docstring.

lib/vanagon/component.rb
jpartlow: Well, this is mostly a tweaked variant of the Component#{config}files() methods, and while they start with an initial Set, the files() and configfiles() methods call reject and select, respectively, which force it to an Array, and the Project.get_files() expects to work with (and returns) an Array. In particular, Array#flatten() will not flatten Sets, so you get lumpy data that breaks in Project#get_rpm_ghost_files() otherwise.

jpartlow:
@mwaggett Yes, I've used this patch to build all the platforms for the new versioned postgres package variants in puppetlabs/puppet-enterprise-vanagon#157.

I haven't found a Debian equivalent, but, as I recall, it was a requirement for making use of alternatives in rpm.

In the postgres packages I only call it when platform.is_el?(). I think the use would be similar to the DSL#add_rpm_install_triggers() or DSL#add_debian_instance_triggers(), so only for specific platform config. Although since only the respective rpm/deb resources attempt to read the contents of that state,  if you happened to declare it generally, it probably wouldn't matter?

lib/vanagon/component/dsl.rb
jpartlow: fixed

PR #600, opened 2019-04-03
(VANAGON-147) Disable automatic shebang munging

mwaggett:
This commit updates the rpm spec file template to disable automatic shebang
munging, which was introduced in Fedora 28 and RHEL 8. Without this, certain
builds fail on these platforms because the python shebang gets munged
ambiguously. For more info, see
https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/message/2PD5RNJRKPN2DVTNGJSBHR5RUSVZSDZI/.

### DIFF
--- a/resources/rpm/project.spec.erb
+++ b/resources/rpm/project.spec.erb
@@ -22,6 +22,13 @@
 # to resolve: "ERROR: No build ID note found"
 %undefine _missing_build_ids_terminate_build
 
+# Starting in Fedora 28 and RHEL 8, automatic shebang (#!) munging was added.
+# We don't want this in our software and it will interfere with third-party
+# dependencies that we don't control. See
+# https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/message/2PD5RNJRKPN2DVTNGJSBHR5RUSVZSDZI/
+# for more info.
+%undefine __brp_mangle_shebangs
+
 <% @package_overrides.each do |var| %>
 <%= var %>
 <% end -%>


mwaggett:
I was able to reproduce the original build error on el 8 and then build successfully from https://github.com/puppetlabs/pe-installer-vanagon/pull/65 with this change.
I also built on el 7 and ubuntu-18.04 (even though this change doesn't affect debs lol), for good measure.

resources/rpm/project.spec.erb
```diff
@@ -22,6 +22,12 @@
 # to resolve: "ERROR: No build ID note found"
 %undefine _missing_build_ids_terminate_build
 
+# Starting in Fedora 28 and RHEL 8, automatic shebang (#!) munging was added.
+# We must disable this in order to use the appropriate python executable. See
```
shrug:
```suggestion
# We don't really want this on our own software, and it will also complain about third-party deps that we do not control. See
```

resources/rpm/project.spec.erb
```diff
@@ -22,6 +22,13 @@
 # to resolve: "ERROR: No build ID note found"
 %undefine _missing_build_ids_terminate_build
 
+# Starting in Fedora 28 and RHEL 8, automatic shebang (#!) munging was added.
+# We don't want this in our software and it will interfere with third-party
+# dependencies that we don't control. See
```
mwaggett: @shrug How's this?

PR #604, opened 2019-05-09
(PA-2670) Add support for `get_version_forced` in the vanagon component

underscorgan:
This will allow us to force an additional clone of git sources to
determine the version while the component is still being processed. This
is needed to allow for including the component version in the gemspec
file for facter, puppet, and hiera so that ruby correctly identifies
those as installed gems.

### DIFF
--- a/lib/vanagon/component.rb
+++ b/lib/vanagon/component.rb
@@ -383,6 +383,23 @@ class Vanagon
       end
     end
 
+    # Force version determination for components
+    #
+    # If the component doesn't already have a version set (which normally happens for git sources),
+    # the source will be fetched into a temporary directory to attempt to figure out the version if the
+    # source type supports :version. This directory will be cleaned once the get_sources method returns
+    #
+    # @raise Vanagon::Error raises a vanagon error if we're unable to determine the version
+    def force_version
+      if @version.nil?
+        Dir.mktmpdir do |dir|
+          get_source(dir)
+        end
+      end
+      raise Vanagon::Error, "Unable to determine source version for component #{@name}!" if @version.nil?
+      @version
+    end
+
     # Prints the environment in a way suitable for use in a Makefile
     # or shell script. This is deprecated, because all Env. Vars. are
     # moving directly into the Makefile (and out of recipe subshells).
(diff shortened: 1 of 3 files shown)


lib/vanagon/component.rb
```diff
@@ -383,6 +383,23 @@ def get_patches(patch_root)
       end
     end
 
+    # Force version determination for components
+    #
+    # If the component doesn't already have a version set (which normally happens for git sources),
+    # the source will be fetched into a temporary directory to attemt to figure out the version if the
```
e-gris: Trivial: "attempt" misspelled.

lib/vanagon/component.rb
```diff
@@ -383,6 +383,23 @@ def get_patches(patch_root)
...
+    # source type supports :version. This directory will be cleaned once the get_sources method returns
+    #
+    # @raise Vanagon::Error raises a vanagon error if we're unable to determine the version
+    def force_version
+      if @version.nil?
+        Dir.mktmpdir do |dir|
+          get_source(dir)
+        end
+      end
+      raise Vanagon::Error, "Unable to determine source version!" if @version.nil?
```
e-gris: Would it be helpful to specify for which component vanagon cannot determine the source version for or would that be obvious in some other way?

underscorgan:
@e-gris updated!

spec/lib/vanagon/component_spec.rb
```diff
@@ -242,6 +242,49 @@
...
+
+    it 'returns a version if it\'s already set' do
+      expect(versioned_component).not_to receive(:get_source)
+      expect(versioned_component.force_version).to eq('1.2.3')
+    end
+
+    it 'fetches sources if no version is set' do
+      clone = double(Git::Base)
+      allow(::Git).to receive(:clone).and_return(clone)
+      allow(clone).to receive(:describe).and_return('1.2.3')
```
mwaggett: I feel like testing with something other than `1.2.3` would make it clearer that this is getting a different version than the one explicitly set in the test above.

e-gris:
Holding on merge until comments from @mwaggett are resolved.

underscorgan:
updated

underscorgan:
done

PR #607, opened 2019-07-04
(VANAGON-153) Update version in SMF manifest

GabrielNagy:
When upgrading the puppet-agent package, Solaris 11 doesn't automatically restart a service if the SMF manifest contains no changes.

This commit runs a sed on the target service(s) to change the current version (which is always 1) to the puppet-runtime version (since `version` is expected to be an integer value, per SMF documentation).

This commit also changes the Solaris sed to `/usr/gnu/bin/sed` which is the GNU implementation, so that we benefit from in-place editing and extended regular expressions.

### DIFF
--- a/lib/vanagon/component/dsl.rb
+++ b/lib/vanagon/component/dsl.rb
@@ -186,6 +186,8 @@ class Vanagon
           target_mode = '0644'
           default_mode = '0644'
         when "smf"
+          # modify version in smf manifest so service gets restarted after package upgrade
+          @component.install << %{#{@component.platform.sed} -ri 's/(<service.*version=)(".*")/\\1"#{@component.settings[:puppet_runtime_version]}"/' #{service_file}}
           target_service_file = File.join(@component.platform.servicedir, service_type.to_s, "#{service_name}.xml")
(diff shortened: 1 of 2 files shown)


lib/vanagon/component/dsl.rb
```diff
@@ -186,6 +186,8 @@ def install_service(service_file, default_file = nil, service_name = @component.
           target_mode = '0644'
           default_mode = '0644'
         when "smf"
+          # modify version in smf manifest so service gets restarted after package upgrade
+          @component.install << %{#{@component.platform.sed} -ri 's/(<service.*version=)(".*")/\\1"#{@component.settings[:puppet_runtime_version]}"/' #{service_file}}
```
gimmyxd: I'm not sure we should rely on the runtime_version as you may change something in pxp-agent but if there are no changes on runtime, the version will not be bumped. Same for changing something in runtime that does not affect pxp-agent. It seems that we would mostly do arbitrarily restarts of the services

gimmyxd: Each nightly puppet agent build is promoted in PE and tested. Of the pxp-agent will be built with the same runtime version, when it ends up in PE, it will not be restored and that could result in the same failures

PR #613, opened 2019-08-13
Add ruby.yml for github actions

shrug:


shrug:
This is the default ruby config that was suggested, we may wan to look at/tune it

PR #616, opened 2019-09-06
(maint) Update gem publish trigger

underscorgan:
Was able to confirm this works.


underscorgan:
I pushed a temp tag and was able to trigger this action. I miscalculated how long it was going to take for the action to run and deleted the tag before the gem built, so the test gem ended up with a different version, but I'm confident this is working.

PR #624, opened 2019-09-25
(maint) Prepare for 0.15.29 release

sarameisburger:


sarameisburger:
merging because nights watch needs a release and I don't want them blocked

PR #625, opened 2019-10-10
[WIP] Rubocop updates

shrug:


vanagon.gemspec
```diff
@@ -13,7 +13,7 @@ Gem::Specification.new do |gem|
   gem.email    = 'info@puppet.com'
   gem.homepage = 'http://github.com/puppetlabs/vanagon'
   gem.specification_version = 3
-  gem.required_ruby_version = '~> 2.1'
+  gem.required_ruby_version = '~> 2.4'
```
shrug: Do we actually want to require 2.4 or above? I'm not sure about this bit

e-gris: Likely minimum should be whatever we have available on vmpooler machines.

mwaggett: At this point, packaging requires at least 2.3 (via transitive dependencies), so probably don't want to go below that..

.rubocop.yml
```diff
@@ -171,6 +175,27 @@ Layout/CaseIndentation:
...
+  Enabled: false
+
+Style/TrailingCommaInArrayLiteral:
+  Enabled: false
+
+Style/TrailingCommaInHashLiteral:
+  Enabled: false
+
+Style/UnneededCondition:
+  Enabled: false
```
mwaggett: should this be true? seems like a good thing to catch..

Gemfile
```diff
@@ -27,7 +27,7 @@ group(:development, :test) do
   gem 'packaging', *location_for(ENV['PACKAGING_LOCATION'] || '~> 0.99')
   gem 'rake', require: false
   gem 'rspec', '~> 3.0', require: false
-  gem 'rubocop', "~> 0.52.1", require: false
+  gem 'rubocop', "~> 0.75", require: false
```
mwaggett: Do we want to go from three digits to two here? I don't know what kinds of changes go into their Y releases, but seeing as there's no 1.0 yet, we might wanna be more careful? 
Or this could help us stay more up-to-date with rubocop, if we make changes as we go..
Things to think about..

.rubocop.yml
shrug: with this enabled:
```lib/vanagon/common/user.rb:7:24: C: Style/UnneededCondition: Use double pipes || instead.
        @group = group ? group : @name
                       ^^^^^^^^^
lib/vanagon/platform.rb:250:7: C: Style/UnneededCondition: This condition is not needed.
      if instance_variable_get("@#{key}") ...
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^```

Gemfile
shrug: looking at https://github.com/rubocop-hq/rubocop/releases it seems like they rarely go beyond a .1 so it's probably ok.

.rubocop.yml
mwaggett: oh weird

PR #629, opened 2019-10-30
(maint) Add Ubuntu and Debian utility matchers

GabrielNagy:
~~This commit adds a utility matcher to determine whether the platform uses the system provided toolchain as opposed to pl-build-tools.~~

~~This is useful in the long term when adding new platforms to puppet-runtime and puppet-agent, because we avoid code duplication on building various C++ projects. It should also be future proof (except for when we are adding a new OS altogether).~~

Changed this PR to only add Ubuntu and Debian OS matchers. We'll rework the system toolchain/pl-build-tools issue inside puppet-runtime/puppet-agent.

### DIFF
--- a/lib/vanagon/platform.rb
+++ b/lib/vanagon/platform.rb
@@ -344,6 +344,20 @@ class Vanagon
       return !!@name.match(/^fedora-.*$/)
     end
 
+    # Utility matcher to determine is the platform is a debian variety
+    #
+    # @return [true, false] true if it is a debian variety, false otherwise
+    def is_debian?
+      return !!@name.match(/^debian-.*$/)
+    end
+
+    # Utility matcher to determine is the platform is a ubuntu variety
+    #
+    # @return [true, false] true if it is a ubuntu variety, false otherwise
+    def is_ubuntu?
+      return !!@name.match(/^ubuntu-.*$/)
+    end
+
     # Utility matcher to determine is the platform is an aix variety
     #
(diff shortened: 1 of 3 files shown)


lib/vanagon/platform.rb
```diff
@@ -436,6 +450,18 @@ def is_cross_compiled_linux?
       return (is_cross_compiled? && is_linux?)
     end
 
+    # Utility matcher to determine whether the platform uses the system provided toolchain
+    # as opposed to pl-build-tools.
+    #
+    # @return [true, false] true if it is uses the system toolchain, false otherwise
+    def uses_system_toolchain? # rubocop:disable Metrics/PerceivedComplexity
```
mwaggett: My only hesitation with this is that it feels pretty Puppet-specific, whereas vanagon _theoretically_ should be more general than that. 
I do, however, understand the desire to not repeat this code across several projects. I guess I'm wondering if there's a different shared place for these settings? like maybe a shared puppet-runtime setting?
But if we already have a bunch of stuff like this in vanagon, I guess I won't stop y'all from adding more.. 
Just something to consider..

gimmyxd: Agree with this. Maybe we can find a better way around this.

GabrielNagy: @mwaggett fair enough, I'll remove this method and we'll reverse the logic in puppet-runtime and agent instead.

I'll repurpose this PR to only have the Debian/Ubuntu OS matchers.

PR #630, opened 2019-11-05
(VANAGON-159) Updates to `parse_and_rewrite` method

underscorgan:
* If there are no rewrite rules specified, don't do any processing on
the URI, just return it directly.
* If there are rewrite rules and the URI begins with `git:http` (to
specify a git source with an http url), substitute `git:http` with
`http` since fustigit gets rid of additional `/`. **Note** This will
result in deprecation warnings, and will break at some point in the
future.
* If there are rewrite rules and the URI doesn't begin with `git:http`,
keep existing behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,15 @@ This changelog adheres to [Keep a CHANGELOG](http://keepachangelog.com/).
 ## [Unreleased]
 ### Changed
 - (VANAGON-157) Make curl calls fail on error
+
+### Added
 - Add Debian and Ubuntu platform utility matchers.
+
+### Fixed
+- (VANAGON-159) Update `parse_and_rewrite` method. It will now do no processing
+  if there aren't any rewrite rules. If there are rewrite rules and you're
+  specifying your git source with `git:http[...]`, this method will sub out the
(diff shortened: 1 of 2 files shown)


GabrielNagy:
Looks good:

```
Attempting to fetch from canonical URL "git:https://github.com/puppetlabs/puppetlabs-augeas_core.git"
Cloning Git repo 'https://github.com/puppetlabs/puppetlabs-augeas_core.git'
Successfully cloned 'puppetlabs-augeas_core'
Checking out 'refs/tags/1.0.4' from Git repo 'puppetlabs-augeas_core'
Attempting to update submodules for repo 'puppetlabs-augeas_core'
Nothing to verify for 'puppetlabs-augeas_core' (using Git reference 'refs/tags/1.0.4')
```

mwaggett:
I don't know if it's a good idea to make updates to deprecated methods.. but I guess oh well.

underscorgan:
@mwaggett yeah, I'm not entirely sure why we're ending up in that method if there aren't any rewrite rules, but tbh this may have been a bug with the initial implementation of the `git:` prefix so fixing seems good.

PR #632, opened 2019-11-11
(VANAGON-72) Access `settings` hash from platforms

underscorgan:
This updates the `settings` hash to be initialized in the platform and
then copied to the project and components. This will allow the platform
to access the settings hash and set settings to be shared with the
project/component.

### DIFF
--- a/lib/vanagon/project.rb
+++ b/lib/vanagon/project.rb
@@ -139,7 +139,7 @@ class Vanagon
       @components = []
       @requires = []
       @directories = []
-      @settings = {}
+      @settings = platform.settings
       # Environments are like Hashes but with specific constraints
       # around their keys and values.
       @environment = Vanagon::Environment.new
@@ -756,7 +756,8 @@ class Vanagon
         # We don't want to load any of the upstream components, so we're going to
         # pass an array with an empty string as the component list for load_project
         no_components = ['']
-        upstream_project = Vanagon::Project.load_project(upstream_project_name, File.join(working_directory, upstream_source.dirname, "configs", "projects"), platform, no_components)
+        upstream_platform = Vanagon::Platform.load_platform(platform.name, File.join(working_directory, upstream_source.dirname, "configs", "platforms"))
+        upstream_project = Vanagon::Project.load_project(upstream_project_name, File.join(working_directory, upstream_source.dirname, "configs", "projects"), upstream_platform, no_components)
         @settings.merge!(upstream_project.settings)
         upstream_project.cleanup
       end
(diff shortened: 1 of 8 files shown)


lib/vanagon/project.rb
```diff
@@ -756,7 +756,8 @@ def load_upstream_settings(upstream_project_name, upstream_git_url, upstream_git
         # We don't want to load any of the upstream components, so we're going to
         # pass an array with an empty string as the component list for load_project
         no_components = ['']
-        upstream_project = Vanagon::Project.load_project(upstream_project_name, File.join(working_directory, upstream_source.dirname, "configs", "projects"), platform, no_components)
+        upstream_platform = Vanagon::Platform.load_platform(platform.name, File.join(working_directory, upstream_source.dirname, "configs", "projects"))
```
mwaggett: Should this reference the "platforms" directory, rather than "projects"?

underscorgan: oo good catch

spec/lib/vanagon/project_spec.rb
```diff
@@ -6,6 +6,12 @@
 describe 'Vanagon::Project' do
   let(:component) { double(Vanagon::Component) }
   let(:configdir) { '/a/b/c' }
+  let(:platform) do
+    OpenStruct.new(:settings => {})
+  end
+  let(:upstream_platform) do
+    OpenStruct.new(:settings => {})
+  end
```
mwaggett: What's the purpose of having both of these, since they're the same?

underscorgan: Ruby makes shallow copies of objects so we need two platform hashes for use in stubbing to load the settings correctly.

mwaggett: I'm not sure what shallow copy means / how it's relevant here. Are you saying that if we had a single platform object, the setting would perpetrate through all the examples? Is that bad?

underscorgan: The shallow copy means settings changes in any one place would persist for all of them. In the code we make a new platform so we don't overwrite things unintentionally (https://github.com/puppetlabs/vanagon/pull/632/files#diff-5fa3b3d3ed3dd87653614299456ab680R759). In the tests though we stub out the load_platform/load_project calls so we need to have two separate platforms we can use to test that variables are being overwritten properly

mwaggett: ookay, yeah, I also wasn't understanding what `platform` was originally referring to here https://github.com/puppetlabs/vanagon/pull/632/files#diff-5fa3b3d3ed3dd87653614299456ab680L759 , or how it's being used..

mwaggett:
Did you wanna squash those commits?

underscorgan:
@mwaggett done!

PR #639, opened 2020-04-05
VANAGON-161 Including support to clone options

jmarcelos:
Adding support for --branch or other options in Git clone

### DIFF
--- a/lib/vanagon/component/dsl.rb
+++ b/lib/vanagon/component/dsl.rb
@@ -333,6 +333,13 @@ class Vanagon
         @component.options[:ref] = the_ref
       end
 
+      # Sets the clone options for git repository
+      #
+      # @param the_clone_options [Map] branch, depth to checkout for a git source
+      def clone_options(clone_options)
+        @component.options[:clone_options] = clone_options
+      end
+
       # Set a build dir relative to the source directory.
       #
       # The build dir will be created before the configure block runs and configure/build/install commands will be run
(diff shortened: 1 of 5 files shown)


jmarcelos:
@puppetcla 
Sorry, The Contributor License Agreement it is signed now

lib/vanagon/component/source/git.rb
```diff
@@ -55,6 +55,7 @@ def initialize(url, workdir:, **options)
           @url = URI.parse(url.to_s)
           @ref = opts[:ref]
           @workdir = File.realpath(workdir)
+          @clone_options = options[:clone_options] ||= {}
```
underscorgan, line 58: should probably use `opts` here instead of `options`

lib/vanagon/component/dsl.rb
```diff
@@ -333,6 +333,13 @@ def ref(the_ref)
         @component.options[:ref] = the_ref
       end
 
+      # Sets the clone options for git repository
+      #
+      # @param the_clone_options [Map] branch, depth to checkout for a git source
+      def clone_options(clone_options)
+        @component.options[:clone_options] = clone_options
+      end
```
underscorgan, line 341: I think it would be more consistent with the rest of the vanagon codebase if we had clone_option take in an option and the value we want that set to, and use that to populate the `:clone_options` hash. Maybe something like this (I haven't had a chance to test yet)
```suggestion
      def clone_option(option, value)
        option = option.to_sym
        @component.options[:clone_options][option] = value
      end
```

Then in the component have something like
```
pkg.clone_option 'branch', 'foo'
pkg.clone_option 'depth', 50
```

If you have the ability to test out this suggestion that would be great, if not I can take a look in the next day or two

jmarcelos: It works and test added

PR #642, opened 2020-04-28
(RE-13396) Fix debug_package handling in rpmbuilds

e-gris:


resources/rpm/project.spec.erb
```diff
@@ -1,4 +1,4 @@
-%global debug_package %{nil}
+%define debug_package %{nil}
```
underscorgan, line 1: we should make sure this doesn't break things for other platforms, I feel like this was super finicky in the past :\

specifically, the build may work but we should check and make sure the package sizes haven't changed dramatically

PR #643, opened 2020-04-28
(RE-13396) Fix debug_package handling in rpmbuilds

e-gris:
I used PDK as the test package from the Jira ticket.

I built on rhel-7 and rhel-8 using the %global macro definition.

On rhel-7 the rpm was ~254Mb
On rhel-8 the rpm did not build

I built on rhel-7 and rhel-8 using the %define macro definition

On rhel-7 the rpm was ~242Mb
On rhel-8 the rpm was ~254Mb

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,6 +1,6 @@
 ---
 AllCops:
-  TargetRubyVersion: 2.3
+  TargetRubyVersion: 2.4
   Exclude:
     - '**/*.erb'
     - 'spec/**/*'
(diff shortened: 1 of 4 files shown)


.rubocop.yml
```diff
@@ -1,6 +1,6 @@
 ---
 AllCops:
-  TargetRubyVersion: 2.3
+  TargetRubyVersion: 2.4
```
underscorgan, line 3: Yesterday I pinned the version of rubocop rather than bump the minimum ruby version, I think if you rebase on master and just get rid of the commit fixing rubocop stuff this should be good to go

e-gris: We had the same thought at the same time! :-)

PR #646, opened 2020-04-29
(maint) el-8 __debug_package breakage

e-gris:
The __debug_package macro handling on el8 suddenly broke. Not sure
why. This will hack around the problem until we understand it more.


e-gris:
Tested by running pe-r10k-vanagon with VANAGON_LOCATION set to git@github.com:e-gris/vanagon#el8-woes

el-8 and el-7 work correctly

PR #648, opened 2020-05-06
(maint) Build el-8 packages without build-id files

sarameisburger:
this was causing collision errors


sarameisburger:
i'll add a changelog update after I confirm this is all we need

PR #650, opened 2020-05-12
(maint) Output valid JSON build metadata

lucywyman:
As part of the build process we run `echo -e "#{metadata}" >
output/foo.json` where `metadata` is a JSON string. This doesn't
consistently preserve double quotes, causing the resulting file to be
invalid JSON. This PR escapes the double quotes in the metadata before
echoing to ensure the resulting json file can be consumed.

### DIFF
--- a/lib/vanagon/platform.rb
+++ b/lib/vanagon/platform.rb
@@ -504,6 +504,7 @@ class Vanagon
       end
 
       metadata.gsub!(/\n/, '\n')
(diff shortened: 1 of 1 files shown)


underscorgan:
Yeah, this definitely fixes the more immediate issue, but I'm curious if the yaml file might exist elsewhere and could be copied over rather than recreated here? I'm going to poke around a bit more with that today.

underscorgan:
Closing this since #651 was merged

PR #651, opened 2020-05-12
(maint) Don't regenerate build metadata for archives

underscorgan:
We were previously regenerating the build_metadata when building
archives (i.e. for puppet-runtime). This change will make it so that
when we're writing the metadata out to the `ext` directory we also write
it out to the vanagon workdir so we have access to it on the build
host.

Prior to these changes, the build metadata file that landed in the
`output` directory contained invalid json. This fixes that issue.


underscorgan:
@lucywyman This should fix the invalid json!

PR #652, opened 2020-05-15
(RE-13303) Add support for signing additional files for windows

underscorgan:
This will allow us to do things like sign installer powershell scripts
that get bundled into the MSI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ This changelog adheres to [Keep a CHANGELOG](http://keepachangelog.com/).
 ### Fixed
 - (maint) Fix for invalid build metadata JSON when generating compiled archives.
 
+### Added
+- (RE-13303) Add the ability to sign additional files on windows during the MSI
+build step.
+
 ## [0.15.37] - released on 2020-05-06
 ### Changed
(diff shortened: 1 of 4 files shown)


underscorgan:
@e-gris hm, I was expecting defining the signing command in the projects would make sense, but perhaps something inherited from packaging would make sense? Mostly want to avoid having too much tied directly to our infrastructure in vanagon. 

And yea, the signing of aux files is different from signing the MSIs, the tool we use to sign MSIs doesn't support signing arbitrary files (only MSI and JAR iirc)

e-gris:
I see the conflict. On one side we'd want to relieve the user from knowing how to sign by providing the signing mechanism. On the other side, keeping internal implementation separate makes sense.

Delegation to an internal-only signing gem I think would make sense but that's a fair bit of work (& I recall trying that once).

sanfrancrisko:
Hi folks - we're (IAC Team) doing a bit of a tidy up of some PRs. This PR will unblock this one from being merged: https://github.com/puppetlabs/pdk-vanagon/pull/263

Just wondering if it is able to be merged and if so, would someone be so kind to ping us on the above PR when it is?

PR #656, opened 2020-08-27
(VANAGON-100) Match rsync behavior in docker_cp_globs_from

Sharpie:
This commit updates the logic in `Vanagon::Engine::Docker#cp_globs_from`
to match the behavior of `Vanagon::Utilities#rsync_from` by copying
directory contents instead of directories when both the source path
and target path are directories. E.g. `cp_globs_from('foo/', 'bar/')`
is treated as `cp_globs_from('foo/*', 'bar/')`, which matches the
behavior of `rsync -r`.


Sharpie:
This is a small fix-up for changes in PR puppetlabs/vanagon#655.

PR #658, opened 2020-09-11
(RE-13660) Add new 'vanagon' command and cli parser

e-gris:
Rather than vanagon use standalone 'build', 'sign', 'ship', etc. statements, I've added a new 'vanagon' command so that
'vanagon build', 'vanagon sign', 'vanagon ship', etc.  replace the above.


underscorgan (commented):
In general this looks really good! Main question is whether or not there's some clean-ish way to move the commands copied from `bin/[build|sign|ship]` somewhere where they could be called from the CLI and the existing commands just to reduce duplication and risk of code drift

e-gris:
That is totally doable. My initial pass was intended to be generally additive (modulo the deprecation warnings).

For example we can make bin/build be. 

```bash
#!/bin/bash

echo "Deprecation warning"
exec vanagon build "$@"
```
This would keep backwards compatibility.

Further, I ignored some of the other vanagon commands (inspect, render, etc.) only for initial expedience. The complete version would either incorporate them into the new cli or remove them. Not sure which, if any, were still valuable to keep.

The new CLI structure allows us to add/drop/change subcommands fairly easily.

e-gris:
The other immediately desirable/helpful thing would be to delegate the '#run' method to the subcommand objects. I just didn't get that far on Friday.

e-gris:
@underscorgan I think I'm in a good spot. Probably could use some more testing, maybe see if Night's Watch has any opinion?

underscorgan (approved):
I like this!

underscorgan:
Hiya @mihaibuzgau ! Does your team want to take a look at these potential vanagon changes?

PR #662, opened 2020-10-21
(DIO-1066) Update vanagon to go through ABS for VMs

sbeaulie:
Before this change, the ABS engine was partially implemented.
It now includes the select_target and teardown methods needed for
vanagon build lifecycle.

This engine is thus a level of abstraction that supports any downstream
engine (vmpooler, nspooler, AWS) and is forward compatible with
ondemand Vmpooler pools.

The ABS token is read from env var ABS_TOKEN or the first line of ~/.vanagon-token
or the vmfloaty config (using top-level token or the first abs type service found)

The vmpooler token optionaly can be provided through an env var VMPOOLER_TOKEN
or the second line of ~/.vanagon-token or via vmfloaty's existing vmpooler_fallback config

The ABS request is done and checked for up to an hour, with feedback given on the
availability of resources. Teardown is done at the end using the job_id.

Users can track vm usage via bitbar or floaty if they provide a vmpooler token.

### DIFF
--- a/lib/vanagon/engine/always_be_scheduling.rb
+++ b/lib/vanagon/engine/always_be_scheduling.rb
@@ -143,10 +143,7 @@ class Vanagon
 
         warn "Reading ABS token from: #{path}"
         contents = File.read(absolute_path).chomp
-        lines = []
-        contents.each_line do |l|
-          lines << l.chomp
-        end
+        lines = contents.each_line.map(&:chomp)
 
         abs = lines.shift
         @token_vmpooler = lines.shift
@@ -221,7 +218,7 @@ class Vanagon
           end
         end
         warn message
-        abs_token
+        return abs_token
       end
       private :read_vmfloaty_token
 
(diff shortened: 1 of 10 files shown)


sbeaulie:
@beechtom would you mind testing this code for your bolt-vanagon project? I tried it myself with `bundle exec ./bin/vanagon build puppet-bolt fedora-30-x86_64 --engine always_be_scheduling --preserve never` and it was successful.

mattkirby (approved):

beechtom (approved):

lib/vanagon/engine/always_be_scheduling.rb
```diff
@@ -103,6 +111,272 @@ def build_host_name
...
+        end
+
+        res_body = translated(response, @saved_job_id)
+        if res_body["ok"]
+          @target = res_body[build_host_name]['hostname']
+          Vanagon::Driver.logger.info "Reserving #{@target} (#{build_host_name}) [#{@token ? 'token used' : 'no token used'}]"
+        else
+          pooler = ''
+        end
+        pooler
```
e-gris, line 283: Prefer guardian clauses over if/else:

```ruby
return '' unless res_body['ok']
 
@target = res_body[build_host_name]['hostname']
Vanagon::Driver.logger.info "Reserving #{@target} (#{build_host_name}) [#{@token ? 'token used' : 'no token used'}]"
return pooler
```

e-gris (changes requested):
Looks great. Only the CHANGELOG is a required change.

The rest are stylistic comments which are part of a direction I'm hoping to nudge this (and other RE projects) into. 

They are wholly optional and I didn't mark each one.

e-gris (approved):

e-gris:
Merge at will.

PR #664, opened 2020-11-18
(maint) Overwrite unchanged config files

GabrielNagy:
Configuration files installed with `install_configfile` have the
`.pristine` extension appended to them, as a guard to avoid clobbering
existing configfiles.

OS X and Solaris 10 use postinstall scripts to manage configuration
files, removing the extension if a non-pristine config file does not
exist.

However, we don't need to keep the pristine files if there are no
differences between the pristine file and the target file. So, update
the logic to also run a `diff` between the two files. If there are no
differences, go ahead and override the target file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 This changelog adheres to [Keep a CHANGELOG](http://keepachangelog.com/).
 
 ## [Unreleased]
+### Fixed
+- (maint) Pristine config files are no longer kept if there are no differences
+  between the pristine and the target config file (OS X and Solaris 10).
 
 ## [0.17.0] - released 2020-11-02
 ### Added
--- a/resources/osx/postinstall.erb
+++ b/resources/osx/postinstall.erb
@@ -5,7 +5,7 @@
 <%- get_configfiles.each do |config|
 dest_file = config.path.gsub(/\.pristine$/, '') -%>
 
-if [ -f "<%= dest_file %>" ]; then
+if [ -f "<%= dest_file %>" ] && ! diff "<%= config.path %>" "<%= dest_file %>" > /dev/null; then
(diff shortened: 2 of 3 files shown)


GabrielNagy:
/cc @puppetlabs/night-s-watch 

Tested this manually on Solaris 10 and OS X, the previous implementation is not necessarily a bug, but it leaves unneeded files around.

luchihoratiu (approved):

e-gris (approved):

joshcooper:
I think this may be causing puppet-agent upgrades on Solaris 10 to clobber `/etc/puppetlabs/puppet/puppet.conf`. In CI, the agent has the certname and server settings in the `main` section:

```ini
[main]
server = deputy-syringa.delivery.puppetlabs.net
certname = idyllic-stint.delivery.puppetlabs.net
# This file can be used to override the default puppet settings.
...
```

Postupgrade, the file is clobbered with the version from the new package, causing future `puppet agent -t` runs to try to connect to `https://puppet:8140/...` which of course fails.

Solaris 11 doesn't have this problem, so I'm guessing there's an issue with this PR?

Or there's something about the `diff` command on Solaris 10?

@nmburgan also was wondering why we're calling [`mv` on OSX](https://github.com/GabrielNagy/vanagon/blame/db231f0ef9be5ef1d93eff644c043018d5d8bfa5/resources/osx/postinstall.erb#L11) but [`cp` on Solaris](https://github.com/GabrielNagy/vanagon/blame/db231f0ef9be5ef1d93eff644c043018d5d8bfa5/resources/solaris/10/postinstall.erb#L15).

GabrielNagy:
Good catch @joshcooper, it appears I have broken some stuff™. I don't know about the `cp` vs `mv` thing, but I'm working on a fix for the accidental clobbering: https://github.com/puppetlabs/vanagon/compare/master...GabrielNagy:fix-pristine-configfiles

PR #667, opened 2020-11-26
(PA-3497) build arm64 debian packages for aarch64

ciprianbadescu:


ciprianbadescu:
> This is way out of my normal circle of attention so mostly curious: why do we have to special case the translation of aarch64 into arm64? Why not just stick to one naming?

`arm64` is the name Debian is using while `aarch64` is the name used by Linux Kernel, GNU compilers, puppet-runtime/puppet-agent build process.

The solution on translating `aarch64` to `arm64` to dpkg build command was considered the best compromise.

e-gris (approved):

underscorgan (approved):

underscorgan (approved):
Feel free to merge when you're ready!

PR #673, opened 2020-12-09
(maint) fix target argument handling

ciprianbadescu:
After default engine was changed to 'always_be_scheduling' target
argument was ignored

### DIFF
--- a/lib/vanagon/driver.rb
+++ b/lib/vanagon/driver.rb
@@ -74,6 +74,8 @@ class Vanagon
     end
 
     def load_engine_object(engine_type, platform, target)
+      engine_type = 'base' if target
+
       require "vanagon/engine/#{engine_type}"
       @engine = Object::const_get("Vanagon::Engine::#{camelize(engine_type)}")
         .new(platform, target, remote_workdir: remote_workdir)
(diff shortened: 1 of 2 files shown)


donoghuc:
I am seeing what i think is a similar issue https://cinext-jenkinsmaster-enterprise-prod-1.delivery.puppetlabs.net/job/enterprise_pe-bolt-server_pkg-van-ship_daily-2019.8/BUILD_TARGET=ubuntu-18.04-amd64,SLAVE_LABEL=beaker/7/console with the 0.18.0 release

lib/vanagon/driver.rb
```diff
@@ -74,6 +74,8 @@ def pick_engine(options) # rubocop:disable Metrics/PerceivedComplexity
     end
 
     def load_engine_object(engine_type, platform, target)
+      engine_type = 'base' if target
```
donoghuc, line 77: It looks like the bug is in this method https://github.com/puppetlabs/vanagon/commit/eaaea24ed314b00759154f6d86abe5ef6df6234f#r44961052 which i think could use a refactor anyway to be easier to reason about.

e-gris:
@ciprianbadescu , do the changes from @donoghuc work for you? If so, let's go with those. Otherwise, this LGTM.

e-gris (approved):

PR #676, opened 2020-12-09
(maint) Fix #pick_engine method

e-gris:
My first try at isolating the pick_engine method was buggy. Cas provided
us a better one.

### DIFF
--- a/lib/vanagon/driver.rb
+++ b/lib/vanagon/driver.rb
@@ -56,21 +56,19 @@ class Vanagon
     end
 
     def pick_engine(options) # rubocop:disable Metrics/PerceivedComplexity
-      default_engine = 'always_be_scheduling'
-
-      # Use the explicitly configured engine if no target was provided.
-      return options[:engine] if options[:engine] && !options[:target]
-
-      # If the configured engine matches the default engine, use it
-      return options[:engine] if options[:engine] == default_engine
-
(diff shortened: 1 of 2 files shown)


donoghuc (approved):
This looks better to me. I dont have a local repro case though i was using CI...

e-gris:
I'm away for the next couple of days, so I'll run some tests and release next week if we don't find any issues.

lib/vanagon/driver.rb
```diff
@@ -56,26 +56,22 @@ def filter_out_components(only_build)
     end
 
     def pick_engine(options) # rubocop:disable Metrics/PerceivedComplexity
```
donoghuc, line 58: Hopefully this appeases rubocop and we dont have to disable it... i think generally having a ton of return statements in a ruby method will rasie some eyebrowns.

e-gris: I tried without and rubocop complained. 

Typically, I like guardian clauses over if/elses but I agree my previous implementation was too confusing.

donoghuc: ah, i had that disabled locally in a `.rubocop.yml`. sorry.

ciprianbadescu (approved):

underscorgan (approved):

PR #678, opened 2021-01-07
(maint) Fix pristine configfiles

GabrielNagy:
On Solaris 10, postinstall scripts are executed by the `sh` interpreter,
as the package manager does not seem to respect the shebang line at the
top of the file.

This causes the `!` sign to not be recognized, and the configfiles are
clobbered regardless of the differences between the installed config
file and the pristine config file.

Sample output after upgrading puppet-agent on Solaris 10:

```
/var/sadm/pkg/puppet-agent/install/postinstall: !: not found
/var/sadm/pkg/puppet-agent/install/postinstall: !: not found
/var/sadm/pkg/puppet-agent/install/postinstall: !: not found
/var/sadm/pkg/puppet-agent/install/postinstall: !: not found
/var/sadm/pkg/puppet-agent/install/postinstall: !: not found
/var/sadm/pkg/puppet-agent/install/postinstall: !: not found
/var/sadm/pkg/puppet-agent/install/postinstall: !: not found
/var/sadm/pkg/puppet-agent/install/postinstall: !: not found
/var/sadm/pkg/puppet-agent/install/postinstall: !: not found
/var/sadm/pkg/puppet-agent/install/postinstall: !: not found
rootx0:0:Super-User:/:/usr/bin/bash
root::0:
Installation of <puppet-agent> was successful.
```

To avoid these types of errors, do not use `!` in the script, and remove
the pristine files instead of tampering with the non-pristine ones.

This also takes care of the case when `diff` is not present on the
system, causing the check to noop instead of clobbering.

Also, check if a file or directory exists before running `chmod` on it on Solaris 10.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ This changelog adheres to [Keep a CHANGELOG](http://keepachangelog.com/).
 
 ### Fixed
 - (maint) Fix bugs in pick_engine
+- (maint) Fix pristine config files clobbering (Solaris 10)
 
 
 ## [0.18.1] - released 2020-12-09
--- a/resources/osx/postinstall.erb
+++ b/resources/osx/postinstall.erb
@@ -5,8 +5,12 @@
 <%- get_configfiles.each do |config|
 dest_file = config.path.gsub(/\.pristine$/, '') -%>
 
-if [ -f "<%= dest_file %>" ] && ! diff "<%= config.path %>" "<%= dest_file %>" > /dev/null; then
-  echo "Detected file at '<%= dest_file %>'; updated file at '<%= config.path %>'."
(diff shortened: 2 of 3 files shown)


GabrielNagy:
puppet-agent 6 packages built with this vanagon: http://builds.delivery.puppetlabs.net/puppet-agent/48adfa5f14f5636d578d4ebfd91339475793db00/artifacts/

The fix can be tested by first installing puppet-agent 6.19.1 (unaffected), making changes to a config file (e.g. puppet.conf), then installing the package from above. The config file should remain unchanged.

e-gris (commented):
I've not touched any of the osx nor solaris postinstall so I can't say my review is authoritative.

However, I looked this over and the code changes look good to me.

underscorgan (approved):

PR #683, opened 2021-01-16
(RE-13837) Hack around Git.ls_remote failure

e-gris:
Create a slightly sloppy workaround for a failure in Git.ls_remote

When Git.ls_remote calls the git command-line, under moderately common circumstances, ssh can be involved and it can print something like:

    Warning: Permanently added 'github.com,192.30.255.113' (RSA) ...

Git.ls_remote, does not anticipate this and attempts to parse it as actual `git ls-remote` output.

This results in a `NoMethodError: undefined method 'split' for nil:NilClass` because the [#split](https://github.com/ruby-git/ruby-git/blob/bd026d39b17e192e8a710744153ed28139cc64a7/lib/git/lib.rb#L502) fails on the unexpected message.

I think that leaves us with 3 possible solutions.

1. Not use ruby-git for this operation (which was my choice).
1. monkey-patch the Git.ls_remote method (which was my 2nd choice)
1. catch the 'NoMethodError' exception and try to sort it out (yuck!)

It would be good citizenry to submit a PR to ruby-git which I might do to correctly fix this problem. I may endeavor to do so.

### DIFF
--- a/lib/vanagon/component/source/git.rb
+++ b/lib/vanagon/component/source/git.rb
@@ -54,7 +54,8 @@ class Vanagon
               return false unless $?.exitstatus.zero?
               return true
             end
-          rescue RuntimeError
+          rescue RuntimeError # rubocop:disable Lint/ShadowedException
+            ## rubocop is right by flagging this but I think we need it in this case.
             return false
           rescue Timeout::Error
             return false
--- a/lib/vanagon/component/source/git.rb
+++ b/lib/vanagon/component/source/git.rb
@@ -26,14 +26,38 @@ class Vanagon
           #        git command has failed. Useful in instances where a URL
           #        prompts for credentials despite not being a git remote
           # @return [Boolean] whether #url is a valid Git repo or not
+
+          # [RE-13837] This ought to be the way to do this. Unfortunately,
+          # there's a bug in Git.ls_remote that when ssh prints something like
+          #  Warning: Permanently added 'github.com,192.30.255.113' (RSA)
+          # Git.ls_remote attempts to parse it as actual git output and fails
+          # with: NoMethodError: undefined method `split' for nil:NilClass
+          #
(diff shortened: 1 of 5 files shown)


vanagon.gemspec
```diff
@@ -18,7 +18,7 @@ Gem::Specification.new do |gem|
   gem.add_runtime_dependency('docopt')
   # Handle git repos responsibly
   # - MIT licensed: https://rubygems.org/gems/git
-  gem.add_runtime_dependency('git', '~> 1.3.0')
+  gem.add_runtime_dependency('git', '~> 1.8.0')
```
sbeaulie, line 21: yes!

lib/vanagon/component/source/git.rb
```diff
@@ -26,14 +26,39 @@ class << self
...
             Timeout.timeout(timeout) do
-              !!::Git.ls_remote(url)
+              Vanagon::Utilities.local_command("git ls-remote #{url} > /dev/null 2>&1")
+              return false unless $?.exitstatus.zero?
+              return true
             end
-          rescue ::Git::GitExecuteError
-            false
+          rescue RuntimeError # rubocop:disable Lint/ShadowedException
+            ## rubocop is right by flagging this but I think we need it in this case.
```
underscorgan, line 58: Is there a reason the runtime error needs to come before the timeout error?

e-gris: In short, no. What happened was the `::Git::GitExecuteError` from `Git.ls_remote` with `Vanagon::Utilities.local_command`. I swapped the rescue and the latter throws `RuntimeError` on error.

I see the value in moving it after `rescue Timeout::Error`

underscorgan: yeah, and then we can get rid of the rubocop exception :)

sbeaulie (approved):

underscorgan (approved):
merge whenever you're ready!

PR #684, opened 2021-01-26
(VANAGON-123) deb systemv and systemd

ciprianbadescu:
Debian in particular supports running modern systems under either
systemd or sysv. This adds support for passing multiple servicetypes in
your platform definition and also installing multiple services.

This PR contains the original changes from https://github.com/puppetlabs/vanagon/pull/556 with systemd check updated
to be the same as the one used by puppet systemd service provider

### DIFF
--- a/lib/vanagon/component/dsl.rb
+++ b/lib/vanagon/component/dsl.rb
@@ -167,8 +167,8 @@ class Vanagon
       #
       # @param service_file [String] path to the service file relative to the source
       # @param default_file [String] path to the default file relative to the source
+      # @param service_name [String] name of the service
       # @param options optional extra parameters
-      #        service_name [String] name of the service
       #        service_type [String] type of the service (network, application, system, etc)
       #        init_system [String] the init system on which to install service (sysv, systemd)
       #        link_target [String] executable service file should be linked to
--- a/lib/vanagon/component/dsl.rb
+++ b/lib/vanagon/component/dsl.rb
@@ -167,38 +167,43 @@ class Vanagon
       #
       # @param service_file [String] path to the service file relative to the source
       # @param default_file [String] path to the default file relative to the source
-      # @param service_name [String] name of the service
-      # @param service_type [String] type of the service (network, application, system, etc)
-      # @param link_target [String] executable service file should be linked to
-      def install_service(service_file, default_file = nil, service_name = @component.name, service_type: nil, link_target: nil) # rubocop:disable Metrics/AbcSize
-        case @component.platform.servicetype
+      # @param options optional extra parameters
+      #        service_name [String] name of the service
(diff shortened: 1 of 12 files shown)


e-gris (commented):
I read through this. I'm not terribly familiar with this part of the Vanagon code but the changes look sensible to me.

lib/vanagon/component/dsl.rb
```diff
@@ -167,52 +167,56 @@ def conflicts(pkgname, version = nil)
...
       # @param default_file [String] path to the default file relative to the source
-      # @param service_name [String] name of the service
-      # @param service_type [String] type of the service (network, application, system, etc)
-      # @param link_target [String] executable service file should be linked to
-      def install_service(service_file, default_file = nil, service_name = @component.name, service_type: nil, link_target: nil) # rubocop:disable Metrics/AbcSize
-        case @component.platform.servicetype
+      # @param options optional extra parameters
+      #        service_name [String] name of the service
+      #        service_type [String] type of the service (network, application, system, etc)
+      #        link_target [String] executable service file should be linked to
```
underscorgan, line 173: Looks like the code is using an `options[:init_system]`, can you update the comment to include that?

underscorgan (approved):
One minor comment but otherwise looks good to me!

ciprianbadescu:
Is this change breaking for code like bellow? If yes, does it worth to keep backwards compatibility?
```
  case platform.servicetype
  when 'systemd'
```
There are some occurrences in puppet repos of this usage: https://github.com/search?q=org%3Apuppetlabs+platform.servicetype&type=code

lib/vanagon/component/dsl.rb
```diff
@@ -167,52 +167,57 @@ def conflicts(pkgname, version = nil)
       #
       # @param service_file [String] path to the service file relative to the source
       # @param default_file [String] path to the default file relative to the source
-      # @param service_name [String] name of the service
-      # @param service_type [String] type of the service (network, application, system, etc)
-      # @param link_target [String] executable service file should be linked to
-      def install_service(service_file, default_file = nil, service_name = @component.name, service_type: nil, link_target: nil) # rubocop:disable Metrics/AbcSize
-        case @component.platform.servicetype
+      # @param options optional extra parameters
+      #        service_name [String] name of the service
```
underscorgan, line 171: looks like `service_name` is not part of the options hash

underscorgan (commented):
@ciprianbadescu From my reading it doesn't look like this change would affect backwards compatibility for `case platform.servicetype` . I think some projects may need updates around there when adding support for multiple service types. We should probably avoid backwards incompatible changes in general though.

ciprianbadescu:
passed ad-hoc with puppet from master and vanagon from this branch, I think is good to merge

PR #685, opened 2021-02-01
(maint) Allow VANAGON_USE_MIRROR to recognize `false`

stahnma:
Previous the VANAGON_USE_MIRROR env variable only recoginzied `n`. This
seems counter intuitive, so I've added a small adjustment to have it
also recognize `false`.

### DIFF
--- a/lib/vanagon/component.rb
+++ b/lib/vanagon/component.rb
@@ -305,7 +305,7 @@ class Vanagon
     def get_source(workdir) # rubocop:disable Metrics/AbcSize, Metrics/PerceivedComplexity
       opts = options.merge({ workdir: workdir })
       if url || !mirrors.empty?
-        if ENV['VANAGON_USE_MIRRORS'] == 'n'
+        if ENV['VANAGON_USE_MIRRORS'] == 'n' or ENV['VANAGON_USE_MIRRORS'] == 'false'
           fetch_url(opts)
         else
           fetch_mirrors(opts) || fetch_url(opts)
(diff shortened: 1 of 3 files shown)


lib/vanagon/component.rb
```diff
@@ -305,7 +305,7 @@ def deprecated_rewrite_url
     def get_source(workdir) # rubocop:disable Metrics/AbcSize, Metrics/PerceivedComplexity
       opts = options.merge({ workdir: workdir })
       if url || !mirrors.empty?
-        if ENV['VANAGON_USE_MIRRORS'] == 'n'
+        if ENV['VANAGON_USE_MIRRORS'] == 'n' or ENV['VANAGON_USE_MIRRORS'] == false
```
Sharpie, line 308: `false` needs to be quoted as environment variables are strings.

Sharpie (changes requested):

lib/vanagon/component.rb
e-gris: Maybe even?

```ruby
%w[ n no false ].include? ENV['VANAGON_USE_MIRRORS'].downcase
```

stahnma: Thanks @Sharpie. I was banging my head on that test failure for a bit. 

@e-gris That could be added. It involved a few more if statements since you can't downcase a nil object and show you have to check for that first. Not sure how important this is overall. It's not hard to add, but I also don't want to make this more complicated to follow for any future maintainers :)

jackie-kinsler (approved):

PR #689, opened 2021-04-05
(PA-3570) enable clone into a custom dirname

gimmyxd:
puppet-agent#6.x needs to ship both facter#3.x and
facter#main branches using 2 different components(
facter#3.x -> facter, facter#main -> facter-ng).
In order to do this, the component must configure
the directory where the git repository should be cloned.

This PR makes use of `dirname` attribute that can
be set at component level and exposes this to the
Git source in order to enable the ability to clone
and use a repository with a specific name.

Eg. of usage: https://github.com/puppetlabs/puppet-agent/pull/2075/files
```
pkg.load_from_json('configs/components/facter-ng.json')
pkg.dirname("facter-ng")
```

### DIFF
--- a/lib/vanagon/component.rb
+++ b/lib/vanagon/component.rb
@@ -304,7 +304,7 @@ class Vanagon
     #
     # @param workdir [String] working directory to put the source into
     def get_source(workdir) # rubocop:disable Metrics/AbcSize, Metrics/PerceivedComplexity
-      opts = options.merge({ workdir: workdir })
+      opts = options.merge({ workdir: workdir, dirname: dirname })
       if url || !mirrors.empty?
         if ENV['VANAGON_USE_MIRRORS'] == 'n' or ENV['VANAGON_USE_MIRRORS'] == 'false'
           fetch_url(opts)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ This changelog adheres to [Keep a CHANGELOG](http://keepachangelog.com/).
 ### Added
 - (VANAGON-85) Added the `clear_provisioning` method to the platform dsl to clear 
 the provisioning command array. 
+- (PA-3570) add the posibility to clone into a custom dirname
+
 
 ## [0.21.0] - released 2021-04-15
 ### Added
--- a/lib/vanagon/component/source.rb
+++ b/lib/vanagon/component/source.rb
@@ -37,6 +37,7 @@ class Vanagon
               sum: options[:sum],
               ref: options[:ref],
               workdir: options[:workdir],
+              dirname: options[:dirname],
               clone_options: options[:clone_options]
           end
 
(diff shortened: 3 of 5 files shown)


lib/vanagon/component.rb
```diff
@@ -304,7 +304,7 @@ def deprecated_rewrite_url
     #
     # @param workdir [String] working directory to put the source into
     def get_source(workdir) # rubocop:disable Metrics/AbcSize, Metrics/PerceivedComplexity
-      opts = options.merge({ workdir: workdir })
+      opts = options.merge({ workdir: workdir, dirname: dirname })
```
e-gris, line 307: I'm a little confused. Isn't `workdir` already specifying the target clone directory? Is the need just to make it more flexible?

gimmyxd: From what I've seen workdir is where all the repos are cloned. What this change is trying to achieve is the ability to clone for example  the facter repo as facter-ng. Equivalent of executing `git clone github.com/puppetlabs/facter.git facter-ng`

gimmyxd: I'm not sure if there is a way to configure the workdir per component.

e-gris: Ah ha. I see now.

gimmyxd: @e-gris from the `component.dirname` attribute documentation: 
```ruby

    # holds the expected directory name of a given component, once it's
    # been unpacked/decompressed. For git repos, it's usually the directory
    # that they were cloned to. For the outlying flat files, it'll
    # end up being defined explicitly as the string './'
    attr_accessor :dirname
```
https://github.com/puppetlabs/vanagon/blob/main/lib/vanagon/component.rb#L34-L38
https://github.com/puppetlabs/vanagon/blob/main/lib/vanagon/component/dsl.rb#L373-L393

From my understanding, there is nothing wrong in using this with the Git source, which represent and alternative of downloading an archive from web and decompress it.(when you're doing this, `dirname` lets you configure the unpacked folder name).

```
# pkg.url "http://the-internet.com/a-silly-name-that-unpacks-into-not-this.tar.gz"
# pkg.dirname "really-cool-directory"
```

The mapping in my head is like this:
```
dirname=facter-ng

# using web archives
wget facter.tar.gz; tar xf  facter.tar.gz -C $dirname

# using git repos
git clone facter.git $dirname
```

I'm not sure that this was the intent of using `dirname`, if is not, then i'm ok with figuring out what would be needed to provided this kind of flexibility.

e-gris: Yea, you're a few steps ahead of me. I'm just fussing over descriptive naming.

gimmyxd: It also supports specifying clone directory as:
```
❯ cat configs/components/facter-ng.json
{"url":"git://github.com/puppetlabs/facter.git","ref":"0cc9d00237501d5cbd18053cf583f7cdebe8a25c", "dirname": "facter-ng"}
``

gimmyxd: @e-gris or @puppetlabs/release-engineering any change to take another look at this? Would be really helpful for us to have this kind of flexibility

e-gris: No objections here.

(I think the whole #get_source method could stand to be rewritten but that's a different issue)

e-gris (approved):

ciprianbadescu (approved):

PR #690, opened 2021-04-13
(VANAGON-85) Move platform definitions into core vanagon.

jackie-kinsler:

### DIFF
--- a/lib/vanagon/platform/defaults/osx-10.14-x86_64.rb
+++ b/lib/vanagon/platform/defaults/osx-10.14-x86_64.rb
@@ -1,0 +1,22 @@
+platform "osx-10.14-x86_64" do |plat|
+  plat.servicetype 'launchd'
+  plat.servicedir '/Library/LaunchDaemons'
+  plat.codename "mojave"
+
+  plat.provision_with 'export HOMEBREW_NO_EMOJI=true'
+  plat.provision_with 'export HOMEBREW_VERBOSE=true'
+  plat.provision_with 'sudo dscl . -create /Users/test'
+  plat.provision_with 'sudo dscl . -create /Users/test UserShell /bin/bash'
+  plat.provision_with 'sudo dscl . -create /Users/test UniqueID 1001'
+  plat.provision_with 'sudo dscl . -create /Users/test PrimaryGroupID 1000'
+  plat.provision_with 'sudo dscl . -create /Users/test NFSHomeDirectory /Users/test'
+  plat.provision_with 'sudo dscl . -passwd /Users/test password'
+  plat.provision_with 'sudo dscl . -merge /Groups/admin GroupMembership test'
+  plat.provision_with 'echo "test ALL=(ALL:ALL) NOPASSWD: ALL" > /etc/sudoers.d/username'
+  plat.provision_with 'mkdir -p /etc/homebrew'
+  plat.provision_with 'cd /etc/homebrew'
+  plat.provision_with 'su test -c \'echo | /usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"\''
+  plat.provision_with 'sudo chown -R test:admin /Users/test/'
+  plat.vmpooler_template 'osx-1014-x86_64'
+  plat.output_dir File.join('apple', '10.14', 'puppet6', 'x86_64')
+end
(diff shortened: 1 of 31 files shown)


lib/vanagon/platform/dsl.rb
```diff
@@ -85,6 +85,13 @@ def respond_to_missing?(method_name, include_private = false)
         method_name.to_s.start_with?('get_') || super
       end
 
+      def inherit_from_default(name = @name)
```
e-gris, line 88: Question: is there a use case where overriding `@name` is useful? Could we get away with this?:

```ruby
def inherit_from_default
  default_file = File.join(__dir__, "/defaults/#{@name}.rb")
  [etc ...]
```

lib/vanagon/platform/dsl.rb
```diff
@@ -85,6 +85,13 @@ def respond_to_missing?(method_name, include_private = false)
...
+        default_file = File.join(__dir__, "/defaults/#{name}.rb")
```
e-gris, line 89: `File.join` should be used consistently:

```ruby
default_file = File.join(__dir__, 'defaults', "#{name}.rb")
```

lib/vanagon/platform/defaults/osx-10.15-x86_64.rb
```diff
@@ -0,0 +1,22 @@
+platform "osx-10.15-x86_64" do |plat|
+  plat.servicetype 'launchd'
+  plat.servicedir '/Library/LaunchDaemons'
+  plat.codename "catalina"
+
+  plat.provision_with 'export HOMEBREW_NO_EMOJI=true'
+  plat.provision_with 'export HOMEBREW_VERBOSE=true'
```
e-gris, line 7: Quoting-style " vs. ' mismatch. 

I think throughout all the platform-defaults files we should adopt a consistent quoting style. I'm on the fence on which would be better:

-  The typical Ruby: use single-quotes except when you must use double-quotes for variable expansion.

OR:

- (like most of the files here already): always use double quotes.

The first follows good Ruby form but looks a little awkward/fussy inside these DSL files. The second is easy and may end up looking more readable.

e-gris (commented):
A short list of minor comments.

lib/vanagon/platform/dsl.rb
jackie-kinsler: I can't think of one, but I left it like that for flexibility. It'd be easy to modify it to your suggestion.

lib/vanagon/platform/defaults/osx-10.15-x86_64.rb
jackie-kinsler: I'll double quote everything for consistency/readability

lib/vanagon/platform/dsl.rb
e-gris: Yea, let's. I'm trying to gently nudge us more towards a prescriptive approach inside the RE codebase, removing flexibility as much as is sensible.

It's a tough call between trying to fend off unpredictability/weird corner cases and forbidding valid use.

e-gris (approved):

lib/vanagon/platform/defaults/osx-10.14-x86_64.rb
```diff
@@ -0,0 +1,22 @@
...
+  plat.provision_with "sudo dscl . -create /Users/test NFSHomeDirectory /Users/test"
+  plat.provision_with "sudo dscl . -passwd /Users/test password"
+  plat.provision_with "sudo dscl . -merge /Groups/admin GroupMembership test"
+  plat.provision_with "echo 'test ALL=(ALL:ALL) NOPASSWD: ALL' > /etc/sudoers.d/username"
+  plat.provision_with "mkdir -p /etc/homebrew"
+  plat.provision_with "cd /etc/homebrew"
+  plat.provision_with %Q(su test -c 'echo | /usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"')
+  plat.provision_with "sudo chown -R test:admin /Users/test/"
+  plat.vmpooler_template "osx-1014-x86_64"
+  plat.output_dir File.join("apple", "10.14", "puppet6", "x86_64")
```
gimmyxd, line 21: @jackie-kinsler is this forgotten here? I do not see it for other os x platforms

PR #700, opened 2021-06-11
(PA-3755) add --extended-attributes on mac OS

gimmyxd:
`-E, --extended-attributes - copy extended attributes`
 This flag is needed because the code signature is
 stored as a xattr.

### DIFF
--- a/lib/vanagon/utilities/extra_files_signer.rb
+++ b/lib/vanagon/utilities/extra_files_signer.rb
@@ -19,11 +19,13 @@ class Vanagon
             remote_host = "#{project.signing_username}@#{project.signing_hostname}"
             remote_destination_path = "#{remote_host}:#{tempdir}"
             remote_file_location = "#{remote_host}:#{file_location}"
+            extra_flags = ''
+            extra_flags = '--extended-attributes' if project.platform.is_macos?
 
             commands += [
-              "rsync -e '#{Vanagon::Utilities.ssh_command}' --verbose --recursive --hard-links --links  --no-perms --no-owner --no-group #{local_source_path} #{remote_destination_path}",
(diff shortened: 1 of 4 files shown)


e-gris (approved):

ciprianbadescu (approved):

lib/vanagon/utilities/extra_files_signer.rb
```diff
@@ -19,11 +19,13 @@ def commands(project, mktemp, source_dir) # rubocop:disable Metrics/AbcSize
             remote_host = "#{project.signing_username}@#{project.signing_hostname}"
             remote_destination_path = "#{remote_host}:#{tempdir}"
             remote_file_location = "#{remote_host}:#{file_location}"
+            extra_flags = ''
```
GabrielNagy, line 22: I don't think this line is needed

gimmyxd: It is, for other platforms that use the signing DSL, like windows.

GabrielNagy: Yes, but assigning a variable like `abc = 'fgh' if something` will set the variable to nil if the condition is not satisfied, right? Tests still pass for me after removing this line.

gimmyxd: Yes that creates the `extra_flags` variable, I usually like to have a sane default, this keeps the code in concordance with other sides of Vanagon: https://github.com/puppetlabs/vanagon/blob/main/lib/vanagon/driver.rb#L132

GabrielNagy: :+1:

PR #701, opened 2021-06-30
(RE-14154) Exit if `VANAGON_FORCE_SIGNING` is set

GabrielNagy:
By default, Vanagon does not fail if extra files signing fails, it just logs an error and continues building the package. This is unwanted behavior in environments where we expect a hard failure when signing cannot proceed.

To force Vanagon to fail if extra files signing fails, ensure the VANAGON_FORCE_SIGNING variable is set before starting a build.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ This changelog adheres to [Keep a CHANGELOG](http://keepachangelog.com/).
 ### Added
 - (PA-3709) Add Debian 11 64-bit support
 - (PA-3755) add `--extended-attributes` on mac OS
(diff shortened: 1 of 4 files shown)


GabrielNagy:
Validated this locally, confirmed that exporting `VANAGON_FORCE_SIGNING` makes the build fail if the files cannot be signed.

gimmyxd (approved):

e-gris (commented):
Speaking from a position of ignorance, I would like this behavior (fail hard) to be the default, though I agree that this approach isn't breaking.

Not sure if there's a way to introduce this change so that we can get to fail-hard-as-default without it being painful to other users.

e-gris (approved):

PR #715, opened 2021-10-04
(RE-14305) Add dependencies cli command

isaac-hammes:
The dependencies command generates a json file which lists all required gems and gem versions in a json file for a given project and platform.

### DIFF
--- a/lib/vanagon/cli/dependencies.rb
+++ b/lib/vanagon/cli/dependencies.rb
@@ -4,10 +4,10 @@ require 'vanagon/logger'
 
 class Vanagon
   class CLI
-    class Lock < Vanagon::CLI
+    class Dependencies < Vanagon::CLI
       DOCUMENTATION = <<~DOCOPT.freeze
         Usage:
-        lock [options] <project-name> <platforms>
+        dependencies [options] <project-name> <platforms>
 
         Options:
           -h, --help                       Display help
@@ -54,12 +54,28 @@ class Vanagon
           end
         end
 
+        temp_dir = Dir.mktmpdir
+        failures = []
+
         projects.each do |project|
           platforms.each do |platform|
-            artifact = Vanagon::Driver.new(platform, project, options)
-            artifact.lock
+            begin
+              artifact = Vanagon::Driver.new(platform, project, options)
+              artifact.dependencies(temp_dir)
+            rescue => e
+              failures.push("#{project}, #{platform}: #{e}")
+            end
           end
         end
+
+        if !failures.empty?
+          VanagonLogger.info "Failed to generate dependencies for the following:"
+          failures.each do |failure|
+            VanagonLogger.info failure
+          end
+        end
+
+        VanagonLogger.info "Dependency files located at: #{temp_dir}"
       end
 
       def options_translate(docopt_options)
--- a/lib/vanagon/cli/dependencies.rb
+++ b/lib/vanagon/cli/dependencies.rb
@@ -31,49 +31,50 @@ class Vanagon
       end
 
       def run(options)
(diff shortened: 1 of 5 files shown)


lib/vanagon/cli/dependencies.rb
```diff
@@ -0,0 +1,91 @@
...
+        Docopt.docopt(DOCUMENTATION, { argv: argv })
+      rescue Docopt::Exit => e
+        VanagonLogger.error e.message
+        exit 1
+      end
+
+      def run(options)
+        if Dir.exist?(File.join(options[:configdir], 'platforms')) == false ||
+           Dir.exist?(File.join(options[:configdir], 'projects')) == false
+
```
e-gris, line 36: Don't compare directly to `false`(or `true`). Ruby has a notion of `falsey` which includes `false` and `nil` which is implicit with `if` and `unless`. I suggest:

```ruby
platforms_directory = File.join(options[:configdir], 'platforms')
projects_directory = File.join(options[:configdir], 'projects')

unless Dir.exist?(project_directory) && Dir.exist?(platforms_directory)
  VanagonLogger.error (...)
end

File: lib/vanagon/cli/dependencies.rb [hunk 2]
```diff
@@ -0,0 +1,91 @@
...
+        end
+
+        platforms = options[:platforms].split(',')
+        if options[:platforms] == 'all'
+          platforms = Dir.children(File.join(options[:configdir], 'platforms')).map do |platform|
+            File.basename(platform, File.extname(platform))
+          end
+        end
+
+        temp_dir = Dir.mktmpdir
```
e-gris, line 55: Nitpicky: when it doesn't disrupt existing code, prefer whole words over abbreviations: `temporary_directory`

lib/vanagon/cli/dependencies.rb
```diff
@@ -0,0 +1,91 @@
...
+            begin
+              artifact = Vanagon::Driver.new(platform, project, options)
+              artifact.dependencies(temp_dir)
+            rescue => e
+              failures.push("#{project}, #{platform}: #{e}")
+            end
+          end
+        end
+
+        if !failures.empty?
```
e-gris, line 69: Use: `unless failures.empty?`

lib/vanagon/driver.rb
```diff
@@ -182,6 +182,17 @@ def render # rubocop:disable Metrics/AbcSize
       @project.make_makefile(workdir)
     end
 
+    def dependencies(temp_dir) # rubocop:disable Metrics/AbcSize
+      # Simple sanity check for the project
+      if @project.version.nil? or @project.version.empty?
```
e-gris, line 187: `or` does surprising things in Ruby (I've read appeals that it, along with `and` be deprecated). Always favor `||`

lib/vanagon/cli/dependencies.rb
```diff
@@ -0,0 +1,91 @@
...
+      def run(options)
+        if Dir.exist?(File.join(options[:configdir], 'platforms')) == false ||
+           Dir.exist?(File.join(options[:configdir], 'projects')) == false
+
+          VanagonLogger.error "Path to #{File.join(options[:configdir], 'platforms')} or #{File.join(options[:configdir], 'projects')} not found."
+          exit 1
+        end
+
+        projects = [options[:project_name]]
+        if options[:project_name] == 'all'
```
e-gris, line 42: Suggest something like:

```ruby
projects = [options[:project_name]]
if projects.include?('all')
   <go get 'em all>
end
```

I think gathering reused items like `project_name` and

e-gris (changes requested):

PR #717, opened 2021-10-06
(VANAGON-162) Added new instance variable 'log_url' to use in logs rather than the full git url

isaac-hammes:
This 'log_url' only shows the host/path.


jackie-kinsler:
Looking good, but looks like some of the rake tests need to be updated, so it passes CI.

PR #718, opened 2021-10-06
(RE-14305) Changed depencies output from temporary directory to STDOUT

isaac-hammes:


e-gris (approved):
I like this behavior better

PR #720, opened 2021-11-01
(maint) Allow target on CLI to set user and port

nicklewis:
This allows the CLI target argument to override the target user and port
specified in the platform config. This is useful for local dev
environments that don't match the enviroment used in the real build
pipeline. Previously, this required temporarily modifying the platform
config, which is more onerous than simply passing the argument on the
CLI.


e-gris:
@nicklewis Looks like this was dropped. I'm going to approve/merge if you still think its worthwhile.

nicklewis:
That would be very helpful!

e-gris (approved):

PR #722, opened 2021-12-01
(VANAGON-179) Support ruby 3 for vanagon localhost

traylenator:
Allow vanagon to be run with ruby 3. In particular CentOS stream 9
is ruby 3.0.2p107.

aws-sdk was bumped as the the existing 2.2.37 was non functional with ruby 3

### DIFF
--- a/.github/workflows/ruby3.yml
+++ b/.github/workflows/ruby3.yml
@@ -1,0 +1,20 @@
+name: Ruby3
+
+on: [push]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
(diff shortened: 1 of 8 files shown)


traylenator:
Re the CLA there is an agreement between CERN and Puppetlabs that all CERN staff have signed the CLA.

stahnma:
FWIW, this patch seems to work quite nicely. I applied all the bits that didn't conflict (like changelog) and it fixed my issues for using vanagon with a more modern ruby.

underscorgan:
Was able to confirm that we have a CLA with cern, opened #735 to resolve merge conflicts

PR #723, opened 2021-12-02
(VANAGON-174) Add el-9 definitions to vanagon

traylenator:
Add platforms to support building on el-9 based systems

In CentOS Stream 9

```
dnf install '/bin/mkdir'
Last metadata expiration check: 0:00:09 ago on Thu 02 Dec 2021 12:25:13 PM CET.
No match for argument: /bin/mkdir
Error: Unable to find a match: /bin/mkdir
```

where as

```
dnf install '/usr/bin/mkdir'
```

does install `coreutils-single`

Note that Fedora < 26 has already been deprecated and that case can be
ignored.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ This changelog adheres to [Keep a CHANGELOG](http://keepachangelog.com/).
 - (VANAGON-168) Remove Fedora 30 x86_64
 - (VANAGON-169) Remove Fedora 31 x86_64
 - (VANAGON-170) Remove OSX 10.14 x86_64
+- (VANAGON-174) Addition 'el-9' platform
 - (RE-14305) Add 'vanagon dependencies' command to generate gem dependencies as a json file
 - (VANAGON-162) Added new instance variable 'log_url' to use in logs rather than the full git url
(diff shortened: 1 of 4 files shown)


traylenator:
Re CLA CERN and PUPPETLABS have an agreement.

Unclear if this is what you want to support CentOS 9 stream - maybe those pooler names should be different with different platform.

The above does build a (patched)puppet-runtime and  (patched)puppet-agent on CentOS 9 stream however.

e-gris:
@underscorgan We may need to investigating intervening with the CLAassistant bot.

ciprianbadescu (approved):

ciprianbadescu:
This is good to merge, pending only CLA agreement

ciprianbadescu:
@traylenator, we need to get this change merged, can you sign the CLA?

traylenator:
> @traylenator, we need to get this change merged, can you sign the CLA?

I believe I am not able to since CERN must sign. This I believe to be case but I checking from  this side as well.

 /cc @binford2k

traylenator:
>  can you sign the CLA?

Can confirm now CLA has now been signed by CERN and delivered to PLs.

PR #725, opened 2022-01-06
VANAGON-165

isaac-hammes:


jackie-kinsler (approved):

lib/vanagon/cli/sign.rb
```diff
@@ -28,7 +28,7 @@ def run(_)
 
         require 'packaging'
         Pkg::Util::RakeUtils.load_packaging_tasks
```
e-gris, line 30: This is likely no longer needed. Since we're not using tasks, we shouldn't need to load them.

e-gris (commented):
One comment. Not sure if it'll have any funny side effects.

PR #730, opened 2022-01-11
(maint) prep for 0.24.0 release

isaac-hammes:


e-gris:
Looks like vanagon is suffering from the now broken 'git://'

I think we should repair that (likely in a separate PR) before issuing a new release.

PR #732, opened 2022-01-18
(maint) Use Priority 3 ABS requests, identifiable user / job name

suckatrash:
Before this change ABS requests were made at priority 1 and the job ID would look like: `unknown-1642527707`

This change lowers the priority to 3 and makes the job id more identifiable

### DIFF
--- a/lib/vanagon/engine/always_be_scheduling.rb
+++ b/lib/vanagon/engine/always_be_scheduling.rb
@@ -346,7 +346,7 @@ class Vanagon
       end
 
       def build_request_object
-        user = ENV['USER'] || ENV['USERNAME'] || 'unknown'
+        user = ENV['USER'] || ENV['USERNAME'] || 'vanagon'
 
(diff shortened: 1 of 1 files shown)


lib/vanagon/engine/always_be_scheduling.rb
```diff
@@ -346,7 +346,7 @@ def translated(response_body, job_id)
       end
 
       def build_request_object
-        user = ENV['USER'] || ENV['USERNAME'] || 'unknown'
+        user = ENV['USER'] || ENV['JOB_NAME'] || 'vanagon'
```
suckatrash, line 349: Will JOB_NAME work here?  I'm actually not sure

suckatrash: Only changing the fallback value just in case

isaac-hammes (approved):

e-gris (approved):

e-gris:
Jenkins, please deploy experimental enterprise puppet-enterprise-vanagon

PR #745, opened 2022-04-26
(VANAGON-192) Github no longer supports git://

yakatz:
Tests are broken as of March 15, 2022 due to Github dropping support for non-encrypted connections


yakatz:
Not as simple as I thought - these requests are now blocked by the testing setup?

```
  WebMock::NetConnectNotAllowedError:
       Real HTTP connections are disabled. Unregistered request: HEAD https://artifactory.delivery.puppetlabs.net/artifactory/generic/buildsources/facter with headers {'Accept'=>'*/*', 'Host'=>'artifactory.delivery.puppetlabs.net', 'User-Agent'=>'Ruby'}
```

PR #757, opened 2022-07-08
(maint) Removes optional Mac installer attribute

mhashizume:
`hostArchitecture` is an optional attribute and what we specify
now (i386) is outdated and causes our arm64/M1 packages to not
properly install.


e-gris (commented):
LGTM, but also needs a note in CHANGELOG.md

cthorn42:
When running this change in the adhoc pipeline we ran into the same problem when trying to install the macos-12-arm64 pkg. When I tried Josh's suggestion with https://github.com/puppetlabs/vanagon/pull/758, I was able to install the pkg on the macos-12-arm64 host. Going to close this out in favor of the above PR.

PR #759, opened 2022-07-22
(maint) Add brew method for macOS

mhashizume:
In 2020, Apple began releasing machines that used new ARM-based  processors, as opposed to the x86 processors it had used for many years prior.

Homebrew, a macOS package manager, changes its path depending on whether it's running on x86-64 or ARM64.

This commit adds a new `brew` method to Vanagon that defaults to the x86 path (/usr/local/bin/brew) but can be overriden to the ARM path (/opt/homebrew/bin/brew).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ This changelog adheres to [Keep a CHANGELOG](https://keepachangelog.com).
 
 ## [Unreleased]
 
+### Added
+- (maint) Adds `brew` attribute to specify the path of Homebrew on macOS on different architectures.
+
 ## [0.37.1] - release 2023-06-21
 ### Changed
 - (maint) Add TLS 1.2 to the platform add command for Windows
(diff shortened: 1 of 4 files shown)


joshcooper (commented):
In order to incrementally add this for macOS 13 ARM, I'm wondering if vanagon should add a method for configuring the brew path like we do for https://github.com/puppetlabs/vanagon/blob/d6da768a0ba72ffe5fa4f09089b0dae7e82331bc/lib/vanagon/platform/dsl.rb#L179 If the brew path isn't set then default to `/usr/local/bin/brew`. And then in `lib/vanagon/platform/defaults/osx-13-arm64.rb` add something like:

```ruby
plat.brew '/opt/homebrew/bin/brew'
```

mhashizume:
I updated this PR to add the method that @joshcooper suggested and ran it through the Vanagon generic builder to build agent-runtime-7.x: https://jenkins-platform.delivery.puppetlabs.net/view/vanagon-generic-builder/job/platform_vanagon-generic-builder_vanagon-packaging-ship-repos_generic-builder/1069/

e-gris (approved):

joshcooper (approved):

PR #762, opened 2022-08-02
(maint) Update extra file sigining to accommodate new windows extra file signing. 

isaac-hammes:


e-gris (approved):
LGTM

A future vanagon improvement would be to take the RemoteScript class I wrote (https://github.com/puppetlabs/enterprise-dist-tasks/blob/main/lib/remote_script.rb) and migrate to that over Vanagon::Utilities.ssh_command. This should alleviate most quoting issues.

PR #772, opened 2022-09-28
(maint) osx-10.15 installs Homebrew via the bash method

cthorn42:
Homebrew has removed the ability to be installed via ruby, so this PR changes to the bash method the other newer versions of MacOS use.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ This changelog adheres to [Keep a CHANGELOG](http://keepachangelog.com/).
 - (VANAGON-197) Remove support for Debian 9 (x86 and x86-64)
 
 ### Changed
+- (maint) osx-10-15.x86-64 installs homebrew via bash rather then ruby
 
 ## [0.30.0] - release 2022-08-18
 ### Changed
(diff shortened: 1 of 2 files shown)


cthorn42:
Currently all runtimes for osx-10.15 are failing with:
```
Error: The Ruby Homebrew installer is now disabled and has been rewritten in
Bash. Please migrate to the following command:
  /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
```
The above method is what the newer MacOS is using, trying out a build locally with that and it got past this PR initial homebrew install.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 This changelog adheres to [Keep a CHANGELOG](http://keepachangelog.com/).
 
 ## [Unreleased]
+- (maint) osx-10-15.x86-64 installs homebrew via bash rather then ruby
```
e-gris, line 8: Should be under a '### Changed' section.

e-gris (changes requested):
Small request regarding the CHANGELOG

e-gris (approved):

PR #776, opened 2022-10-13
(RE-15209) Exempt github URLs from validation checking

e-gris:
Exempt github URLs from being checked as valid git repositories in
order to avoid rate-limiting from excessive traffic.

### DIFF
--- a/lib/vanagon/component/source/git.rb
+++ b/lib/vanagon/component/source/git.rb
@@ -18,11 +18,12 @@ class Vanagon
 
         class << self
           # Attempt to connect to whatever URL is provided and
-          # return True or False depending on whether or not
+          # return true or false depending on whether or not
           # `git` thinks it's a valid Git repo.
           #
-          # @param url
-          # @param timeout Time (in seconds) to wait before assuming the
+          # @param url [#to_s] A URI::HTTPS, URI:HTTP, or String with the the URL of the
+          #        remote git repository.
+          # @param timeout [Number] Time (in seconds) to wait before assuming the
           #        git command has failed. Useful in instances where a URL
           #        prompts for credentials despite not being a git remote
           # @return [Boolean] whether #url is a valid Git repo or not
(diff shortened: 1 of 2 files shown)


lib/vanagon/component/source/git.rb
```diff
@@ -34,30 +35,20 @@ class << self
...
+                Vanagon::Utilities.local_command("git ls-remote --heads #{url} > /dev/null 2>&1")
+                return $?.exitstatus.zero?
+              end
+            rescue StandardError
+              return false
             end
-          rescue Timeout::Error
-            return false
-          rescue RuntimeError
-            return false
```
joshcooper, line 60: The `valid_remote?` method will now rescue `StandardError` (or derived classes) that aren't `RuntimeError`. I think it only matters if callers are expecting the build to fail if one of those other exceptions are raised like `ArgumentError`?  This picture is helpful https://www.honeybadger.io/blog/understanding-the-ruby-exception-hierarchy/

e-gris: Yup, that was too aggressive. Remedy could be putting it back:

```ruby
rescue Timeout::Error
  return false
rescue RuntimeError
  return false
```
but this reads as misleading to me because I take this as an effective shadow and it smells a little bit.

But, getting rid of the shadow:

```ruby
rescue RuntimeError
  return false
```
relieves the smell but then explicitly loses the reason for the failure, Timeout vs. git ls-remote. 

I'm actually leaning towards the former. Any thoughts?

joshcooper: > because I take this as an effective shadow

As long as the more specific/derived class is rescued first, which it is in this case, then it's ok. That pattern is often used when you need different behavior based on the exception class:

```ruby
rescue A => a
  puts a.method1
rescue B => b
  puts b.method2
end
```
 
However, in this case, `Timeout::Error` extends `RuntimeError` and we return `false` in both cases, so there's no real benefit to rescuing them separately.

isaac-hammes (approved):

joshcooper (approved):

PR #777, opened 2022-11-08
(VANAGON-206) Add solaris 10 sparc

tvpartytonight:

### DIFF
--- a/lib/vanagon/platform/defaults/solaris-10-sparc.rb
+++ b/lib/vanagon/platform/defaults/solaris-10-sparc.rb
@@ -1,0 +1,10 @@
+platform "solaris-10-sparc" do |plat|
+  plat.servicedir "/lib/svc/manifest"
+  plat.defaultdir "/lib/svc/method"
+  plat.servicetype "smf"
+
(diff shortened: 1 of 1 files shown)


lib/vanagon/platform/defaults/solaris-10-sparc.rb
```diff
@@ -0,0 +1,10 @@
+platform "solaris-10-sparc" do |plat|
+  plat.servicedir "/lib/svc/manifest"
+  plat.defaultdir "/lib/svc/method"
+  plat.servicetype "smf"
+
+  plat.cross_compiled true
+  plat.vmpooler_template "solaris-10-x86_64"
+  plat.add_build_repository "http://solaris-10-reposync.delivery.puppetlabs.net:81", "puppetlabs.com"
```
tvpartytonight, line 8: This `solaris-10-reposync` resource doesn't actually exist, but it is what I imagine would need to be here if these defaults are similar to what we do for solaris 11 sparc.

e-gris:
@tvpartytonight Is still interesting or relevant?

PR #780, opened 2022-11-30
(maint) update deprecated Bundler call

e-gris:
It's now `with_unbundled_env` rather than `with_clean_env`


isaac-hammes:
@e-gris Do we need vanagon to work with bundler < v2.1? I think the with_unbundled_env will fail on older bundler versions.

e-gris:
Too much risk for the reward. pl-build-tools or some of our cross-compilation environments could be lagging here.  I'll postpone until I've had an opportunity to do a thorough test.

PR #793, opened 2023-03-24
(PUP-11786) Ruby 3.1 compatibility updates

mhashizume:
This PR:

- Updates fusitgit to enable compatibility with newer version of the URI module included in Ruby >= 3.1.
- Updates method calls to explicitly use keyword arguments instead of implicitly assuming that the last positional argument is a keyword argument.

More details in the commit messages.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ This changelog adheres to [Keep a CHANGELOG](http://keepachangelog.com/).
 
 ### Changed
 - (PUP-11786) Update fustigit gem to version 0.3.0.
+- (PUP-11786) Use keyword arguments in method calls.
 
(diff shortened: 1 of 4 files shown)


mhashizume:
Looks like this is failing the Ruby check because the check runs on Ruby 2.6 and fustigit 0.3.0 requires Ruby 2.7.

Now that PE 2019.8.x / Puppet 6 have been retired, the current OSP and PEs use Ruby >= 2.7. Would it be alright for me to include this version of fustigit and bump the Ruby version in the check to 2.7?

e-gris:
See: https://github.com/puppetlabs/vanagon/pull/794

mhashizume:
I'll close this in favor of #794 , thanks @e-gris !

PR #802, opened 2023-06-12
(maint) Use keyword arguments

mhashizume:
In Ruby >= 3.0, keyword arguments are separated from positional arguments:

https://bugs.ruby-lang.org/issues/14183

This commit updates method calls that passes a keyword argument as the last argument to separate out those keyword arguments.


e-gris (approved):
Hmm, I'm not sure if we still need to test against 2.7, they seem to be missing now.

Nevertheless, I'm nearly certain this is allowed in >=2.6 but might be worthy to check.

PR #813, opened 2023-08-28
(maint) Handle an invalid mirror gracefully

nmburgan:
Because the HTTP source's valid_url? check was always returning truthy before, a mirror URL to a file would always seem to be valid, even if it wasn't. Now that https://github.com/puppetlabs/vanagon/commit/9995d5ef7f0370eb41f327ca36dad5a40ea95f9d fixed it, a Vanagon::Error would get thrown saying it was an unknown file type. This means that fetch_url would never get called since this exception in fetch_mirrors would be unhandled.

This handles that particular error, and continues to throw other Vanagon::Error types, as they usually indicate a problem with the definition of the component or some other type of fatal error. Additionally, this changes the catching of RuntimeError to StandardError, so that we can handle issues talking to the particular URL, such as OpenSSL errors.

### DIFF
--- a/lib/vanagon/component.rb
+++ b/lib/vanagon/component.rb
@@ -260,14 +260,21 @@ class Vanagon
           VanagonLogger.info %(Attempting to fetch from mirror URL "#{mirror}")
           @source = Vanagon::Component::Source.source(mirror, **options)
           return true if source.fetch
+        rescue Vanagon::InvalidSource
+          # This means that the URL was not a git repo or a valid downloadable file,
+          # which means either the URL is incorrect, or we don't have access to that
+          # resource. Return false, so that the pkg.url value can be used instead.
+          VanagonLogger.error %(Invalid source "#{mirror}")
         rescue SocketError
           # SocketError means that there was no DNS/name resolution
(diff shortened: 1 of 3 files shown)


e-gris (approved):
LGTM, but please add a CHANGELOG entry.

lib/vanagon/component.rb
```diff
@@ -260,14 +260,28 @@ def fetch_mirrors(options)
           VanagonLogger.info %(Attempting to fetch from mirror URL "#{mirror}")
           @source = Vanagon::Component::Source.source(mirror, **options)
           return true if source.fetch
+        rescue Vanagon::Error => e
+          if e.message =~ /Unknown file type/
```
joshcooper, line 264: I think I'd prefer raising an subclass of `Vanagon::Error` specifically for the unknown source case and rescue the most specific exceptions first like:

```ruby
rescue Vanagon::UnknownSourceError
  VanagonLogger.error %(Unable to access mirror URL "#{mirror}")
rescue Vanagon::Error 
  raise
rescue SocketError
  ...
rescue StandardError
  ...
```

joshcooper (approved):

PR #815, opened 2023-09-18
(PA-5328) Add Debian 11 (ARM) platform definition to vanagon

shubhamshinde360:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ This project adheres to [Semantic Versioning](https://semver.org).
 This changelog adheres to [Keep a CHANGELOG](https://keepachangelog.com).
 
 ## [Unreleased]
+### Added
+- (PA-5328) Add support for Debian 11 (ARM64)
+
 ### Changed
 - (maint) Force psych gem to use >= 4.0 to avoid breaking change in safe_load between v3 and
   v4 of the gem.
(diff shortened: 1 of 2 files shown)


shubhamshinde360:
Seems some github actions checks are failing due to incompatibility of latest psych gem version with [this](https://github.com/puppetlabs/vanagon/blob/3655b254dc4c53a7165664388478cfcd977df4df/lib/vanagon/project.rb#L879C1-L890C8) code. Looks like the issue might be with the newer version of psych (5.1.0) and not ruby?

e-gris:
I just merged a fix. This should work after a rebase.

mhashizume:
The psych issue looks like to be because of a transitive dependency. The Gemfile requires debug >= 1.0.0:
https://github.com/puppetlabs/vanagon/blob/3655b254dc4c53a7165664388478cfcd977df4df/Gemfile#L24
Which in turn requires irb. irb >= 1.8.0 (1.8.0 was released on August 30th of this year) introduces a dependency on rdoc, which has a dependency on psych. 

I think a good solution to this would be to introduce an explicit development dependency on irb in the Gemfile and pin to the last version that doesn't depend on rdoc:
```
gem 'irb', '1.7.4'
```

e-gris (approved):

joshcooper (approved):

PR #819, opened 2023-10-18
(PA-5786) Stop using posttrans for post-install steps

tvpartytonight:
Before this commit, vanagon used the `%posttrans` section of an rpm spec for #add_postinstall_action. Specifying items in the `%posttrans` section means that rpm considers the transaction done, and rpm continues to excecute irregardless of the result of the `%posttrans` scriptlet. This behavior breaks puppet-agent 8 fips upgrades because certain configuration files are created in a #add_postinstall_action, and those configuration files were not yet ready before the agent daemon is killed and then started as part of the upgrade.

This change eliminates the `%posttrans` section entirely in favor of moving all #add_postinstall_action code to the `%post` section.

### DIFF
--- a/resources/rpm/project.spec.erb
+++ b/resources/rpm/project.spec.erb
@@ -242,13 +242,17 @@ fi
 
 
 %post
-<%- if @platform.is_aix? || (@platform.is_el? && @platform.os_version.to_i == 4) -%>
-## EL-4 and AIX RPM don't have %posttrans, so we'll put them here
-# Run postinstall scripts on install if defined
-if [ "$1" -eq 1 ] ; then
+# Run post-transaction scripts on install if defined
+if [ -e %{_localstatedir}/lib/rpm-state/%{name}/install ] ; then
   <%= get_postinstall_actions("install") %>
+  rm %{_localstatedir}/lib/rpm-state/%{name}/install
 fi
-<%- end -%>
+
+# Run post-transaction scripts on upgrade if defined
+if [ -e %{_localstatedir}/lib/rpm-state/%{name}/upgrade ] ; then
+  <%= get_postinstall_actions("upgrade") %>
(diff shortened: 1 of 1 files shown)


e-gris (approved):
Overall, I think this is right by removing unnecessary hand-waving. However, I was curious _why_ we added %posttrans in the first place:

    commit 9a8930c8f62da177d7ca32320853e232b8e4e8ca
    Author: Melissa Stone <melissa@puppetlabs.com>
    Date:   Thu Oct 22 11:22:27 2015 -0700

    (CPR-234) Teach Vanagon how to add packaging scriptlets to rpm packaging
    
    Prior to this commit, rpm packaging only new how to handle pre and post
    install scripts, and had no ability to handle special cases, like if we
    wanted to only run a script on install and not upgrade. This commit adds
    the ability to handle pre and post removal scripts, and adds the ability
    to handle special instructions, even in the %posttrans section.

Unfortunately, I think we've lost CPR-234 to the Jira migration, but I will spend a little time to see if its archived somewhere. I'm curious what prompted it.

Further, I had a look at the corresponding deb commit: 10b157a just in case some parallel action was needed. From my reading, it isn't.

Also: please add an entry in CHANGELOG.md

e-gris:
Oh, cool, we do have the original ticket: https://tickets-archive.ops.puppetlabs.net/browse/CPR-234

My summary: "we need this for razor" so, I'm doubling my approval here.

joshcooper (approved):
There are some other vanagon projects where `add_postinstall_action` is called: https://github.com/search?q=org%3Apuppetlabs+add_postinstall_action+Language%3ARuby&type=code

@nmburgan thoughts about whether this will break ace-vanagon, pe-bolt-vanagon, puppet-enterprise-vanagon, etc?

PR #820, opened 2023-10-20
(PA-5786) Add ability to execute direct post installation scriptlets

tvpartytonight:
The existing `#add_postinstall_action` adds actions to the `%posttrans` section of an rpm spec file. This can mean that the action could fail, but the installation/upgrade could still succeed. This commit adds the ability to add actions that must succeed post installation for rpm installations by adding the action to the `%post` section.

### DIFF
--- a/lib/vanagon/project.rb
+++ b/lib/vanagon/project.rb
@@ -494,6 +494,18 @@ class Vanagon
       end
     end
 
+    # Collects the postinstall packaging actions that must exit successfully for the project and it's components
+    #  for the specified packaging state
+    #
+    # @param pkg_state [String] the package state we want to run the given scripts for.
+    #   Can be one of 'install' or 'upgrade'
+    # @return [String] string of Bourne shell compatible scriptlets to execute during the postinstall
+    #   phase of packaging during the state of the system defined by pkg_state (either install or upgrade)
+    def get_postinstall_required_actions(pkg_state)
+      scripts = components.flat_map(&:postinstall_required_actions).compact.select { |s| s.pkg_state.include? pkg_state }.map(&:scripts)
+      return ': no postinstall required scripts provided' if scripts.empty?
+      scripts.join("\n")
+    end
     # Collects the preremove packaging actions for the project and it's components
     # for the specified packaging state
     #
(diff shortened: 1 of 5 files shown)


tvpartytonight:
This might be preferable to #819 as it should be safer and would only affect projects using `#add_postinstall_required_action`.

joshcooper (approved):

lib/vanagon/project.rb
```diff
@@ -494,6 +494,21 @@ def get_postinstall_actions(pkg_state)
...
+    #
+    # @param pkg_state [String] the package state we want to run the given scripts for.
+    #   Can be one of 'install' or 'upgrade'
+    # @return [String] string of Bourne shell compatible scriptlets to execute during the postinstall
+    #   phase of packaging during the state of the system defined by pkg_state (either install or upgrade)
+    def get_postinstall_actions(pkg_state)
+      scripts = components.flat_map(&:postinstall_required_actions).compact.select { |s| s.pkg_state.include? pkg_state }.map(&:scripts)
+      if scripts.empty?
+        return ': no postinstall required scripts provided'
+      else
```
e-gris, line 508: Style only: `else` is redundant. Favor a guardian clause:

```ruby
return ': no postinstall required scripts provided' if scripts.empty?

[return] scripts.join("\n")
```

The 2nd return is optional depending on your aesthetic preferences.

e-gris (approved):
A couple of ignorable nitpicky style comments but otherwise LGTM.

I actually liked the previous one better because if the cleanliness it introduced, but understand the reasoning to head this way to avoid breaking things.

lib/vanagon/project.rb
```diff
@@ -485,7 +485,7 @@ def get_activate_triggers()
     #   Can be one of 'install' or 'upgrade'
     # @return [String] string of Bourne shell compatible scriptlets to execute during the postinstall
     #   phase of packaging during the state of the system defined by pkg_state (either install or upgrade)
-    def get_postinstall_actions(pkg_state)
+    def get_postinstall_required_actions(pkg_state)
```
tvpartytonight, line 488: This is titled incorrectly and should be the name of the function below.

e-gris:
Also, when ready, please add a CHANGELOG update.

tvpartytonight:
@e-gris I think I have address all comments, let me know if there is anything I missed.

e-gris:
Put an '### Added' header above the changelog entry and all will be well.

PR #825, opened 2023-10-27
(PA-5786) Add ability to execute direct post installation scripts

tvpartytonight:
The existing `#add_postinstall_action` adds actions to the `%posttrans` section of an rpm spec file. This can mean that the action could fail, but the installation/upgrade could still succeed. This commit adds the ability to add actions that must succeed post installation for rpm installations by adding the action to the `%post` section.

A previous attempt to add this was done in 93ba1b4; that introduced a bug where the install and upgrade files were incorrectly deleted after the %post step. They needed to remain so that the %posttrans rpm section would detect the install/upgrade via vanagon's

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ This project adheres to [Semantic Versioning](https://semver.org).
 This changelog adheres to [Keep a CHANGELOG](https://keepachangelog.com).
 
 ## [Unreleased]
+### Added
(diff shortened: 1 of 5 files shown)


tvpartytonight:
Should wait until releases are live before merging/releasing this.

tvpartytonight:
@e-gris I think this should be good to merge and release now?

e-gris:
Today or tomorrow, just sorting through post-release priorities.

e-gris (approved):

PR #839, opened 2024-03-18
Use build-uri instead of fustigit

joshcooper:
The library is MIT licensed, so should be ok to use.

The author is a ruby core maintainer https://rubygems.org/profiles/ioquatix


joshcooper:
ah nice, thanks!

PR #847, opened 2024-04-17
(VANAGON-242) add amazon linux2 in vanagon

amitkarsale:
Add Amazon Linux2 definition in Vanagon


amitkarsale:
### Not required for Intel.

PR #848, opened 2024-04-17
optimize fetch to use local download

h0tw1r3:
Greatly improves the efficiency of testing local rebuilds by reusing the same workdir.


h0tw1r3:
this change is sorely needed to implement a rudimentary local source cache. very useful between builds with docker.

https://github.com/puppetlabs/puppet/issues/9305

h0tw1r3:
rebased on #857 so tests will pass

PR #851, opened 2024-05-03
(PA-6262) Change cmake to cmake3 form amazon linux 2 arm

imaqsood:

### DIFF
--- a/lib/vanagon/platform/defaults/amazon-7-aarch64.rb
+++ b/lib/vanagon/platform/defaults/amazon-7-aarch64.rb
@@ -3,7 +3,7 @@ platform "amazon-7-aarch64" do |plat|
   plat.defaultdir "/etc/sysconfig"
   plat.servicetype "systemd"
 
-  packages = %w(autoconf automake createrepo gcc gcc-c++ rsync cmake make rpm-libs rpm-build libarchive)
+  packages = %w(autoconf automake createrepo gcc gcc-c++ rsync cmake3  make rpm-libs rpm-build libarchive)
   plat.provision_with("yum install -y --nogpgcheck  #{packages.join(' ')}")
   plat.install_build_dependencies_with "yum install --assumeyes"
   plat.vmpooler_template "amazon-7-arm64"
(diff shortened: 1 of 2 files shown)


isaac-hammes (approved):

lib/vanagon/platform/defaults/amazon-7-aarch64.rb
```diff
@@ -3,8 +3,9 @@
   plat.defaultdir "/etc/sysconfig"
   plat.servicetype "systemd"
 
-  packages = %w(autoconf automake createrepo gcc gcc-c++ rsync cmake make rpm-libs rpm-build libarchive)
+  packages = %w(autoconf automake createrepo gcc gcc-c++ rsync cmake3  make rpm-libs rpm-build libarchive)
   plat.provision_with("yum install -y --nogpgcheck  #{packages.join(' ')}")
+  plat.provision_with("ln -sf /usr/bin/cmake3 /usr/bin/cmake")
```
e-gris, line 8: I somewhat disagree with this. While this instance seems somewhat innocuous, it somewhat violates the principle of least surprise.

If the cmake3 program is called 'cmake3', we should respect that an not, by default, create a somewhat surprising alias behind-the-scenes. As a silly hyperbolic, extension, it would be bad to alias `ls` to `rm`. 

I can think of some possible alternatives:

- In projects that use `cmake`, specify cmake3 when it's available: `cmake = '/usr/bin/cmake3' if File.exist?('/usr/bin/cmake3')`
- In this file set the CMAKE environment variable to '/usr/bin/cmake3'
- As a last resort, put the symlink in /usr/local/bin: /usr/local/bin/cmake --> /usr/bin/cmake3

Each one of these suggestions is, admittedly, less-than-ideal but I feel they are better than making unexpected modifications inside a default file.

e-gris (changes requested):
See my inline comment

lib/vanagon/platform/defaults/amazon-7-aarch64.rb
imaqsood: Agreed, it doesn't seem like a good idea to add a symlink. Let's go ahead and remove it. We'll handle the cmake3 path within the project and its components. Thanks for suggesting this approach.

PR #860, opened 2024-08-23
(VANAGON-257) Enable windowsfips-2016 to use '-fips' candle and heat …

shubhamshinde360:
…at flags

### DIFF
--- a/lib/vanagon/platform/windows.rb
+++ b/lib/vanagon/platform/windows.rb
@@ -192,14 +192,14 @@ class Vanagon
         #   -dr             - Root DirectoryRef to point all components to
         #   -var            - Replace "SourceDir" in the @source attributes of all components with a preprocessor variable
         app_heat_flags = " -dr INSTALLDIR -v -ke -indent 2 -cg AppComponentGroup -gg -srd -t wix/filter.xslt -sreg -var var.AppSourcePath "
-        app_heat_flags += " -fips" if project.platform.name =~ /windowsfips-2012r2/
+        app_heat_flags += " -fips" if project.platform.is_windows? && project.platform.is_fips?
 
         app_source_path = "SourceDir/#{project.settings[:base_dir]}/#{project.settings[:company_id]}/#{project.settings[:product_id]}"
         # Candle.exe preprocessor vars are required due to the above double run of heat.exe, both runs of heat use
         # preprocessor variables
         candle_preprocessor = "-dAppSourcePath=\"#{app_source_path}\" "
         candle_flags = "-arch #{@architecture} #{wix_extensions}"
-        candle_flags += " -fips" if project.platform.name =~ /windowsfips-2012r2/
+        candle_flags += " -fips" if project.platform.is_windows? && project.platform.is_fips?
 
         # Enable verbose mode for the moment (will be removed for production)
         # localisation flags to be added
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -13,21 +13,6 @@ AllCops:
   - examples/**/*
   - lib/vanagon/platform/defaults/*
 
-Capybara/MatchStyle:
-  Enabled: true
-
-Capybara/NegationMatcher:
-  Enabled: true
-
-Capybara/SpecificActions:
-  Enabled: true
-
-Capybara/SpecificFinders:
-  Enabled: true
-
-Capybara/SpecificMatcher:
-  Enabled: true
-
 Gemspec/DeprecatedAttributeAssignment:
   Enabled: true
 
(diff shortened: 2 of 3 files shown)


Gemfile
```diff
@@ -29,8 +29,11 @@ group(:development, :test) do
   gem 'rake', require: false
   gem 'rspec', '~> 3.0', require: false
   gem 'rubocop', '~> 1.0', require: false
+  gem 'rubocop-capybara', require: false
+  gem 'rubocop-factory_bot', require: false
   gem 'rubocop-rake', require: false
   gem 'rubocop-rspec', require: false
+  gem 'rubocop-rspec_rails', require: false
```
bastelfreak, line 36: Why did you add that rubocop plugin? Isn't that specific to rails applications?

shubhamshinde360: @bastelfreak 
We need the gem because it seems we are using some rails cops in vanagon which have now been moved to the 'rubocop-rspec_rails' gem: https://github.com/puppetlabs/vanagon/blob/main/.rubocop.yml#L419C1-L432C16.

joshcooper: This might be a case where we need to remove the cops from rubocop.yml. I don't believe vanagon depends on any rails functionality. It would be very surprising if it did.

lib/vanagon/platform/windows.rb
```diff
@@ -192,14 +192,14 @@ def generate_msi_package(project) # rubocop:disable Metrics/AbcSize
         #   -dr             - Root DirectoryRef to point all components to
         #   -var            - Replace "SourceDir" in the @source attributes of all components with a preprocessor variable
         app_heat_flags = " -dr INSTALLDIR -v -ke -indent 2 -cg AppComponentGroup -gg -srd -t wix/filter.xslt -sreg -var var.AppSourcePath "
-        app_heat_flags += " -fips" if project.platform.name =~ /windowsfips-2012r2/
+        app_heat_flags += " -fips" if project.platform.name =~ /windowsfips-(2012r2|2016)/
```
joshcooper, line 195: To future proof this could we do something like

```suggestion
app_heat_flags += " -fips" if project.platform.is_windows? && project.platform.is_fips?
```

And since this is heat/wix which only makes sense on windows, maybe we can drop the first condition?

Gemfile
```diff
@@ -29,6 +29,8 @@ group(:development, :test) do
   gem 'rake', require: false
   gem 'rspec', '~> 3.0', require: false
   gem 'rubocop', '~> 1.0', require: false
+  gem 'rubocop-capybara', require: false
+  gem 'rubocop-factory_bot', require: false
```
joshcooper, line 33: I think we can drop these too? (and the corresponding cops in rubocop.yml)

shubhamshinde360:
Thanks for your suggestions @joshcooper!
I have incorporated them.

lib/vanagon/platform/windows.rb
```diff
@@ -192,14 +192,14 @@ def generate_msi_package(project) # rubocop:disable Metrics/AbcSize
         #   -dr             - Root DirectoryRef to point all components to
         #   -var            - Replace "SourceDir" in the @source attributes of all components with a preprocessor variable
         app_heat_flags = " -dr INSTALLDIR -v -ke -indent 2 -cg AppComponentGroup -gg -srd -t wix/filter.xslt -sreg -var var.AppSourcePath "
-        app_heat_flags += " -fips" if project.platform.name =~ /windowsfips-2012r2/
+        app_heat_flags += " -fips" if project.platform_is_windows? && project.platform.is_fips?
```
joshcooper, line 195:
```suggestion
        app_heat_flags += " -fips" if project.platform.is_windows? && project.platform.is_fips?
```

lib/vanagon/platform/windows.rb
```diff
@@ -192,14 +192,14 @@ def generate_msi_package(project) # rubocop:disable Metrics/AbcSize
...
 
         app_source_path = "SourceDir/#{project.settings[:base_dir]}/#{project.settings[:company_id]}/#{project.settings[:product_id]}"
         # Candle.exe preprocessor vars are required due to the above double run of heat.exe, both runs of heat use
         # preprocessor variables
         candle_preprocessor = "-dAppSourcePath=\"#{app_source_path}\" "
         candle_flags = "-arch #{@architecture} #{wix_extensions}"
-        candle_flags += " -fips" if project.platform.name =~ /windowsfips-2012r2/
+        candle_flags += " -fips" if project.platform_is_windows? && project.platform.is_fips?
```
joshcooper, line 202:
```suggestion
        candle_flags += " -fips" if project.platform.is_windows? && project.platform.is_fips?
```

lib/vanagon/platform/windows.rb
shubhamshinde360: Ah, sorry. Thanks for noticing!

isaac-hammes (approved):

PR #865, opened 2025-08-29
Respect Provides version in Debian control files

ekohl:
While the version can be passed to RPM spec files, it was completely ignored for Debian. Now there's alignment.

This is a draft because it also needs a patch in openvox-agent because I broke something.


ekohl:
Wrong repo.

